PR #80, opened 2017-01-07
Add base of RSAR support

magcius:
As heard in [Wii Shop Bling](https://www.youtube.com/watch?v=4Y_D6koUbWI), this PR implements RSAR support in vgmtrans. It's not 100% correct, and is missing stuff like proper ADSR / envelope support, but I wanted to put this out there before everything was ironed out. This was heavily based on the SSEQ / NDS support because the formats are very similar.

I'm also very new to the codebase and C++ in general so it's possible I'm committing some horrible sins here. I welcome code review and feedback. I only edited the VS Studio projects because I couldn't attempt to build with Qt / Linux.

Thank you!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,14 +18,18 @@
 # Build results
 [Dd]ebug/
 [Dd]ebugPublic/
+[Dd]ebug_WTL/
 [Rr]elease/
 [Rr]eleases/
+[Rr]elease_WTL/
 x64/
 x86/
 build/
 bld/
 [Bb]in/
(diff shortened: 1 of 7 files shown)


magcius:
Huh. Well, apparently Travis doesn't have a new enough version of g++ to have support for codecvt. Not sure how I'm supposed to convert from ASCII to wstring then.

paulsapps:
You'd have to bump up the min required compiler version, you can install newer ones in travis scripts.

PR #82, opened 2017-01-14
Changes made for Siiva

magcius:
Decided to put everything in one branch so it would be easier to maintain. Closing the other two PRs.


src/main/VGMColl.cpp
```diff
@@ -161,6 +161,12 @@ SF2File *VGMColl::CreateSF2File() {
   return sf2file;
 }
 
+static std::string wstring_to_string(std::wstring str) {
```
mikelow: You should use wstring2string() in common.h. Granted, I prefer your function name.

mikelow:
Appreciate the contribution. Few things:
1) a function for converting wstrings to strings already exists in common.h: wstring2string()
2) what was the reason for removing the SetPatch call in windows/MusicPlayer.cpp? The example in the directmidi docs suggests it's important: http://directmidi.sourceforge.net/midiports.htm#dm33261
3) Can you change the indentation from tabs to 2 spaces for consistency (except for windows/* files already using tabs, which was an oversight on my part)

magcius:
1. Thanks, couldn't find it. I'll fix it up to use this.

2. So, the actual issue is that in Kirby Super Star Ultra (DS), playing certain sequences causes a crash with `DMUS_E_INVALIDPATCH`, since the `SetPatch` isn't wrapped by the try/catch. I don't know that much about DirectMIDI, but it chains up here and sets the `m_dwPatchInCollection` so I thought it was fine: https://github.com/vgmtrans/vgmtrans/blob/master/src/ui/windows/directmidi/CCollection.cpp#L208-L213

Music playback is still fine as far as I can tell, but it's possible I'm missing something.

3. Is there a way to set Visual Studio to do this?

mikelow:
1. Yeah, it's pretty hidden. Should really be in a string helper file. Probably should change the name of the function too.
2. The SetPatch() call was already wrapped in a try/catch. Do you mean that GetInstrument() is causing the exception? If so, why not just include that call in the try/catch block? I'd like to adhere to the directmidi documentation example, if possible.
3. Don't have VS handy, but I'm guessing you need to change the indentation settings:
https://msdn.microsoft.com/en-us/library/ms165343(v=vs.100).aspx

and then use one of the format commands: 
http://stackoverflow.com/questions/5755942/how-do-you-auto-format-code-in-visual-studio

loveemu:
3. If we had an EditorConfig, the indentation settings could be shared easily, among various editors (including Visual Studio with EditorConfig extension)

magcius:
OK, this branch apparently passes Travis now too. I might want to back out the JaiSeq stuff for now since yeah it's a bit hodge-podged together with the two different BMS versions, and banks, since there are 4 different bank formats and they're all scattered across between .aw and .baa files usually.

mikelow:
I quite obliviously missed that most of the code wasn't immediately visible in the Files Changed list until expanded, because all of the format logic is stuffed into the Scanner.cpp files and those files were long enough that github contracted them. So, first piece of feedback would be to place the sequence and instrument set classes into their own files separate from the Scanner files, as the other format decoders do. Second would be to make sure you use header files - declaring classes, structs, enums, etc within them instead of throwing everything into the cpp files.

Glancing over the format decoding code, I probably would have more feedback to give, but it was unclear to me from your last comment if you plan on further revision. At any rate, it would be easier for me to review the code once the larger aforementioned changes are implemented.

magcius:
I'm new to C++ codebases, sorry. I opted to use one .cpp file because of compile times -- none of these classes were used outside of their scanners to begin with, and changing a header file means we recompile the whole codebase.

I'll see about splitting them out this weekend. I do want to upstream this, though maybe it isn't quite ready yet. Maybe let's merge some of the simpler cleanup patches for now?

mikelow:
Sounds good. Would you open a new PR for the changes unrelated to the new formats?

magcius:
Opened #85

magcius:
Split out the first format (BRSAR) into separate files.

sykhro:
Still interested in merging this? The misc fixes are good to go imo, the other format work should be moved to a different pr

loveemu:
I am interested in this wonderful enhancement itself, but I have no enough
motivations and knowledges to review these changes by myself. I entrust the
handling of the change to other maintainers. Thank you.

sykhro:
Closing since some of these have already been merged and format work is being reviewed elsewhere

loveemu:
>  format work is being reviewed elsewhere

#157

3e2j:
Honestly the .baa support is worth it

sykhro:
> Honestly the .baa support is worth it

You are welcome to submit a PR :)

PR #86, opened 2017-02-10
Add basic support for RSAR

magcius:
This gets basic RBNK and RSEQ data working yet. I still have to figure out
RBNK in more detail for things like ADSR envelopes.

Split out change from #82.

### DIFF
--- a/VGMTrans_WTL.vcxproj
+++ b/VGMTrans_WTL.vcxproj
@@ -279,6 +279,9 @@
     <ClCompile Include="src\main\formats\PandoraBoxSnesScanner.cpp" />
     <ClCompile Include="src\main\formats\PandoraBoxSnesSeq.cpp" />
     <ClCompile Include="src\main\formats\RareSnesInstr.cpp" />
+    <ClCompile Include="src\main\formats\RSARInstrSet.cpp" />
+    <ClCompile Include="src\main\formats\RSARScanner.cpp" />
+    <ClCompile Include="src\main\formats\RSARSeq.cpp" />
     <ClCompile Include="src\main\formats\SNESDSP.cpp" />
     <ClCompile Include="src\main\formats\SoftCreatSnesScanner.cpp" />
     <ClCompile Include="src\main\formats\SoftCreatSnesSeq.cpp" />
@@ -480,6 +483,10 @@
     <ClInclude Include="src\main\formats\PandoraBoxSnesScanner.h" />
     <ClInclude Include="src\main\formats\PandoraBoxSnesSeq.h" />
     <ClInclude Include="src\main\formats\RareSnesInstr.h" />
+    <ClInclude Include="src\main\formats\RSARFormat.h" />
+    <ClInclude Include="src\main\formats\RSARInstrSet.h" />
+    <ClInclude Include="src\main\formats\RSARScanner.h" />
+    <ClInclude Include="src\main\formats\RSARSeq.h" />
     <ClInclude Include="src\main\formats\SNESDSP.h" />
     <ClInclude Include="src\main\formats\SoftCreatSnesFormat.h" />
     <ClInclude Include="src\main\formats\SoftCreatSnesScanner.h" />
--- a/VGMTrans_WTL.vcxproj.filters
+++ b/VGMTrans_WTL.vcxproj.filters
@@ -247,6 +247,12 @@
     <Filter Include="Source Files\windows\DxErr">
       <UniqueIdentifier>{f60aad48-6a14-4c0d-a400-60b77e9a16af}</UniqueIdentifier>
     </Filter>
+    <Filter Include="Source Files\formats\rsar">
+      <UniqueIdentifier>{97a0567d-3d55-4f8a-b33f-f133e474faf6}</UniqueIdentifier>
+    </Filter>
+    <Filter Include="Header Files\formats\rsar">
+      <UniqueIdentifier>{421ed46d-2c16-4746-9bca-ea6899534bdb}</UniqueIdentifier>
+    </Filter>
   </ItemGroup>
   <ItemGroup>
     <ClCompile Include="src\ui\windows\aboutdlg.cpp">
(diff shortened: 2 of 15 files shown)


mikelow:
This is going to take a while to review. Here's feedback on RSARScanner:

High level:
* First, isn't "RSAR" the name of the archive file format? You'll have to enlighten me on this, since I don't know much about the format and what systems/companies used it. The naming convention we use tries to make a distinction between the archive file format, and the name of the sound format itself. If the format has a known name, then we use that. Otherwise if it's say the common format Capcom used on the SNES, we call it CapcomSnes. Or say the format Nintendo provided in its N64 SDK, we'd call it NinN64. I know we're not consistent, but it's something we should strive for.
* There are a bunch of structs declared in the cpp file. They should be declared in the .h. Their names are also generic ("File") and which could cause conflicts given that we're not using namespaces  (not that I'm opposed, but that's a bigger change) or are too non-specific since they aren't nested.
* There are a bunch of static functions, and I don't see why they need to be static. At the very least, they should be methods of the RSARScanner class.
* A solution to both of the above problems would be to repurpose/rename your "Context" struct to something that represents the entire RSAR data structure (maybe just call it "RSAR") since, as it stands, you're storing most of the high-level parsed information in that struct. It would then make sense to nest inside of it all of the other structs it uses, and within them, their parsing methods that populate their data (or don't nest structs if that gets messy). I'm imagining each struct, including RSAR itself, has its own Parse() method.
* The RBNK struct similarly doesn't seem to belong in RSARInstrSet.cpp but in RSARScanner.cpp, since it's a nested piece of the RSAR data structure. I'm also confused that it retains pointers to an instrSet and sampColl. Why? I would recommend removing the creation of these VGMFiles outside of the Parse() method (see below)
* RSARSeq::Parse() violates the convention of having a format's Scanner handle the validation and creation of VGMFiles (sequences, instrsets, and sampcolls). It makes sense to me to have an RSEQ struct as child of the RSAR struct, but I would keep this separate from the VGMFile class that represents the sequence.
* So, if we take this approach of parsing the entire RSAR file into a RSAR struct, I think we can simplify RSARScanner::Scan() a bit. Conceptually, it does two things: It attempts to parse the RawFile as an RSAR archive. If this succeeds, it then iterates over the data collected to create the instrument sets, sample collections, and sequences.

A few low-level points:
* You should use RawFile::MatchBytes() in place of your MatchMagic() function.
* CheckBlock() would become a method of the RSAR struct.
* enum declarations should be multiline, just as structs are.

magcius:
* "BRSAR" is Binary Revolution Sound ARchive. It's part of a piece of middleware Nintendo provides to third parties, much like SSEQ on Nintendo DS. It's used by practically every game that uses sequences. I used "RSAR" since that's the first four bytes / magic of the file but I'm in no way attached to the name.
* Structs having generic names are exactly why I kept them locally scoped to the .cpp files. They closely map to the structures in the RSAR directly and are used in the lookup process -- RSAR files have sort of a really complex mapping between files stored in it where to find the bank used for a sequence, you have to look up the file ID, then the corresponding item in the group with that group ID, etc. Since it won't be used by anything other than the scanner, I didn't see the need to put them in the .h file, especially since RSARScanner.h is included by the root file, causing large compilation times whenever it changes.
* As for the RBNK struct, sample collections and instrument data are stored together in an "RBNK" struct, they're not independently mapped like in some other formats. It made sense for me to construct both of the sub-objects as part of one such "RBNK" object. What would you prefer here?

mikelow:
* Is this a Wii-specific format, then? In that case, just call it NinWii.
* Declaring structs in header files is a common C++ convention that we're following. So please do move the struct declarations there. My apologies for slow compile time. I do wonder if it's something to do with unnecessary header file dependence.
* I understand that RBNK couples together a pointer to the instrument data and the sample data, but I don't see how this is different from most other formats. The RBNK struct will capture this important association, which will be used to group the VGMFiles together into collections, but you don't have to make those VGMFiles part of the RBNK struct. The RSAR struct and all of its child data structures should effectively be dumb: they store the parsed data of the file and nothing more. They should not be aware of the VGMFile classes; only the original data found in the SRAR file (offsets, lengths, ids, etc). After you have populated the SRAR struct, then the second step of the RSARScanner::Scan() method will be to iterate over it to create the sample collections, instrument sets, and sequences and put them together into collections.

mikelow:
Actually, for pt #2, if I'm not mistaken putting the RSAR structs into their own header file and including it in RSARScanner.cpp would solve the compilation time problem. Feel free to do that. I guess you could just call it RSAR.h. We should eventually put all of these format code files into their own directories.

magcius:
BRSAR is actually one of two formats commonly used on the Wii. There's also a GameCube-era format, "JaiSeq" / "BMS", which is also used in some first-party Nintendo games for the Wii (Super Mario Galaxy 1/2, Twilight Princess).

Working on the other tasks now.

Is the FileRange struct (just an offset/length) pair something that would be wanted elsewhere? I imagine it could be useful as a common format.

magcius:
Updated PR for review.

* I didn't change the name RSAR, let's do that once we agree on a name (NinBRSAR? NinWiiRSAR?).
* I'm not a huge fan of making RBNK and such "pure", since it's possible for the sample collection to effectively be one of two types depending on whether an offset value is 0 or not. I used a enum tag here, but it's a lot more natural to just use a natural subclass. Additionally, I just like the code of the old bit better, especially the code that builds the collection in RSARScanner.
* I put FileRange in RSARScanner.h in a global header. I'm sorry. We have three options: namespace / prefix it, expand it to two variables everywhere, and make it a common utility.

Otherwise, I've tried to follow your suggestions as best as possible.

paulsapps:
You've got a build error:

RSARInstrSet.cpp:69:5: error: 
      constant expression evaluates to -723 which cannot be narrowed to type
      'uint16_t' (aka 'unsigned short') [-Wc++11-narrowing]
    -723, -722, -721, -651, -601, -562, -530, -503,

-723 obviously isn't unsigned :)

magcius:
yikes, I wonder why Visual Studio allows such nonsense

magcius:
Have you had any time to look at this?

MDHEXT:
if anyone is still monitoring this, can you add support for bcsar as well??????? its like brsar, but it is for 3ds. please add support.

sykhro:
Closing for inactivity

PR #87, opened 2017-02-11
Added drumkit/percussion support to the AkaoSnes format.

ElectronBlast:
The commit comment on bf25c3c has more details, but this adds percussion support to the AkaoSnes format (which was previously listed in the source as "TODO: percussion note"). As far as I've found, the percussion table is used in Chrono Trigger, Front Mission, and Front Mission - Gun Hazard.

Some percussion is implemented using the SPC700 white noise capability, and this commit doesn't fix that, but it does fix major parts of a lot of the songs from those games.

### DIFF
--- a/src/main/formats/AkaoSnesInstr.h
+++ b/src/main/formats/AkaoSnesInstr.h
@@ -11,11 +11,14 @@
 class AkaoSnesInstrSet:
     public VGMInstrSet {
  public:
+  static const uint32_t DRUMKIT_PROGRAM = (0x7F << 14);
+
   AkaoSnesInstrSet(RawFile *file,
                    AkaoSnesVersion ver,
                    uint32_t spcDirAddr,
                    uint16_t addrTuningTable,
                    uint16_t addrADSRTable,
+                   uint16_t addrDrumKitTable,
                    const std::wstring &name = L"AkaoSnesInstrSet");
   virtual ~AkaoSnesInstrSet(void);
 
@@ -28,6 +31,7 @@ class AkaoSnesInstrSet:
   uint32_t spcDirAddr;
   uint16_t addrTuningTable;
   uint16_t addrADSRTable;
+  uint16_t addrDrumKitTable;
   std::vector<uint8_t> usedSRCNs;
 };
 
@@ -57,6 +61,34 @@ class AkaoSnesInstr
   uint16_t addrADSRTable;
 };
 
+// *************
+// AkaoSnesDrumKit
+// *************
+
+class AkaoSnesDrumKit
+  : public VGMInstr {
+public:
+  AkaoSnesDrumKit(VGMInstrSet *instrSet,
+                  AkaoSnesVersion ver,
+                  uint32_t programNum,
+                  uint32_t spcDirAddr,
+                  uint16_t addrTuningTable,
+                  uint16_t addrADSRTable,
+                  uint16_t addrDrumKitTable,
+                  const std::wstring &name = L"AkaoSnesDrumKit");
+  virtual ~AkaoSnesDrumKit(void);
+
+  virtual bool LoadInstr();
+
+  AkaoSnesVersion version;
+
+protected:
+  uint32_t spcDirAddr;
+  uint16_t addrTuningTable;
+  uint16_t addrADSRTable;
+  uint16_t addrDrumKitTable;
+};
+
 // ***********
 // AkaoSnesRgn
 // ***********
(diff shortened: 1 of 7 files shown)


mikelow:
This looks great! @loveemu, being that you wrote the akaosnes code, have any input?

src/main/formats/AkaoSnesInstr.h
```diff
@@ -11,11 +11,14 @@
 class AkaoSnesInstrSet:
     public VGMInstrSet {
  public:
+  static const uint32_t DRUMKIT_PROGRAM = (0x7F << 14);
```
mikelow: So I was confused why you used this value instead of simply 128 (which would result in a bank select MSB 0 LSB 1), but now I _think_ I understand. You use this value because GM Drum Kits are on bank MSB 127 LSB 0, right? 

This makes me think we should just use MSB for bank changes and drop support for LSB altogether (always select LSB 0), which would obviate the need for that SF2File.cpp change. But let's keep what you have for now and I can update that later.

ElectronBlast: I've read a fair amount of conflicting information online about what the GM drum bank is, so I had to go look at the actual MIDI standard; it looks like the Wikipedia entry is correct that it's actually 120 (0x78)/0, not 127/0. You're correct in assuming that was my thinking behind that choice, yes, but it probably doesn't actually matter that much since it's not actually GM. Still, I wanted to make it easy to change since I wasn't 100% certain at the time. I can change it to 0x78 if you'd like a little bit of consistency with GM where possible anyway.

Actually, I started to do the same thing with the SuzukiSnes format locally, and in that one, it turns out that the transpose events actually affect the otherwise fixed percussion pitches. Given the similarity to AkaoSnes, that makes me think I need to go back and verify that the case isn't the same with the AkaoSnes format (some of the songs from Chrono Trigger have a transpose immediately before turning on percussion, and the two formats seem very similar). To my knowledge, there's no transpose equivalent in MIDI (not counting pitch bend, which I want to avoid in case actual pitch slide support is implemented), so if it turns out that's the case, the two alternatives I see are A) play the original patches from bank 0/0 at the correct pitch, or B) use the whole bank for the percussion sounds instead of just a single patch.

Option A would require that the instrument set loading phase were performed before the sequence loading phase (it's currently after), and would make the sequence loading phase dependent on the instruments so that the correct patch numbers and keys could be looked up for the percussion notes, whereas Option B could be just "every percussion instrument is played at exactly the same note unless transposed" and the keying could just be handled in the sound bank with the unity key.

I'm not sure which is preferable; having the two phases (instruments and sequence) be totally independent seems to be the way it's written right now, which tilts towards option B, but option A requires no bank switches. I'd welcome some input on which might be better (or if you've got a third option).

ElectronBlast: Also, regarding the MSB vs. LSB choice, I will say that the MIDI driver I'm using to play back with SF2 files outside of vgmtrans (BASSMIDI) treats the bank numbers in SF2 files as MSB values only and ignores LSB.

mikelow: I just dug into this a bit deeper to better understand how the drumkit spec works. First, I think I would forget about trying to conform to GM drumkit bank stuff because this format doesn't share much in common with GM, aside from the concept of condensing percussion instruments into a single patch. Still, point taken that we should switch to using MSB for bank switching.

I like your initial implementation of this. I like that you anchored the lower bound of the note indexes to middle c and adjusted the unityKey of each percussion instrument accordingly. 

The fact that a transpose event can be applied to drumkit notes is _really_ inconvenient. You are correct, to my knowledge, GM does not offer any transpose event. So we're faced with a difficult problem. In my mind, your Option A is the ideal solution, but it would require some significant refactoring and will add complexity to the core code. Interestingly, the fft driver has a similar problem where the instrument set needs information from the sequence, as some songs set ADSR envelopes via sequence events at the beginning of their tracks (but fixing that would lead to different exporting a different soundfont per song - and you can see how that's inelegant/confusing).

I'm not entirely opposed to Option B, but I don't like it. It will definitely result in some ugly, hard to decipher MIDI.

Maybe it's worth testing if pitch bend is ever being used on percussion notes in any sequence of the format. If it's not, I would consider that the best solution.

Finally, doing nothing for the moment is perfectly acceptable in terms of merging the code. Not supporting transpose on drumkits is better than having no drumkits.

ElectronBlast: So far, I haven't seen any percussion transpose outside the -2/+2 range, nor have I seen a percussion pitch slide, so those seem like good conditions for trying to hijack pitch bend to do the job, but I'll admit I haven't done a comprehensive survey.

Regarding not bothering with GM drumkit matching, I'm totally on board. It's the reason I didn't bother using AddPercNoteByDur() in the sequence - moving the notes to channel 9/10 would have broken up the MIDI channels matching up exactly to the DSP voices, and with an appropriate matching soundfont, there's no need anyway.

I've seen some similar sequence-driven ADSR in some AkaoSnes files, and yeah, that seemed like a can of worms.

In any case, I've poked at one of the Chrono Trigger SPC files and confirmed that the transpose applies to the drumkit notes there too, so it looks like this is indeed a thing for both AkaoSnes and SuzukiSnes.

I'm glad you're happy with this PR; you all have done some fine work with this project and it was interesting to me to dig into the audio guts of some old favorite games.

mikelow: Fortunately, MIDI does have an event for setting the pitch bend range, so we're not limited to the default -2/+2 semitones. Should be easy to rig up some test code to see what games use drum kits and if any of them use pitch bends.

Yeah, sequence-driven ADSR throws a wrench into things. The instrument set would have to be aware of the sequence, and we'd have to make a new instrument for each ADSR permutation. Then there's the issue of the "sustain rate" curve in SNES, Playstation, and QSound envelopes which isn't supported in SF2 and DLS, but I'll stop myself...

Anyway, the PR looks fantastic. Thanks for submitting it and thanks for the kind words. I'll go ahead and merge it pending any feedback from @loveemu.

loveemu:
Reviewed the changes. The implementation is very well written and I don't find any issues on it.

Drums make songs much listenable. I am so happy with this. Thank you for your contribution.

Issue #197, opened 2021-03-10
Playback problems with songs that use more than 16 MIDI channels


gocha:
**Describe the bug**
Some songs don't play correctly, some sounds are lost or different instruments are used. When a song uses more than 16 MIDI channels, the same MIDI channel can be shared by multiple tracks, resulting in a different sound than expected. As far as I can confirm, the sound output is similar to the result when I export MIDI+SF2 and play it back with foobar2000.

**To Reproduce**
Steps to reproduce the problem.
1. Load Castlevania - Dawn of Sorrow.nds
2. Start playing SDL_BGM_BOSS1_ (Evil Invitation).

**Expected behavior**
[Castlevania: Dawn of Sorrow OST - Evil Invitation - YouTube](https://www.youtube.com/watch?v=NUodpncVH-Q)
You will easily notice that the string ensemble on the first track has been replaced by crash cymbals on the 16th track.

> **To Reproduce**
> Steps to reproduce the problem.
> 1. Load Final Fantasy X: 415 Seymour Battle.minipsf2
> 2. Start playing. You'll notice that some of the first synth melody is missing.
> 3. (Additional step: if you stop and restart playback, the playback result will be different.)
> 
> https://twitter.com/loveemuchan/status/1369657585696055296
> 
> **Expected behavior**
> You will hear the synth sound correctly. In the last release (where DirectMusic is used), the synth will sound correct.

**Desktop (please complete the following information):**
 - OS: Windows 10 21H2
 - VGMTrans version: VGMTrans-55393d9c519ba8bad2563eb869a60afa51b53221-master-x86_64-Windows (Qt)
 - VGMTrans log: No errors or warnings
 - Game ROM or music dump SHA-256: I will write if necessary

**Information that may be relevant**
https://github.com/FluidSynth/fluidsynth/pull/677

sykhro:
After some digging, it seems like the bug is actually about SF2: the new playback system uses that instead of DLS. Something in the SynthFile -> SF2 conversion is going wrong. To make things worse, it also seems to be random; only random tracks from random games have this kind of problem. There's a whole realm of possibilities here: SF2 not being spec-compliant, numerical errors in conversion, data type mismatches.. it's going to take a while

sykhro:
On top of that, it also seems like our DLS files are not compliant either as some playback engines don't recognize them.

loveemu:
[...]

PR #291, opened 2022-01-29
RFC: New MusicPlayer playback logic

### LINKED ISSUE #264: Inconsistency in bank select strategy between MIDI data and in-app player
## Case 1

**Describe the bug**
In Qt build, some instruments will not be played or will be played as different instruments. This is caused by a bank strategy mismatch between MIDI data and BASSMIDI.

In general, SeqTrack outputs banks in the "mma" style (CC0*128+CC32) as described in Fluidsynth. Therefore, the MIDI data generated by VGMTrans usually selects a bank using CC32. On the other hand, BASSMIDI seems to expect CC0 as a bank select message. (This seemed to be the case even when XG Reset sysex was sent.)

This did not happen in the WTL version. This is probably because DirectMusic uses DLS, and DLS should be able to set the bank number in MSB/LSB.

**To Reproduce**
1. Load SNES Donkey Kong Country 2: Klubba's Reveille (dkq-23.spc)
2. The low brass sound is played by a square-like synth (some other instruments also have problems)

Later AKAOs (e.g. Final Fantasy IX) are also good for checking the performance of banked instruments.

## Case 2

The following problem used to occur with Fluidsynth, but no longer occurs after switching to BASSMIDI. Just be careful not to let it happen again in the process of fixing the problem above.

**Describe the bug**
The drum sounds of late SNES AKAO songs don't sound right, but are instead played by a saw-like synth or something.

In this case, the bank number of the drum is defined as `(127 << 14)`, so VGMTrans will output bank MSB=127 to the MIDI data.

**To Reproduce**
1. Load the Battle theme from Chrono Trigger (ct-112.spc), for example
    * In addition to Chrono Trigger, the same problem can be seen in Romancing SaGa 3 and Gun Hazard
2. The drum sounds such as snares are played by a saw-like synth

 - OS: Windows 10
 - VGMTrans version: e579b9c, master
 - Game ROM or music dump SHA-256: [Super Donkey Kong 2: Dixie & Diddy ~ SNESmusic.org](http://snesmusic.org/v2/profile.php?profile=set&selected=15924)
 - Game ROM or music dump SHA-256: [Chrono Trigger ~ SNESmusic.org](http://snesmusic.org/v2/profile.php?selected=554&profile=set)

### LINKED ISSUE #289: Qt/Fluidsynth: MIDI channel 10 does not select instruments in the same way as other channels
**Describe the bug**
In the Qt build, the long note of the cello ensemble at the beginning of Kaede Theme 1 in PS2 Onimusha: Warlords does not sound. I assume that since the notes are output to MIDI channel 10, I can't select and play the melodic instrument in the same way as the other channels.

That may seem obvious, but this problem does not occur in WTL builds.

**To Reproduce**
1. Load 111 Kaede Theme 1.psf2
2. Play

**Expected behavior**
That all channels be played, regardless of the MIDI channel. Or, to attempt to assign a melody instrument to any format but MIDI channel 10 (some formats explicitly bypass MIDI channel 10).

 - OS: Windows 10
 - VGMTrans version: 3e46443, master
 - Game ROM or music dump SHA-256: Written as needed

sykhro:
This replaces the current playback logic in MusicPlayer with a custom sequencer using `fluid_sequencer_t`.

## Description
The MIDI player was replaced with an instance of a sequencer. All the events from the MIDI file are mapped onto FluidSynth events and are scheduled to play at their initial tick.

All of that dynamic casting isn't pretty; it's not strictly necessary so it can be changed to static_cast, which is faster. I will do that once all the MIDI events are implemented and we don't risk mis-casting.

Pausing and seeking will be tricky to implement, as they require manually stepping the synth - and this is not supported when using the synth like we do.

The master volume event remains unimplemented as FluidSynth only has limited SysEx support.

## Motivation and Context
This new mechanism allows us to define custom channel mappings and to interact with the underlying synth directly,
letting us bypass limitations imposed by Standard MIDI.

## How Has This Been Tested?
By playing back various files.

## Types of changes

## TODO:
- [x] Implement MIDI events 
- [x] Implement proper bank select (fixes #264)
- [x] Improve the situation with #197
- [ ] ~~Implement global volume event~~
- [x] Bypass MIDI channel 10 (fixes #289)
- [ ] Implement pausing
- [ ] Implement seeking

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,9 +41,9 @@ jobs:
           sudo apt clean
           sudo apt update
           sudo apt install -y pkg-config qt515base qt515svg qt515wayland libjack-dev libsndfile1-dev libpulse-dev libglib2.0-dev ninja-build libgl1-mesa-dev
-          wget https://github.com/FluidSynth/fluidsynth/archive/v2.1.9.zip
-          unzip v2.1.9.zip
-          cd fluidsynth-2.1.9
+          wget https://github.com/FluidSynth/fluidsynth/archive/v2.2.5.zip
+          unzip v2.2.5.zip
+          cd fluidsynth-2.2.5
           mkdir build && cd build
           cmake -DCMAKE_C_COMPILER="gcc-10" -DCMAKE_CXX_COMPILER="g++-10" -Denable-network=off -Denable-ipv6=off -DCMAKE_INSTALL_PREFIX=/usr/ -DLIB_INSTALL_DIR=lib -GNinja ..
           cmake --build . --config "Release" --parallel
--- a/src/ui/qt/CMakeLists.txt
+++ b/src/ui/qt/CMakeLists.txt
@@ -1,7 +1,7 @@
 add_subdirectory("${PROJECT_SOURCE_DIR}/lib/gsl-lite"
                  "${CMAKE_CURRENT_BINARY_DIR}/gsl-lite" EXCLUDE_FROM_ALL)
 
-set(FluidSynth_FIND_VERSION "2.0.0")
+set(FluidSynth_FIND_VERSION "2.2.0")
 find_package(FluidSynth)
 if(FluidSynth_FOUND AND "${FluidSynth_VERSION}" MATCHES "2(.)*")
   message(STATUS "Using FluidSynth ${FluidSynth_VERSION}")
--- a/src/ui/qt/MusicPlayer.cpp
+++ b/src/ui/qt/MusicPlayer.cpp
@@ -386,10 +386,6 @@ void MusicPlayer::processMidiFile(std::unique_ptr<MidiFile> midi) {
     return;
   }
 
-  /* We want events in their absolute time order, sorted by priority */
-  std::stable_sort(std::begin(events), std::end(events), PriorityCmp());
-  std::stable_sort(std::begin(events), std::end(events), AbsTimeCmp());
-
   /* Default to 120 BPM */
   auto ppqn = midi->GetPPQN();
   fluid_sequencer_set_time_scale(m_sequencer.get(), 1000000.0 * (ppqn / 500000.0));
--- a/src/ui/qt/MusicPlayer.cpp
+++ b/src/ui/qt/MusicPlayer.cpp
@@ -442,11 +442,6 @@ void MusicPlayer::processMidiFile(std::unique_ptr<MidiFile> midi) {
         break;
       }
 
-      case MIDIEVENT_MASTERVOL: {
-        // todo: this needs to be done with some sort of callback to change the synth gain at a
-        // specific tick count
-      }
-
       case MIDIEVENT_MODULATION: {
         auto modulation = dynamic_cast<ModulationEvent *>(event);
         fluid_event_modulation(seq_event, channel, modulation->dataByte);
@@ -482,6 +477,7 @@ void MusicPlayer::processMidiFile(std::unique_ptr<MidiFile> midi) {
       default:
       /* These 3 events are not useful when listening to the track */
       case MIDIEVENT_ENDOFTRACK:
+      case MIDIEVENT_MASTERVOL:
       case MIDIEVENT_TEXT: {
         delete_fluid_event(seq_event);
         continue;
--- a/src/ui/qt/MusicPlayer.cpp
+++ b/src/ui/qt/MusicPlayer.cpp
@@ -229,25 +229,26 @@ void MusicPlayer::makeSequencer() {
 }
 
 bool MusicPlayer::playing() const {
-  return m_sequencer != nullptr;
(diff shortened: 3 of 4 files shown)


sykhro:
Cherry-picking the log-related commits to `master`

loveemu:
@sykhro Thanks for the attempt. I'll report what I noticed after playing various songs.

* SNES Chrono Trigger: Battle 1 (ct-112.spc): The first bass sample plays incorrectly in the synth sample. The MIDI track has multiple program changes on the same tick, and I would expect the later one to take precedence, but the player is probably prioritizing the earlier one.
* SNES Gokinjo Bouken Tai: Town (gbt-04.spc): A tempo version of the problem above. Due to the fact that there are two tempos at the same time, the song will play at a very slow tempo.
* DS Casltlevania Dawn of Sorrow: SDL_BGM_L2_BOSS_ (Dance of Illusions): The brass on the right side will be played about two semitones higher. This track is correctly detuned by pitch bend, and is expected to play at a pitch about 9.375 cents lower.
* SNES Mega Man X: Spark Mandrill (mmx-16.spc): The entire sound is lowered by about six semitones. There is no pitch bend in the MIDI data, but there is some RPNs (RPN:1 fine tuning for detune) and portamento, so I think those are involved.
* PS1 Final Fantasy 7: 110 Fighting.minipsf: When the number of channels in a song exceeds 16, the playback result will be corrupted between shared MIDI channels (#197). Currently, the MIDI file does not contain any meta-events to avoid channel conflicts, so if the player expects such meta-events to appear, this is the expected behavior.
* PS1 Star Ocean: The Second Story: 225 Mighty Blow.psf: When I press the play button, nothing happens. It's as if nothing is being pressed. (The same goes for FFX 415 Seymour Battle.minipsf2)
* SNES Donkey Kong Country 2: Klubba's Reveille (dkq-23.spc): The low brass notes of the melody (bank LSB=1, program=34) will be replaced by the sample notes of the synth. I am not sure, but I think it is most likely bank related.
* UI: The Pause button icon will not change to a Play button when I stop playing by pressing the spacebar.
* UI: The Pause button icon will not change to a Play button when the song reaches the end.

sykhro:
> * SNES Chrono Trigger: Battle 1 (ct-112.spc): The first bass sample plays incorrectly in the synth sample. The MIDI track has multiple program changes on the same tick, and I would expect the later one to take precedence, but the player is probably prioritizing the earlier one.
> 
> * SNES Gokinjo Bouken Tai: Town (gbt-04.spc): A tempo version of the problem above. Due to the fact that there are two tempos at the same time, the song will play at a very slow tempo.

Curious that this didn't happen with DirectMusic as we are sorting events in the same way. I feel like the easiest fix here is to scan the vector of events and simply discard the earlier event.

> DS Casltlevania Dawn of Sorrow: SDL_BGM_L2_BOSS_ (Dance of Illusions): The brass on the right side will be played about two semitones higher. This track is correctly detuned by pitch bend, and is expected to play at a pitch about 9.375 cents lower.

No idea why this would happen. Then again, fluidsynth doesn't support detuning to cents, only semitones, so in the event of a fix it would be stuck to -9 semitones.

> SNES Mega Man X: Spark Mandrill (mmx-16.spc): The entire sound is lowered by about six semitones. There is no pitch bend in the MIDI data, but there is some RPNs (RPN:1 fine tuning for detune) and portamento, so I think those are involved.

This is odd, FluidSynth does implement RPNs (as shown [here](https://github.com/FluidSynth/fluidsynth/wiki/FluidFeatures#nrpn-control-change-implementation-chart)). Another sorting issue, perhaps?

> PS1 Final Fantasy 7: 110 Fighting.minipsf: When the number of channels in a song exceeds 16, the playback result will be corrupted between shared MIDI channels (Qt/FluidSynth: Playback problems with songs that use more than 16 MIDI channels #197). Currently, the MIDI file does not contain any meta-events to avoid channel conflicts, so if the player expects such meta-events to appear, this is the expected behavior.

Yes, once we add these meta events the idea is to catch the realtime sysex and figure out how to map the channels.

> PS1 Star Ocean: The Second Story: 225 Mighty Blow.psf: When I press the play button, nothing happens. It's as if nothing is being pressed. (The same goes for FFX 415 Seymour Battle.minipsf2)

Noticed the same thing, but it looks like the files are missing the sample collection. There's a bug somewhere else..

> SNES Donkey Kong Country 2: Klubba's Reveille (dkq-23.spc): The low brass notes of the melody (bank LSB=1, program=34) will be replaced by the sample notes of the synth. I am not sure, but I think it is most likely bank related.

Not sure either, will check

> UI: The Pause button icon will not change to a Play button when I stop playing by pressing the spacebar.
UI: The Pause button icon will not change to a Play button when the song reaches the end.

Thanks for noticing, easy fixes. There's still the issue of how to implement pausing/seeking, with this setup it's not possible in the same way as it was with the standard midi player

PR #296, opened 2022-01-30
New player implementation using BASSMIDI

### LINKED ISSUE #289: Qt/Fluidsynth: MIDI channel 10 does not select instruments in the same way as other channels

sykhro:
Supersedes #291.

## Description
This replaces the existing player based on FluidSynth with a new implementation based on BASSMIDI.
It has all the same new features: pause, stop, seek, live progress bar. Bank select works out of the box.
It has none of the [issues mentioned](https://github.com/vgmtrans/vgmtrans/pull/291#issuecomment-1025061301) in the other implementation
It could solve the problem with #289 by adding a filter for MIDI events that routes them away from channel 10. This is will be a future bugfix. It replicates the current `MusicPlayer` API 1:1.

## Motivation and Context
FluidSynth is an excellent library but we ran into some roadblocks with its synth. BASSMIDI's softsynth (supposedly) has the SysEx we need to address #197. It can also work to implement in-app loops.
When it comes to platform support:
| Operating system | Architecture | Supported |
|-|-|-|
| Windows | x86_64 | Yes, uses WASAPI if available; otherwise it falls back to DirectSound. |
| Windows | aarch64 | No. Might change as Windows on ARM gains more support. For now, this means discarding our plans of releasing aarch64 VGMTrans on Windows. |
| Linux | x86_64, aarch64 | Yes, uses ALSA. ~If it turns out this is problematic, we can write a driver of sorts for PulseAudio in about 200 lines of code.~  Users will need `pulseaudio-alsa` or `pipewire-alsa` depending on their distribution. |
| *BSD | x86_64, aarch64 | Unknown, should be the same as Linux. |
| macOS | x86_64, Apple Silicon | Yes, uses CoreAudio. | 

BASS can also be used for playback of audio files of various kind, meaning that we could also make use of it to implement a sample player.  

## How Has This Been Tested?
By playing back various files. Testing  is welcome to identify new issues.

## Types of changes

## Checklist:


sykhro:
We have been greenlit - the project can use BASS(MIDI) without a license. Before merging I'd like to confirm there are no issues with the Linux builds.

PR #298, opened 2022-02-01
Address bank select issues

### LINKED ISSUE #11: [FEATURE REQ] Allow to choose bank select MSB/LSB by preference
Current VGMTrans converts program number 128 to bank MSB=0/LSB=1 program 0. This is arithmetically correct, but many editors/plugins seem to use bank MSB and just ignore bank LSB.

So I hope VGMTrans will swap the bank output.

*Reported by SOU1, thanks.

### LINKED ISSUE #264: Inconsistency in bank select strategy between MIDI data and in-app player

sykhro:
Fixes a long standing issue with the format of the bank selects we use.

## Description
Details are in #264.
This adds support for GS and MMA style bank selection. Perhaps GM and XG could be handy too?

## Motivation and Context
Fixes #264, fixes #11

## How Has This Been Tested?
By playing back various songs and checking the MIDI output.

Appearance of the selector (only tested on macOS)
image

## Types of changes

## Checklist:

### DIFF
--- a/src/main/SeqTrack.cpp
+++ b/src/main/SeqTrack.cpp
@@ -1203,8 +1203,13 @@ void SeqTrack::AddProgramChange(uint32_t offset,
 void SeqTrack::AddProgramChangeNoItem(uint32_t progNum, bool requireBank) {
   if (readMode == READMODE_CONVERT_TO_MIDI) {
     if (requireBank) {
-      pMidiTrack->AddBankSelect(channel, (progNum >> 14) & 0x7f);
-      pMidiTrack->AddBankSelectFine(channel, (progNum >> 7) & 0x7f);
+      if (auto style = ConversionOptions::the().GetBankSelectStyle();
+          style == BankSelectStyle::GS) {
+        pMidiTrack->AddBankSelect(channel, (progNum >> 7) & 0x7f);
+      } else if (style == BankSelectStyle::MMA) {
+        pMidiTrack->AddBankSelect(channel, (progNum >> 14) & 0x7f);
+        pMidiTrack->AddBankSelectFine(channel, (progNum >> 7) & 0x7f);
+      }
     }
     pMidiTrack->AddProgramChange(channel, progNum & 0x7f);
   }
@@ -1212,8 +1217,13 @@ void SeqTrack::AddProgramChangeNoItem(uint32_t progNum, bool requireBank) {
 
 void SeqTrack::AddBankSelectNoItem(uint8_t bank) {
   if (readMode == READMODE_CONVERT_TO_MIDI) {
-    pMidiTrack->AddBankSelect(channel, bank / 128);
-    pMidiTrack->AddBankSelectFine(channel, bank % 128);
+    if (auto style = ConversionOptions::the().GetBankSelectStyle();
+        style == BankSelectStyle::GS) {
+      pMidiTrack->AddBankSelect(channel, bank & 0x7f);
+    } else if (style == BankSelectStyle::MMA) {
+      pMidiTrack->AddBankSelect(channel, bank >> 7);
+      pMidiTrack->AddBankSelectFine(channel, bank & 0x7f);
+    }
   }
 }
 
--- a/src/main/SeqTrack.cpp
+++ b/src/main/SeqTrack.cpp
@@ -1375,7 +1375,7 @@ bool SeqTrack::AddLoopForever(uint32_t offset, uint32_t length, const std::wstri
   }
   else if (readMode == READMODE_FIND_DELTA_LENGTH) {
     deltaLength = GetTime();
-    return (this->foreverLoops < ConversionOptions::GetNumSequenceLoops());
+    return (this->foreverLoops < ConversionOptions::the().GetNumSequenceLoops());
   }
   return true;
 
--- a/src/main/VGMColl.cpp
+++ b/src/main/VGMColl.cpp
@@ -7,6 +7,7 @@
 #include "VGMRgn.h"
 #include "ScaleConversion.h"
 #include "Root.h"
+#include "Options.h"
 
 using namespace std;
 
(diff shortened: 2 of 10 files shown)


src/main/SeqTrack.cpp
```diff
@@ -1203,17 +1203,27 @@ void SeqTrack::AddProgramChange(uint32_t offset,
...
   }
 }
 
 void SeqTrack::AddBankSelectNoItem(uint8_t bank) {
   if (readMode == READMODE_CONVERT_TO_MIDI) {
-    pMidiTrack->AddBankSelect(channel, bank / 128);
-    pMidiTrack->AddBankSelectFine(channel, bank % 128);
+    if (auto style = ConversionOptions::the().GetBankSelectStyle();
+        style == BankSelectStyle::GS) {
+      pMidiTrack->AddBankSelect(channel, bank >> 7);
```
loveemu, line 1222:
```suggestion
      pMidiTrack->AddBankSelect(channel, bank & 0x7f);
```

loveemu:
Thanks for the nice changes. ❤️ 

* MIDI+DLS playback in the "GS" style is problematic because the bank options are not reflected in the DLS bank numbers. Also, since GS style is internally selected as the default value, the DirectMusic output of the WTL version is affected by this.

sykhro:
> Thanks for the nice changes. ❤️
> 
>     * MIDI+DLS playback in the "GS" style is problematic because the bank options are not reflected in the DLS bank numbers. Also, since GS style is internally selected as the default value, the DirectMusic output of the WTL version is affected by this.
> 
>     * UI: Initially, none of the bank styles are selected.

I have hard-coded WTL to use the MMA-style bank-select, so that it will keep working as it always has (this means it will continue outputting MIDIs with MMA bs, but it has been doing so for years so I'd say it's fine to keep it this way).
As for the UI, I can't reproduce; the code for the default selection is [here](https://github.com/vgmtrans/vgmtrans/pull/298/files#diff-33e6393e424821922e6b7e8998740cc8b7898960589eb8ce57019fb006c1edb9R39)

loveemu:
Thank you. It worked.

Is it outside the scope of this PR to reflect the bank number preference in the DLS export?
I'm sorry for making you do so many things. I just want to keep track of everything I notice, so there's no need to rush.

sykhro:
> Thank you. It worked.
> 
> Is it outside the scope of this PR to reflect the bank number preference in the DLS export? I'm sorry for making you do so many things. I just want to keep track of everything I notice, so there's no need to rush.

No worries! I'm not sure what you mean, is it something that needs to change in DLSFile?

sykhro:
I did it, not 100% sure if it is correct

src/main/VGMColl.cpp
```diff
@@ -196,7 +197,19 @@ bool VGMColl::MainDLSCreation(DLSFile &dls) {
...
+      * The ulBank field follows this structure:
+      * F|00000000000000000|CC0|CC32
+      * where F = 0 if the instrument is melodic, 1 otherwise
+      * (length of each CC is 7 bits, obviously)
+      */
+      if (auto bs = ConversionOptions::the().GetBankSelectStyle(); bs == BankSelectStyle::GS) {
+        bank_no = bank_no << 7;
+      } else if (bs == BankSelectStyle::MMA) {
+        bank_no = ((bank_no / 128) << 7) | bank_no;
+      }
```
loveemu, line 211:
```suggestion
      auto bank_no = vgminstr->bank;
      /*
      * The ulBank field follows this structure:
      * F|00000000000000000|CC0|0|CC32
      * where F = 0 if the instrument is melodic, 1 otherwise
      * (length of each CC is 7 bits, obviously)
      */
      if (auto bs = ConversionOptions::the().GetBankSelectStyle(); bs == BankSelectStyle::GS) {
        bank_no &= 0x7f;
        bank_no = bank_no << 8;
      } else if (bs == BankSelectStyle::MMA) {
        const uint8_t bank_msb = (bank_no >> 7) & 0x7f;
        const uint8_t bank_lsb = bank_no & 0x7f;
        bank_no = (bank_msb << 8) | bank_lsb;
      }
```

The incorrect description of `ulBank` confused me a lot. 😞

With this change, I believe `VGMInstr` will have the same internal MMA-style bank number as `VGMSeq`, and map it properly during DLS export.

Before the change, the bank number of `VGMInstr` used the same format as the bank number of DLS. However, upon investigation, it turned out that only a small bank number was actually being used, and only `AkaoSnesDrumKit` was intentionally trying to use bank MSB.

I will fix bank mapping in AkaoSnesDrumKit in another comment.

loveemu: I wasn't sure where best to comment, so I'll post a diff here to fix the bank number for AkaoSnesDrumKit.

```diff
 src/main/formats/AkaoSnesInstr.cpp | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/main/formats/AkaoSnesInstr.cpp b/src/main/formats/AkaoSnesInstr.cpp
index a0d9a59..04ba393 100644
--- a/src/main/formats/AkaoSnesInstr.cpp
+++ b/src/main/formats/AkaoSnesInstr.cpp
@@ -154,7 +154,7 @@ AkaoSnesDrumKit::AkaoSnesDrumKit(VGMInstrSet *instrSet,
                                  uint16_t addrADSRTable,
                                  uint16_t addrDrumKitTable,
                                  const std::wstring &name) :
-  VGMInstr(instrSet, addrTuningTable, 0, ((programNum >> 6) & 0x7F00) | ((programNum >> 7) & 0x7F), programNum & 0xFF, name), version(ver),
+  VGMInstr(instrSet, addrTuningTable, 0, programNum >> 7, programNum & 0x7F, name), version(ver),
   spcDirAddr(spcDirAddr),
   addrTuningTable(addrTuningTable),
   addrADSRTable(addrADSRTable),
 src/main/formats/AkaoSnesInstr.h | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/main/formats/AkaoSnesInstr.h b/src/main/formats/AkaoSnesInstr.h
index 0b48c3c..8b9d259 100644
--- a/src/main/formats/AkaoSnesInstr.h
+++ b/src/main/formats/AkaoSnesInstr.h
@@ -11,7 +11,7 @@
 class AkaoSnesInstrSet:
     public VGMInstrSet {
  public:
-  static const uint32_t DRUMKIT_PROGRAM = (0x7F << 14);
+  static const uint32_t DRUMKIT_PROGRAM = (0x7F << 7);
 
   AkaoSnesInstrSet(RawFile *file,
                    AkaoSnesVersion ver,
```

Before the change, bank MSB=127 was used, but after the change, bank LSB=127 will be used for the MMA style. It might be possible to abstract "bank for drums" so that it always prefers bank MSB=127, but for now we simply focused on fixing the bank number mismatch.

sykhro: Thank you for the review, will add this.

PR #328, opened 2022-10-27
[NinSnes] Don't increment the timer after EVENT_END ($00 VCMD)

### LINKED ISSUE #155: Extra rests are being added to exported MIDIs (Terranigma and other N-SPC games?)
**Describe the bug**
As the title says, there is an annoying issue when exporting MIDI files from Terranigma SPCs as well as some other N-SPC games (I haven't tried many) like Super Metroid, Kirby's Dreamland 3 and Kirby Super Star.

Basically, what happens is that the program will, for some reason, add a rest every so often at the start of a measure, which ends up pushing all the notes forward a bit. The way it does it seems to be different for every game, however. This is what I've found so far:

- **Terranigma**: The program will add a 2-tick long rest starting from measure 8. It adds another at measures 34, 38, 42, 47, 50, 52... I can't draw any pattern, unfortunately. Tested with Dark Side Field and Final Battle.
- **Super Metroid**: Adds a 2-tick long rest starting from measure 3 in a 4-4-8 pattern? (3, 7, 11, 19, 23, 27...). Tested with both Brinstar songs.
- **Both Kirby games**: The added 2-tick rests seem to start from measure 2 onward. It doesn't seem to follow a pattern, as it's very erratic (sometimes a rest is added every 6 measures, sometimes every 7, sometimes every other measure, etc). Tested with Grassland 1 (KDL3) and Gourmet Race (KSS).

It's important to note that the rests stack with each other, so when a new one is added it becomes a 4-tick rest and it increases every time. Also, by tick I mean the shortest note possible in a MIDI file (1/192 of a note, I think?).

**To Reproduce**
Export a MIDI file from Terranigma or the aforementioned N-SPC games, but the bug might affect other games as well, seeing that all of the ones I tested are recognized as "NinSnes" by the program.

**Expected behavior**
MIDIs without those added rests.

**Desktop:**
 - OS: Windows 10
 - VGMTrans version: Interim Build 2017-02-25 (I can't compile the program to test the newer code updates...)

KungFuFurby:
## Description

Had to create a new track variable that will skip incrementing the timer when set to true. The timer is incremented when all tracks are done processing, which EVENT_END forces to happen. However, the timer is then incremented, and thus incurs a one tick delay. This normally does not happen (Yoshikazu Yao's later builds, not yet supported, is a possible exception because there is some sort of two-pass system in there, but it could also be caused by lag, which vgmtrans does not need to emulate), as I recall that the next section is then immediately processed in at least some of the cases.

## Motivation and Context

This merge request closes #155.

## How Has This Been Tested?

My test environment is:
* Operating System+version: macOS Catalina 10.15.7
* VGMTrans version: 1.1

This is my test case that I used (mostly because I had used a low enough tempo that I could clearly hear the tick delay... and it's also my own creation at that, straight out of the original source Kankichi-kun variant):
* Game ROM/music dump name and sha256: [SoftWave.spc](https://github.com/vgmtrans/vgmtrans/files/9863639/SoftWave-000.spc.zip) using original N-SPC/Kankichi-kun source build, SHA256 is 23d439bfd3c6d9696b6fd6b1817071347ec7a2de87d8ed53ac9dbe11381b51cc

## Types of changes

## Checklist:

### DIFF
--- a/src/main/VGMSeq.cpp
+++ b/src/main/VGMSeq.cpp
@@ -27,6 +27,7 @@ VGMSeq::VGMSeq(const string &format, RawFile *file, uint32_t offset, uint32_t le
       bAlwaysWriteInitialMono(false),
       bAllowDiscontinuousTrackData(false),
       bLoadTickByTick(false),
+      bIncTickAfterProcessingTracks(true),
       initialVol(100),                    //GM standard (dls1 spec p16)
       initialExpression(127),            //''
       initialReverb(40),                //GM standard
@@ -215,8 +216,10 @@ void VGMSeq::LoadTracksMain(long stopTime) {
         itrSlider = itrNextSlider;
       }
 
-      time++;
-
+      if (bIncTickAfterProcessingTracks == true) {
+        time++;
+      }
+      bIncTickAfterProcessingTracks = true;
       if (readMode == READMODE_CONVERT_TO_MIDI) {
         for (uint32_t trackNum = 0; trackNum < nNumTracks; trackNum++) {
           if (aTracks[trackNum]->pMidiTrack != NULL) {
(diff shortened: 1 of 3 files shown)


sykhro:
Thank you for your contribution!

Issue #304, opened 2022-04-16
libbassmidi.so is not installed when compiling


ettolrach:
**Describe the bug**
After compiling the program and installing it with the install target, it doesn't install libbassmidi.so and thus cannot start.

**To Reproduce**
Steps to reproduce the problem.
1. Compile the program using the following AUR PKGBUILD:
```
# Maintainer: Charlotte Ausel <ettolrach@disroot.org>

_pkgname=vgmtrans
pkgname=$_pkgname-git
# Use the latest commit's date and then the number of revisions using the same command as on line 19.
pkgver=20220206.r1116.8fa6f4a
pkgrel=1
pkgdesc="Converter for sequenced videogame music"
arch=("x86_64")
url="https://github.com/vgmtrans/vgmtrans"
license=('ZLIB')
depends=("qt5-base" "fluidsynth" "qt5-svg" "minizip")
makedepends=("qt5-tools" "cmake" "git")
optdepends=("pulseaudio-alsa: to use PulseAudio instead of ALSA" "pipewire-alsa: to use PipeWire instead of ALSA")
source=("$_pkgname::git+https://github.com/vgmtrans/vgmtrans")
sha256sums=("SKIP")

pkgver() {
	cd "$srcdir/$_pkgname"
	printf "%s.r%s.%s" "$(git show -s --format="%cd" --date=short HEAD)" "$(git rev-list --count HEAD)" "$(git rev-parse --short HEAD)" | sed 's/-//g'
}

build() {
	cmake -B build -S "vgmtrans" -DCMAKE_BUILD_TYPE="None" -DCMAKE_INSTALL_LIBDIR=lib -DCMAKE_INSTALL_PREFIX="/usr" -Wno-dev
	cmake --build build --target vgmtrans --parallel $(nproc)
}

package() {
    DESTDIR="$pkgdir" cmake --install build
	#install -Dm755 "build/bin/vgmtrans" "$pkgdir/usr/bin/$_pkgname"
	#install -Dm644 "LICENSE.txt" "$pkgdir/usr/share/licenses/$_pkgname/LICENSE"
	#install -Dm644 "src/ui/qt/resources/VGMTrans.desktop" "$pkgdir/usr/share/applications/VGMTrans.desktop"
	#install -Dm644 "src/ui/qt/resources/vgmtrans.png" "$pkgdir/usr/share/icons/hicolor/512x512/apps/vgmtrans.png"
}

```

**Expected behavior**
[...]

PR #354, opened 2023-04-14
DLSFile: Add missing header

TheComputerGuy96:
This fixes a compile error (at least on my Linux system with GCC)

I initially added this in the AUR package but with a note to upstream this :frog:

### DIFF
--- a/src/main/DLSFile.cpp
+++ b/src/main/DLSFile.cpp
@@ -6,7 +6,6 @@
 
 #include "pch.h"
 #include <algorithm>
-#include <memory>
 #include <numeric>
 #include "DLSFile.h"
 #include "VGMInstrSet.h"
--- a/src/main/DLSFile.h
+++ b/src/main/DLSFile.h
@@ -7,6 +7,7 @@
 #pragma once
 
 #include "common.h"
+#include <memory>
 #include "RiffFile.h"
 
 struct Loop;


sykhro:
Thanks for your contribution, will merge once CI passes. If you're packaging for Arch, you might want to take a look at #304

TheComputerGuy96:
@sykhro I'm currently manually copying the libbass stuff in the AUR package: 
```sh
	install -Dm644 "../lib/bass/libbass.so" "${pkgdir}/usr/lib/${_pkgname}/libbass.so"
	install -Dm644 "../lib/bass/libbassmidi.so" "${pkgdir}/usr/lib/${_pkgname}/libbassmidi.so"
```

I might make another PR for copying that in CMake, but distros with an unique file structure (like NixOS) might be a bit problematic

I also have a weird segfault that I can't seem to figure out (might make an issue for that later as well)

BTW the FreeBSD 12.3 failure seems unrelated to this PR 🐸

sykhro:
Yeah, don't worry abour FreeBSD. The Linux job is stuck because GitHub has removed 18.04, which is annoying because Canonical supports it until the end of May. I'll upgrade to the closest LTS but `linuxdeplotqt` might not be happy

TheComputerGuy96:
@sykhro My next PR will likely be cleaning up the CI stuff 🐸 

Also I don't think VGMTrans works properly on FreeBSD because of no BSD support for BASS libraries (which are a requirement)

PR #359, opened 2023-04-29
Simplify ReadVarLen

Meerkov:
The previous code was really confusing and redundant. I fixed it.

The code is equivalent, but much less confusing.

What it does now:
Uses a while loop to check while if the 0x80 flag is set (variable length number). If so, left shift by 7, and then add the current 7 bits ( & 0x7F).

What it used to do:
The same thing, except it unrolled the first layer of the loop.


Meerkov:
Actually It's not quite right still.

Meerkov:
Ok It should be right now. Please take a look, thanks.

Meerkov:
Actually I clearly was confused when writing this last month, and forgot what I was doing. Closing, and I'll return to this later.

PR #360, opened 2023-04-29
Simplify ReadVarLen

Meerkov:
Simplify ReadVarLen

The previous code was really confusing and redundant. I fixed it.

Hopefully it's more clear to you as well.

PR #364, opened 2023-05-25
Fix SeqTrack::ReadVarLen

khang06:
#360 had a mistake with operator precedence between & and ==. `if(c & 0x80 == 0)` is always false because it's basically equivalent to `if(c & (0x80 == 0))`.


sykhro:
Thanks! This is exactly why I'm wary of PRs that rewrite small portions of working code, since we have no test suite. Going forward we will probably reject those.

PR #368, opened 2023-06-26
Add support for more CPS formats

mikelow:
**NOTE: I see we're still doing the WTL build (yech!). I'll update accordingly when I have time to install VC++ (blech).**

First up, thank you skyhro and loveemu for keeping this project alive!

I'm pulling in some of the non-UI code from my very old `ui_refactor` branch. First up is updates to the QSound driver, which I have renamed to CPS. The code may be a little sloppy, but I don't think it's worse than what's already there, and I'd like to make sure it doesn't get lost. The conversion is far from perfect, but it's a start.

Among the changes:
* rename QSound driver to CPS
* add support for CPS3 sequence/instr/samp formats
* add support for CPS1 sequences
* add CPS3 decryption to MAMELoader

This also adds support for a handful of Capcom games that run on Sony ZN-1 and ZN-2 hardware, namely:
* Rival Schools
* Star Gladiator
* Strider 2
* Street Fighter EX

Some CPS3 games and ZN-1/2 games are missing from mame_roms.xml (ex: Red Earth). I probably ran into difficulty with these games, but honestly I just don't remember.

Both "CPS" and "QSound" are misnomers, since the driver is used in games that don't use either hardware 🤷.


sykhro:
Thanks for your PR! I’ll check it out asap.

> Are we still supporting compilation on Visual Studio C++? I did not update the vcproj file in this PR, as I don't have an install handy. It would be great to migrate off of it now that we have functional Qt support.

I would love to drop the old windows code, but the playback code in Qt doesn’t work as well. I’m planning on going back to fluidsynth and rewriting it completely

mikelow:
FYI, I'll install VC++ to update the windows project file.

What are the problem points with the Qt playback code vs Window/DirectMusic? Hadn't noticed any issues, but I wasn't looking for it.

sykhro:
> FYI, I'll install VC++ to update the windows project file.
> 
> What are the problem points with the Qt playback code vs Window/DirectMusic? Hadn't noticed any issues, but I wasn't looking for it.

DirectMusic supports MIDI streams with more than 16 channels out of the box. FluidSynth doesn’t. We tried BASS for a better feature set, but that one doesn’t seem to cut it either and it’s closed source. At this point the only way is to use fluidsynth to implement a player that uses multiple midi synths to achieve that (the same way foo_midi did it)

mikelow:
Yes, the 16 channel limit is the worst. I thought I had solved that limitation with fluidsynth, albiet not gracefully. Is my memory faulty? Or were you looking for a better solution? What are the issues with BASS (aside from being closed source)? Didn't notice any problems.

My dream would be to migrate to using a virtual instrument standard like VST. That would pave the way for non-sampled synth playback.

sykhro:
> Yes, the 16 channel limit is the worst. I thought I had solved that limitation with fluidsynth, albiet not gracefully. Is my memory faulty? Or were you looking for a better solution? What are the issues with BASS (aside from being closed source)? Didn't notice any problems.
> 
> My dream would be to migrate to using a virtual instrument standard like VST. That would pave the way for non-sampled synth playback.

We were using the default `fluid_player_t`, which is a standarrd MIDI player and is capped to 16 channels.
We migrated to BASS because building fluidsynth on windows was painful (it depends on GLib) and it seemed like BASS suppported the same out of the box.

As for the issues, see #197. It might have been fixed with 0907dc4216c4b51e93af1fec452c1e9f66af7f66, not sure though.

VST support would be sweet, quite a big task though :)

sykhro:
LGTM, needs a rebase on `master` to let linux ci run

PR #369, opened 2023-06-26
Add reverb per instrument, and bug fixes

mikelow:
Improvements and bug fixes, many from ui_refactor branch:

* Add reverb percent as a param to VGMInstr and SynthInstr
* NDS Instruments don't use reverb.
* Add hold_time to SynthArt and VGMRgn, and add corresponding blocks to DLSFile and SF2File
* Fix SF2 files not properly handling sample fine tune values where the # cents didn't fit in a single signed byte. Now modifies both byOriginKey and chCorrection.
* Fix a bug in MAMELoader that was causing crashes when attempting to load .zip files with filenames > 10 chars


sykhro:
Rebased onto latest master to check if everything compiles correctly with the Ubuntu CI update. Will merge once done.

PR #373, opened 2023-07-13
Improve CPS format, add portamento time, portamento control, and small cleanup

mikelow:
I've made changes to the CPS format to better (perhaps even properly) handle portamento, and use a simple 8bit pitch bend range value as FluidSynth ignores the finetune LSB byte. In the process:

* I added support for Portamento Time LSB controller events.
* I added support for Portamento Control events (even though I wound up not using them)
* I made some small changes related to monophonic tracks

There's just one change at the bottom of MidiFile.cpp that I was slightly concerned might cause issues - there's conditional logic to not record prevKey when writing notes on a monophonic MidiTrack. However, I don't see other formats even marking the monophonic flags for tracks or seqs, and I'm pretty sure I was being dumb when I originally wrote that code. At any rate, the change there is needed to allow notes to slightly overlap, enabling a legato portamento slide.

### DIFF
--- a/src/main/MidiFile.cpp
+++ b/src/main/MidiFile.cpp
@@ -349,6 +349,18 @@ void MidiTrack::InsertPortamentoTime(uint8_t channel, uint8_t time, uint32_t abs
   aEvents.push_back(new PortamentoTimeEvent(this, channel, absTime, time));
 }
 
+void MidiTrack::AddPortamentoTimeFine(uint8_t channel, uint8_t time) {
+  aEvents.push_back(new PortamentoTimeFineEvent(this, channel, GetDelta(), time));
+}
+
+void MidiTrack::InsertPortamentoTimeFine(uint8_t channel, uint8_t time, uint32_t absTime) {
+  aEvents.push_back(new PortamentoTimeFineEvent(this, channel, absTime, time));
+}
+
+void MidiTrack::AddPortamentoControl(uint8_t channel, uint8_t key) {
+  aEvents.push_back(new PortamentoControlEvent(this, channel, GetDelta(), key));
+}
+
 void MidiTrack::AddMono(uint8_t channel) {
   aEvents.push_back(new MonoEvent(this, channel, GetDelta()));
 }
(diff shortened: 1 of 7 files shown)


src/main/MidiFile.cpp
```diff
@@ -698,25 +728,15 @@ NoteEvent::NoteEvent(MidiTrack *prntTrk,
...
-//}
-
 uint32_t NoteEvent::WriteEvent(vector<uint8_t> &buf, uint32_t time) {
   WriteVarLength(buf, AbsTime - time);
   if (bNoteDown)
     buf.push_back(0x90 + channel);
   else
     buf.push_back(0x80 + channel);
 
-  if (prntTrk->bMonophonic && !bNoteDown)
```
mikelow, line 713: This is the line referred to in the description.

sykhro: IIRC the only format in the that makes use of the monophonic flag(s) is CPS, and  the default constructor sets that to false. So this should be fine

src/main/formats/CPSSeq.cpp
```diff
@@ -210,12 +202,19 @@ bool CPSSeq::PostLoad() {
...
+          // A vibrato event, it will provide us the frequency range of the lfo
           if (marker->name == "vibrato") {
             vibratoCents = vibrato_depth_table[marker->databyte1] * (100 / 256.0);
-            pitchbendRange = (int) ceil(vibratoCents + fmtPitchBendRange);    //+50 cents to allow for pitchbend values, which range -50/+50
+            pitchBendRangeMSB = static_cast<uint8_t>(ceil(static_cast<double>(vibratoCents + fmtPitchBendRange) / 100.0));
+            pitchbendRange = pitchBendRangeMSB * 100;
             printf("Vibrato byte: %X  Vibrato cents: %f  Converted to range: %d in cents\n", marker->databyte1, vibratoCents, pitchbendRange);
-            track->InsertPitchBendRange(channel, pitchbendRange / 100, pitchbendRange % 100, curTicks);
 
+            // Fluidsynth does not support pitch bend range LSB, so we'll round up the value to the
```
sykhro, line 213: Did you mean BASSMIDI or is this actually an issue with the fluidsynth player?

sykhro (approved):
Thanks for the patch! I might make use of this for other formats.

src/main/formats/CPSSeq.cpp
mikelow: FluidSynth. I'm implementing playback via a VST host using the cross-platform JUCE framework. The virtual instrument I'm using is basically a wrapper around fluidsynth. I've hackily worked around the 16 track limitation and playback works well.

The move to using a VST that we control will allow us to solve the limitations of the SoundFont/DLS formats and expand into FM synthesis down the road. For example, related to this commit, I added a new "portamento time mode" parameter to the fluidsynth fork utilized by the VST. This lets the VST interpret portamento time as a fixed rate of pitch change over time, which the CPS format requires, rather than a fixed duration of time for the slide (regardless of from/to keys).

Another relatively quick win would be adding sustain rate, used by SNES and PSX hardware and also the CPS software format, but missing from the SoundFont2 and DLS specs.

The plugin and fluidsynth forks are under my personal account, but I intend to move them into vgmtrans. The work I'm doing VGMTrans side is in the `juce` branch. I can provide build instructions if you're interested.

sykhro: Oh, that sounds great. At that point the goal is making the VST also available for people doing remixes/midi editing, I imagine. Cool stuff :)

PR #374, opened 2023-07-14
Add SysexEvent class and update CPS portamento handling

mikelow:
* Add SysexEvent class to simplify defining sysex midi events
* Update Reset events to use SysexEvent
* Add PortamentoTimeMode event for usage in upcoming VST playback
* Add portamentoTimeMode property to VGMSeq
* Improve portamento implementation again in CPS format
* delete midi in VGMSeq destructor
* rename MastVolEvent -> MasterVolEvent

### DIFF
--- a/src/main/formats/CPSTrackV1.cpp
+++ b/src/main/formats/CPSTrackV1.cpp
@@ -65,7 +65,7 @@ bool CPSTrackV1::ReadEvent(void) {
         }
         else if (key != prevTieNote) {
           AddNoteOn(beginOffset, curOffset - beginOffset, key, 127, L"Note On (tied)");
-          InsertNoteOffNoItem(prevTieNote, GetTime()+2);
+          InsertNoteOffNoItem(prevTieNote, GetTime());
         }
         else
           AddGenericEvent(beginOffset, curOffset - beginOffset, L"Tie", L"", CLR_NOTEON);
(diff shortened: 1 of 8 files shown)


src/main/formats/CPSTrackV1.cpp
```diff
@@ -74,16 +74,16 @@ bool CPSTrackV1::ReadEvent(void) {
       }
       else {
         if (bPrevNoteTie) {
-          AddPortamentoNoItem(false);
           if (key != prevTieNote) {
-            AddNoteOffNoItem(prevTieNote);
+            InsertNoteOffNoItem(prevTieNote, GetTime()+2);
```
sykhro, line 78: Why is this +2?

mikelow: Good Question. Kind of a hack. The purpose behind adding a constant value is to force synthesizers to render the portamento. By default, most only perform a portamento glide for a legato transition between notes, as in: I press a second note while still holding the first.

So why +2? Because I want compatability with common DAWs and Cubase was cutting down time resolution, so a +1 was effectively a noop. Again - hack.

Best solution in my mind would be to remove the constant and get the upcoming virtual instrument to work without it. You can set the [portamento mode](https://www.fluidsynth.org/api/group__channel__setup.html) in fluid synth via fluid_synth_set_portamento_mode() and pass in a constant to trigger on EACH_NOTE instead of LEGATO_ONLY, but I was having some difficulty getting it to trigger. I need to revisit this.

sykhro: If you're not planning on touching this again soon, I would be fine with a short explanation comment above it. Otherwise, LGTM

mikelow: Ok, I removed the constant value altogether. It seems my fix at the fluidsynth level by setting the portamento mode to each note via fluid_synth_set_portamento_mode() was working after all, but was probably masked by a slew of other complications I had to solve.

PR #375, opened 2023-07-18
Compile fixes

TheComputerGuy96:
This allows VGMTrans to compile again on my Arch Linux system (these are pretty trivial changes and so shouldn't really break anything) 🐸

I also have another PR coming that will fix an audio playback regression for SNES formats (maybe other ones too)


sykhro:
Thanks, will merge once CI passes. What compiler (+ version) are you using?

PR #376, opened 2023-07-18
SNES music regression fix

TheComputerGuy96:
Here's the other PR that I promised to do to fix the SNES music playback

Note: I'm not sure if the channel parameter change fixes anything (it definitely won't break anything though)

### DIFF
--- a/src/main/MidiFile.h
+++ b/src/main/MidiFile.h
@@ -450,7 +450,7 @@ class MasterVolEvent
     : public SysexEvent {
 public:
   MasterVolEvent(MidiTrack *prntTrk, uint8_t channel, uint32_t absoluteTime, uint8_t msb)
-      : SysexEvent(prntTrk, absoluteTime, {0x7F, 0x7F, 0x04, 0x01, 0, static_cast<uint8_t>(msb) }, PRIORITY_HIGHER) { }
+      : SysexEvent(prntTrk, absoluteTime, {0x07, 0x7F, 0x7F, 0x04, 0x01, 0, static_cast<uint8_t>(msb) }, PRIORITY_HIGHER) { }
   virtual MidiEventType GetEventType() { return MIDIEVENT_MASTERVOL; }
 };
 


sykhro:
What playback problem did you experience prior to this change? cc'ing @mikelow as he's been rewriting this stuff and I want to avoid causing conflicts with his branch

mikelow:
Thanks, I’ll take a look at this shortly. I want to verify and understand why adding channel to SysexEvent would have an effect. SysexEvents are channel agnostic.

TheComputerGuy96:
@mikelow Adding channel is not the change that fixed my issue

mikelow:
Thanks for the fix. Sorry that I missed your note about the channel thing in your summary. Because Sysex events are channel agnostic, I don't see a reason to add channel into their constructor.

As I expect you understand, the reason for the break is that the BASS audio library reads the first data byte of every Sysex event as a length value. I screwed up during my refactor and forgot to add this length byte to Master Volume. FWIW, this length byte format is non-standard, and I will remove it when we migrate away from BASS.

I did a filter-repo force push the other day that unfortunately breaks the history of branches prior to the change. If it's ok with you, I can create a new branch, cherry-pick your commit that fixes the Master Volume event, create a new PR and merge it. Otherwise, if you'd rather author a new PR I'm happy to leave you to it.

TheComputerGuy96:
@mikelow I noticed that other events (except PortamentoTimeModeEvent which might be also broken) had an extra byte added at the start with your most recent commit that after some calculations happened to be the number of bytes that the event took up (including that size byte of course)

As for the PR I might end up force pushing this branch with the channel parameter change removed (as you said that change isn't necessary at all) or I could make a new PR if that's required

sykhro:
Force-pushing here is okay

mikelow:
Whoa, hold off on merging for just a minute. I think we'll need to make a new branch altogether because the history of master was changed significantly.

sykhro:
> Whoa, hold off on merging for just a minute. I think we'll need to make a new branch altogether because the history of master was changed significantly.

Oh, sorry, I misinterpreted your message. The force-push was to master? That's odd, it should be disabled...

sykhro:
Ah, nevermind. For some reason "lock branch" isn't turned on, so anyone with maintainer access can push to master bypassing PRs. I'll leave it on for now, but we'll want to disable it asap.

mikelow:
Separate discussion, but I deliberately and temporarily turned off force push to master.

@TheComputerGuy96 the easiest course of action would be to clone the repo again and create a new branch off of master, then push the change again or cherry-pick it from this branch (I _think_ cherry pick would work). Again, I'm happy to do this for you.

PR #378, opened 2023-08-12
Replace QHexView with a new HexView class

mikelow:
Add a new custom HexView QWidget and remove QHexView.
(also, added a fix for a MacOS-specific issue in VGMFileTreeView that was causing exceptions to be thrown on focus and item selection)

## Motivation and Context
I was unhappy with the QHexView implementation. Draw performance was poor (I believe it did a full screen update on scroll), the text was not colorized contiguously, it allowed text editor style individual character highlighting, arrow key cursor movement was that of a text editor instead of event-aware, and I wanted to adjust the text background coloring so that hexadecimal text was centered. I started to implement these fixes within QHexView but eventually ran up against a wall.

## Description
This adds the HexView class, which is a new QWidget subclass replacement for QHexView (and all of its dependencies). 

For performance, the view only draws lines that become visible in a given paintEvent. Instead of drawing a red outline around a selected event, it now highlights the event by darkening the rest of the hex/ascii areas and drawing a shadow around the selected bytes. This is done using two child QWidgets - one which draws semi-translucent rects to dim the hex/ascii areas, and another to draw the selected event on top of this. The opacity of the dimming overlay is animated.

One possible design flaw is that the HexView does not own the QScrollArea it is contained within. QHexView was itself a QAbstractScrollArea, but given my experience trying to update QHexView, I found that this approach greatly complicated the positional logic around paintEvent. In several places, HexView checks for a QScrollArea parent and interacts with it, such as when it checks the height of the QScrollArea to determine how large to size the overlay views, or when it repositions the scroll position. I also need to hook into the QScrollArea event stream via installEventFilter.

Whenever the the parent QScrollArea resizes, the overlays must in turn be resized. I found that HexView's own resizeEvent callback was not called consistently, even when QScrollArea::setWidgetResizable(true) was called, so I hooked into QScrollArea's resize event and use its full height (rather than its viewport height, which is strangely also inconsistent). Likewise, the overlay is repositioned as the scroll position changes, so that the overlay can be sized to the viewport rather than the entire virtual size of HexView (which would indeed cause performance problems).

Oh, and I updated the event color scheme, taking inspiration from the dark mode color scheme of a popular IDE.

## How Has This Been Tested?
I ran this in Windows, MacOS Monterey, and Ubuntu 22.04. Performance was acceptable on all. Interestingly, Windows rendered the fastest.

## Screenshots:
Screenshot from 2023-08-12 12-34-37
Screen Shot 2023-08-12 at 3 28 47 PM


sykhro:
Thank you! I haven't had time to look at the code yet, but I noticed an issue which I believe is an off-by-one or pixels somehow "accumulating". See for example:
Donkey Kong Country, Theme.spc
This is the instrument set:
image
and this is the sample collection:
image

At some point I think it would be nice to reimplement:
- clicking on the item in the hex view also highlights the corresponding item in the tree view
- item tooltips in the hex view

Also, it might be nice to somehow also color-code the items on the tree view on the right so that you can match them at a glance; e.g. make it so that pitch slide up/down in the following image have a slight green tint
image

mikelow:
FYI, I just added another commit that improves draw performance by rendering each line into a QPixmap that gets cached.

Good catch with the font spacing issue. I think I know why this is happening, and you're indeed correct - it's because the true fixed font width is floating point, and all of our positioning logic uses an int. Instead of calling drawText for each individual byte, we render each VGMItem in a single call. We do the same for stretches with no VGMItem. The true floating point vs int spacing difference accumulates in these longer strings. I'll have to explore if positioning with the true floating point value makes the text look blurry. If it does, I'll explore other options.

Clicking on an item in the HexView should be selecting the item in the TreeView. If it's not, let me know what environment you're running in.

I'll look into tooltips. It's nice to have but, if it turns out difficult to implement or noticably detrimental to performance, I'd be of a mind to forego it. Frankly I hadn't even noticed it was there, so I expect discoverability is low, and the information is accessible on selection. As a heads up, I'm planning to revamp the TreeView item appearance to be more minimal, as it is in the old Windows version. For such an incredibly long list of items, I think each cell takes up far too much height. The description, offset, and length information can be displayed in a status bar.

I had the same idea about using color coding in the TreeView, though the more I thought about it, the more I figured it would not be visually helpful except to establish color/event association. The issue is simply that the hex view lists items left to right while the treeview is purely vertical, so visually there's no commonality. Easy to do an experiment, though.

mikelow:
Oh, I misread the suggestion on tree view color. Making the selection background color match is an interesting idea.

sykhro:
> FYI, I just added another commit that improves draw performance by rendering each line into a QPixmap that gets cached.
> 
> Good catch with the font spacing issue. I think I know why this is happening, and you're indeed correct - it's because the true fixed font width is floating point, and all of our positioning logic uses an int. Instead of calling drawText for each individual byte, we render each VGMItem in a single call. We do the same for stretches with no VGMItem. The true floating point vs int spacing difference accumulates in these longer strings. I'll have to explore if positioning with the true floating point value makes the text look blurry. If it does, I'll explore other options.
> 
> Clicking on an item in the HexView should be selecting the item in the TreeView. If it's not, let me know what environment you're running in.
> 
> I'll look into tooltips. It's nice to have but, if it turns out difficult to implement or noticably detrimental to performance, I'd be of a mind to forego it. Frankly I hadn't even noticed it was there, so I expect discoverability is low, and the information is accessible on selection. As a heads up, I'm planning to revamp the TreeView item appearance to be more minimal, as it is in the old Windows version. For such an incredibly long list of items, I think each cell takes up far too much height. The description, offset, and length information can be displayed in a status bar.
> 
> I had the same idea about using color coding in the TreeView, though the more I thought about it, the more I figured it would not be visually helpful except to establish color/event association. The issue is simply that the hex view lists items left to right while the treeview is purely vertical, so visually there's no commonality. Easy to do an experiment, though.

i tried it again and it works now.. might have been some sort of focus issue, I'll try to reproduce

mikelow:
Glad to hear it's working. Be aware, I just pushed a change a minute ago to fix the font spacing issues. I also changed the font to [Roboto Mono](https://fonts.google.com/specimen/Roboto+Mono) (was debating between that and [Chivo Mono](https://fonts.google.com/specimen/Chivo+Mono))

sykhro:
Clicking on the hex view to reveal the tree item doesn't work with the instrument set in DKC's Theme.spc (I'm guessing because no style is being set?). Also, clicking on "Sample 0" in the sample collection of the same song causes a crash (both from the tree view and the hex view).
We might have some funky memory access there, if not UB or an use after free of some kind. Might be worth bound-checking the addresses in the hex view to at least avoid a crash.

mikelow:
I fixed the crash on selecting Sample 0. It wasn't a memory out of bounds issue, but something related to the selectionView. On that subject, I added more optimization, in the form of caching the selectionView into a pixmap.

Not sure why the instrument set items are messed up. I'll take a look.

mikelow:
Just to follow up about the problems in DKC, there are a few things going on there. 
* Messiness with VGMRgn being a container item / VGMContainerItem::GetItemFromOffset() not having enough flexibility to discriminate about what item it selects (we want to select those leaf container items, but never the VGMFile itself, for example)
* The sequence item structure is also a mess, with every track overlapping to the end of the file

This isn't necessarily a fault of how the format is programmed, but complications we hit when old formats do wacky things to preserve space, like use goto statements across multiple tracks that all jump to the same block of events (that's probalby what's going on).

Anyway, I can happily report this behavior isn't a regression, since QHexView handled it no better. My old dirty windows HexView passed includesContainer: true to GetItemFromOffset(), which results in better behavior in some circumstances, and worse in others (VGMRgns get selection, but clicking in the overlapping space of the sequence tracks no longer selects an event, but the first track).

Long story short, I don't think this needs to be addressed before merge. To address it we should probably define an enum of options for GetItemFromOffset() to get the precise behavior we want in different circumstances.

Oh and I forgot to mention that I did add Tooltips.

src/ui/qt/workarea/VGMFileView.cpp
```diff
@@ -8,61 +8,57 @@
...
-  markEvents();
+//  QFont font("Ubuntu Mono", QApplication::font().pointSize() + 5);
+//  font.setPointSizeF(QApplication::font().pointSizeF() + 3.0);
+//
+//  QFontInfo font_info(font);
+//  if (font_info.fixedPitch()) {
+//    m_hexview->setFont(font);
+//  } else {
+//    qtVGMRoot.AddLogItem(new LogItem(L"error loading font.", LOG_LEVEL_WARN, L"VGMFileView"));
+//  }
```
sykhro, line 32: Dead code

src/ui/qt/workarea/HexView.cpp
```diff
@@ -0,0 +1,713 @@
...
+  return (vgmfile->unLength + BYTES_PER_LINE - 1) / BYTES_PER_LINE;
+}
+
+void HexView::setSelectedItem(VGMItem *item) {
+  selectedItem = item;
+
+  if (selectedItem == nullptr) {
+    showOverlay(false, true);
+    selectionView->hide();
+    prevSelectedItem = nullptr;
```
sykhro, line 115: For legibility, I would return early here and drop the else branch

src/ui/qt/workarea/HexView.cpp
```diff
@@ -0,0 +1,713 @@
...
+    return true;
+  }
+
+  return QWidget::event(e);
+}
+
+void HexView::changeEvent(QEvent *event) {
+  if (event->type() == QEvent::ParentChange) {
+    QScrollArea* scrollArea = getContainingScrollArea(this);
+    if (scrollArea) {
```
sykhro, line 192: Ditto for the early return (just for this line, we still want to keep the `event->type()` checks as they are since that makes things quicker when adding code to handle more events)

src/ui/qt/workarea/HexView.cpp
```diff
@@ -0,0 +1,713 @@
...
+
+      QPainter pixmapPainter = QPainter(&pixmap);
+
+      int baseOffset = static_cast<int>(selectedItem->dwOffset - vgmfile->dwOffset);
+      int startLine = baseOffset / BYTES_PER_LINE;
+      int startColumn = baseOffset % BYTES_PER_LINE;
+      int numLines = ((startColumn + static_cast<int>(selectedItem->unLength)) / BYTES_PER_LINE) + 1;
+
+      uint8_t* data = new uint8_t[selectedItem->unLength];
+      vgmfile->GetBytes(selectedItem->dwOffset, selectedItem->unLength, data);
```
sykhro, line 313:
```suggestion
      auto itemData = std::vector<uint8_t>(selectedItem->unLength):
      vgmfile->GetBytes(selectedItem->dwOffset, selectedItem->unLength, itemData.data());
```
This will increase performance as most of the items are small and would benefit from Small Buffer Optimization

sykhro (commented):
Looks very good to me overall, I've left some minor suggestions

PR #379, opened 2023-08-13
remove high dpi scaling and pixmaps for Qt version < 6

mikelow:
When running Windows 10 on a 4k monitor with 150% display size (the default), the app looks huge. Removing `setAttribute(Qt::AA_EnableHighDpiScaling);` fixes the issue.

I'm curious if there is another configuration that prompted the use of these settings. For me, running on a high dpi monitor, it seems to be conflicting with the default method Windows' uses to scale up.


sykhro:
It's because the app used to be DPI-unaware. Now that absolute pixels calculations are gone everywhere, this is no longer needed

Issue #370, opened 2023-06-26
Upgrade builds to Qt 6


sykhro:
**Is your feature request related to a problem? Please describe.**
Current builds use Qt5 which is EOL and receives no new feature upgrades.

**Describe the solution you'd like**
We should migrate to Qt6, which has:
- a nicer CMake interface
- better system integration (e.g. macOS dialogs)
- better accessibility options
- better handling of dpi scaling
- faster rendering

**Additional context**
Qt is a big dependency and version updates are a big chore, so this change will be made gradually.
Related to #356 

- [x] Upgrade macOS CI builds to use Qt6
- [x] Upgrade Ubuntu CI builds to use Qt6
- [x] Update Qt Windows submodule to use prebuilt Qt6
- [x] Upgrade Windows CI builds to use Qt6
- [x] Drop Qt5 from CMakeLists
- [x] Drop any Qt5-specific code

PR #381, opened 2023-08-15
Upgrade macOS builds to Qt6, native M1 builds

sykhro:
## Description
Adds a new build job for compiling a Apple Silicon version of VGMTrans so that it can be run without Rosetta.
Note that these builds would need to be re-signed by users that want to run them locally as we don't have a codesigning certificate.
In bumping the runner version of Qt to Qt6, I needed to explicitly link to `Qt::DBus` to workaround a problem with `macdeployqt` not copying the relevant Framework in the App Bundle.
The CI no longer sets the minimum macOS version, as that's inforced by the Qt version we're using. Users compiling locally can still target older macOS versions.

## Motivation and Context
See #370.

## How Has This Been Tested?
Tested on macOS Ventura 13.5 with both Qt5 and Qt6.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,23 +6,60 @@ on:
   pull_request:
   workflow_dispatch:
 jobs:
-  macos_x64:
-    name: macOS (Intel)
+  macos_arm64:
+    name: macOS (Apple Silicon)
     runs-on: macos-12
-    env:
-      osx_min_ver: "11.0"
-      osx_arch: "x86_64"
     steps:
       - uses: actions/checkout@v3
       - name: "Install dependencies"
         run: |
-          brew upgrade qt@5 ninja || brew install qt@5 ninja
+          brew upgrade || brew install ninja
+          brew install \
+            --force-bottle \
+            --ignore-dependencies \
+            "$(brew fetch --deps --bottle-tag=arm64_monterey qt@6)"
+
           echo "BRANCH=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_ENV
           echo "/usr/local/bin" >> $GITHUB_PATH
       - name: "Prepare build"
         run: |
-          cmake -DCMAKE_OSX_DEPLOYMENT_TARGET=${{ env.osx_min_ver }} \
-                -DCMAKE_OSX_ARCHITECTURES=${{ env.osx_arch }} \
+          # CMAKE_PREFIX_PATH:
+          # On homebrew for arm, the default path is /opt/homebrew
+          # On Intel macs, it's /usr/local
+          # CMAKE_OSX_ARCHITECTURE:
+          # For arm, it's arm64. This selects the appropriate SDK
+          # CMAKE_SYSTEM_PROCESSOR:
+          # Has to be set so that CMake knows if we're cross-compiling
+          cmake \
+                -D CMAKE_PREFIX_PATH="/opt/homebrew;/usr/local/opt/qt"  \
+                -D CMAKE_OSX_ARCHITECTURES=arm64 \
+                -D CMAKE_SYSTEM_PROCESSOR=arm64 \
+                -D CMAKE_BUILD_TYPE=Release \
+                -G Ninja \
+                -B build
+      - name: "Build project"
+        run: cmake --build build --config "Release" --parallel
+      - working-directory: "build"
+        name: "Package DMG (macOS)"
+        run: /usr/local/opt/qt6/bin/macdeployqt bin/VGMTrans.app -dmg -always-overwrite
+      - name: "Upload artifact"
+        uses: actions/upload-artifact@v3
+        with:
+          name: VGMTrans-AppleSilicon-${{ github.sha }}-${{ env.BRANCH }}-${{ runner.os }}.dmg
+          path: "build/bin/VGMTrans.dmg"
+  macos_x64:
+    name: macOS (Intel)
+    runs-on: macos-12
+    steps:
+      - uses: actions/checkout@v3
+      - name: "Install dependencies"
+        run: |
+          brew upgrade qt@6 ninja || brew install qt@6 ninja
+          echo "BRANCH=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_ENV
+          echo "/usr/local/bin" >> $GITHUB_PATH
+      - name: "Prepare build"
+        run: |
+          cmake \
                 -DCMAKE_BUILD_TYPE=Release \
                 -G Ninja \
                 -B build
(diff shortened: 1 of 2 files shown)


mikelow:
I can't easily test, but feel free to merge.

sykhro:
> I can't easily test, but feel free to merge.

No worries, I'll need to see how big of a hack getting the M1 build on GHA is. If it turns out to be unmaintainable, I'll only leave in the Qt6 update.

mikelow:
I now have access to an M1 mac and I can compile and run the master branch with Qt6. Is the change to CMakeLists.txt only necessary for fixing GHA, or is the fact I can build on Qt6 unexpected?

sykhro:
> I now have access to an M1 mac and I can compile and run the master branch with Qt6. Is the change to CMakeLists.txt only necessary for fixing GHA, or is the fact I can build on Qt6 unexpected?

This is CI-only. The build system currently allows compiling with Qt5 and Qt6.
The issue is that `macdeployqt` is apparently broken and it creates an archive that doesn't actually work, at least with the Qt version from homebrew.

I can't observe the failure building locally (I have an M1 mac too), suggesting this is probably a problem with the x86 version that homebrew ships.

Also, we can't ship arm64 because:
- GHA still doesn't have arm runners in the free tier
- even if we did, the binaries are useless because signing + notarizing is now mandatory. if an arm64 binary isn't signed, it will SIGSEV with a cryptic error message

mikelow:
Ah, I didn't understand at first that you're attempting to cross-compile to arm64. That makes sense, and it sucks it's not working.

Codesigning on Apple platforms is its own circle of hell. Are you sure that it's impossible to sign the app for release without a paid Apple Developer account?

I just happened upon the [SuperCollider](https://github.com/supercollider/supercollider) project, which seems to be analogous to ours. It's a multiplatform Qt C++ project using GHA for deployment to github. I do wonder if [their workflow](https://github.com/supercollider/supercollider/blob/develop/.github/workflows/actions.yml) could be instructive to us. There's a [PR](https://github.com/supercollider/supercollider/pull/5650) on the subject of using ad-hoc code-signing (which they've integrated into [CMake](https://github.com/supercollider/supercollider/blob/develop/editors/sc-ide/CMakeLists.txt)). I wondering if they're able to distribute using ad-hoc signing. I'd be totally satisfied if we could deploy a release that mandates users allow non-App Store apps in Settings and which gives a simple "Are you sure you want to download this application you downloaded from the Internet" warning on first-run.

sykhro:
> Are you sure that it's impossible to sign the app for release without a paid Apple Developer account?

We can self-sign it, but it doesn't change the end result - if the cert is not (a paid one) issued by Apple, the signature check fails. At the moment, if we were to distribute an ARM64 executable, users would have to do:
```zsh
xattr -d com.apple.quarantine VGMTrans.app # Removes the "this app is from an unknown developer" warning
codesign --deep --force --sign <LOCAL_SELF_SIGNED_CERT> VGMTrans.app # Signs the executable with a system self-signed certificate
```
rendering the whole CMake & CI signing machinery useless (you can run those commands on any binary)

Other OSS projects have the same problem. Some bigger ones can get those 99$/year through donations. No luck for the rest of us.

> I'd be totally satisfied if we could deploy a release that mandates users allow non-App Store apps in Settings and which gives a simple "Are you sure you want to run this application you downloaded from the Internet" warning on first-run.

That's how it already is, for Intel binaries at least.

_It should be noted that when compiling locally, (Apple)Clang does self-signing automatically. That's why you can compile & execute binaries even without invoking codesign explicitly._

mikelow:
I did a bit of experimentation between two M1 machines. While I ran into errors which I believe are related to macdeployqt on the current brew version (6.6.0_1) for Apple Silicon, I believe it is possible to deploy the app for Apple Silicon with ad-hoc codesigning, albeit with inconvenience to the user.

Specifically, it is necessary to codesign not just VGMTrans.app itself, but every framework and dylib file in the /Content/Frameworks directory of the app, and similarly recursively in /Contents/PlugIns. An example of the command I used is:
 `codesign -s - -f ./VGMTrans.app`

Without doing this, I get an error message that the app is corrupted. With this comprehensive codesigning, the error switches to a security warning, and we find ourselves in the state where the app can be opened by right/ctrl clicking and selecting Open from the menu, albeit with further security warnings. Fortunately, those only appear the first time you attempt to open the app, and normal app opening is possible thereafter.

However, as I alluded to, I ran into runtime errors that seem to stem from macdeployqt not transferring the QtDbus framework into the Frameworks dir. Copying the framework in manually doesn't work as it seems the paths for loading dependences are set to absolute brew paths and not relative paths. 🤷‍♂️

sykhro:
> Without doing this, I get an error message that the app is corrupted.

Yes, that's what I was trying to say. We might as well just distribute the app and tell people to run these two commands to have it working.

> However, as I alluded to, I ran into runtime errors that seem to stem from macdeployqt not transferring the QtDbus framework into the Frameworks dir. Copying the framework in manually doesn't work as it seems the paths for the framework's dependencies are set to absolute brew paths and not relative paths. 🤷‍♂️'

Yup, it's the same issue we have with CI here. I will try to see if I can change the @rpath with libtool or something; it's a hack, but it should work.

mikelow:
> So wait, the app you deep sign on a machine works on the other? Do they have the same icloud keychain?

The machines do not share an icloud keychain. Depends on what you mean by "works". I have seen the app in two different states, depending on whether I sign everything necessary or not:

1) Not signed: opening app leads to an error about corruption. Can't do anything.
2) Signed: opening app leads to a security warning. App can be executed by right/ctrl clicking then selecting Open from the menu. This leads to another warning dialog gives the option to actually run the app.

Were you able to get the app to state #2?

sykhro:
I never managed to get the app to state #2, but I was using a different codesign command. I will try with this and I'll let you know.

sykhro:
Superseded by #412

mikelow:
@sykhro just curious, do you plan to still work on getting GHA to cross-compile arm64 MacOS builds? It seems like you got pretty close. I'm happy to look into using otool or something to fix the QtDbus path issue.

sykhro:
> @sykhro just curious, do you plan to still work on getting GHA to cross-compile arm64 MacOS builds? It seems like you got pretty close. I'm happy to look into using otool or something to fix the QtDbus path issue.

Yes. The QtDbus path thing is fixed with commit 1483f7da1a60b38d3f4aa98f3b10a3887fc10c67, I will make a new PR for the M1 binaries job

mikelow:
Oh, awesome. My bad for not noticing the QtDbus fix in your PR.

PR #398, opened 2023-12-04
Add UI_GetResourceDirPath() and update MAMELoader to use it

mikelow:
Also updates the CMakeLists to copy entire /bin directory to bundle's /Contents/Resources directory on MacOS.

## Description
Currently MacOS fails to load the mame_roms.xml file because the default working directory when opening a Mac application is the system root. This adds a system-specific function to get the path to the application's resource directory, which for all other platforms is the same path as the executable, at least for now.

## Motivation and Context
This fixes the MAME loader on MacOS

## Testing
Tested on MacOS Sonoma, Windows 10, and Ubuntu 22.

### DIFF
--- a/src/main/Root.cpp
+++ b/src/main/Root.cpp
@@ -304,3 +304,7 @@ void VGMRoot::AddLogItem(LogItem *theLog) {
   vLogItem.push_back(theLog);
   UI_AddLogItem(theLog);
 }
+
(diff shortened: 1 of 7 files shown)


sykhro:
Thanks! there was probably an alternative way with the Qt internal fs but this is fine too

PR #399, opened 2023-12-05
Add initial portamento value to CPS, update to use decacents per second

mikelow:
Use decacents per sec instead of cents per sec for portamento timing, aligning with a [change to the fluidsynth fork](https://github.com/mikelow/fluidsynth/commit/b02801e03b410fc24e10af42f1fd160f9c46da93) used by the [vst plugin](https://github.com/mikelow/juicysfplugin)

The reason for this change is to allow faster portamento slides than cents per second allows. CPS sequences often use a portamento time of 0, which emulates a legato effect (no slide, but no attack phase) - effectively the same as instantaneous portamento. A [trill between 5ths](https://www.youtube.com/watch?v=tc7nRcTfB5g) at maximum cents per second value (16,383) still causes audible slide.

I investigated using legato pedal events when CPS sets portamento time to 0, but the default MIDI behavior, at least as implemented in fluidsynth, is not as desired (I believe notes must overlap to have this effect). The other downside would be that introducing the legato pedal state on top of portamento on/off state adds complexity. Better to just use very fast portamento.

### DIFF
--- a/src/main/formats/CPSTrackV1.cpp
+++ b/src/main/formats/CPSTrackV1.cpp
@@ -29,7 +29,7 @@ void CPSTrackV1::ResetVars() {
 
 void CPSTrackV1::AddInitialMidiEvents(int trackNum) {
   SeqTrack::AddInitialMidiEvents(trackNum);
-  AddPortamentoTime14BitNoItem(0x3FFF);
+  AddPortamentoTime14BitNoItem(0);
 }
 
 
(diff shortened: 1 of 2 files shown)


sykhro:
Makes sense. How does this affect playback when not using the VST?

mikelow:
I'm glad you asked, because I thought this would have a relatively insignificant effect. Turns out it totally screws up the sound, primarily because I'm now setting a default portamento time value and it's the max. I can't find any official MIDI spec defining or even suggesting how to interpret the Portamento Time control event values. It seems to be purposefully ambiguous.

But FluidSynth interprets it as a duration in ms to slide between two notes, and it looks like BASS does something similar. With my definition, 0x3FFF is the fastest possible slide. With default FluidSynth and BASS, it's the slowest. 

I'm debating redefining my alternative portamento time mode to be inverted, such that 0x3FFF represents 0 decacents per sec, and 0 represents 0x3FFF decacents per sec. It's more confusing, but at least it won't scale inversely with the common usage of portamento time.

mikelow:
I've made changes to the VST plugin so that the portamento time values scale in the same direction as default Fluidsynth and BASS - smaller portamento time values result in faster slides. Though portamento time values will be inaccurate on BASS, they will be less so. To be clear, outside of using the VST plugin, they've always been inaccurate, as CPS represents portamento time as a rate of pitch change over absolute time, while default Fluidsynth (and likely BASS) interpret portamento time as an absolute time duration for the slide between any two arbitrary notes. This change at least offers an improvement.

I can kind of imagine how you might emulate the rate of change effect within the time duration implementation, but it would be quite tricky and I'd rather not spend time there.  Going to merge this change.

sykhro:
Yes, this approach seems sensible. Wondering if perhaps midi2.0 offers a solution to this

mikelow:
Doesn't appear so. Best I can tell, they defer back to old MIDI 1.0 definitions for most controller events including Portamento / Portamento Time. They do have a section talking about the Portamento Control event (note-specific portamento control), but it seems to defer back to the MIDI 1.0 specs for its "fundamental functions":
image

PR #404, opened 2023-12-11
Use duration portamento time on CPS

mikelow:
Turns out this was not as hard to implement I thought it would be.

This changes the CPS driver to express portamento time as a duration in milliseconds in order to increase compatibility with midi players/instruments. MIDI does not define a standard for interpreting portamento time, but millisecond duration is how Fluidsynth does it. BASS does something similar, but there are audible differences, both in terms of timing and more generally when it decides to apply portamento. (It appears [BASS doesn't support](https://www.un4seen.com/doc/#bassmidi/BASS_MIDI_StreamEvent.html) controller 37 - portamento time fine - though I don't think this entirely accounts for the timing difference).

When comparing this version using Fluidsynth against the prior portamento time-as-rate implementation, I haven't yet detected a difference. 

I also threw in some MarkerEvent related methods that I added in the course of solving this. The change doesn't use them, but I figure we keep them.

### DIFF
--- a/src/main/MidiFile.cpp
+++ b/src/main/MidiFile.cpp
@@ -637,7 +637,14 @@ void MidiTrack::AddMarker(uint8_t channel,
   aEvents.push_back(new MarkerEvent(this, channel, GetDelta(), markername, databyte1, databyte2, priority));
 }
 
-
+void MidiTrack::InsertMarker(uint8_t channel,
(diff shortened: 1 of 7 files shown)


sykhro:
wow, this is more straightforward than I thought it was going to be. thanks!

PR #405, opened 2023-12-12
Fix performance issue when loading files with many collections

mikelow:
I was finding that, on all platforms, dropping multiple mame roms into the app would bring it to a standstill as it loaded collections. Profiling showed that it was the UI rendering of listview items bringing things to a crawl.  The problem is that we're telling the QListView that all rows have changed whenever a new collection is added, and some mame roms add thousands of collections:

In VGMCollListViewModel:
```
  connect(&qtVGMRoot, &QtVGMRoot::UI_AddedVGMColl,
          [=]() { dataChanged(index(0, 0), index(rowCount() - 1, 0));
  });
```

We could potentially fix this a number of ways: add an index to the UI_AddedVGMColl callback so that we only `dateChanged()` that index, assume the new collection is appended at the last index, or wait until the load is complete then reload the model. I went with the last option.

This adds callbacks in Root to tell us when a RawFile is being loaded. We hook these up to some signals in QtVGMRoot that tell our VGMCollListViewModel to call `beginResetModel()` and `endResetModel()` when the loading begins/ends respectively. Since we can still add collections manually, we keep that connect block but prevent it from running during a raw file load.

Related note: we might want to move loading to a background thread.

Sorry about the code clean-up in CPSTrackV1.

## How Has This Been Tested?
Tested on MacOS build. Confirmed it's fixed the performance issue. Tested manual collection creation. Tested deleting collections.

### DIFF
--- a/src/main/Root.cpp
+++ b/src/main/Root.cpp
@@ -137,11 +137,13 @@ bool VGMRoot::CreateVirtFile(uint8_t *databuf,
 // to the RawFile
 bool VGMRoot::SetupNewRawFile(RawFile *newRawFile) {
   newRawFile->SetProPreRatio((float) 0.80);
+  this->UI_OnBeginLoadRawFile();
 
   if (newRawFile->processFlags & PF_USELOADERS)
     for (uint32_t i = 0; i < vLoader.size(); i++) {
       if (vLoader[i]->Apply(newRawFile) == DELETE_IT) {
         delete newRawFile;
+        this->UI_OnEndLoadRawFile();
         return true;
       }
     }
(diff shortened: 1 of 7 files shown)


sykhro (approved):

sykhro:
> Related note: we might want to move loading to a background thread.

I wish! I tried making some parts multithreaded a while ago but it broke in subtle ways. But if we restrict the UI to just playback and not freezing (as in, not loading files in parallel) I think it might be doable in the short-ish term

mikelow:
> I wish! I tried making some parts multithreaded a while ago but it broke in subtle ways. But if we restrict the UI to just playback and not freezing (as in, not loading files in parallel) I think it might be doable in the short-ish term

Oof, sorry to hear that. Yeah, we would definitely use a single thread for file loading. I think even putting up a modal dialog with a progress bar and a cancel button would be fine.

PR #406, opened 2023-12-12
CI: Add FreeBSD 14.0 and remove FreeBSD 12.4.

emaste:
FreeBSD 12.4 is EOL at the end of 2023, and FreeBSD 14.0 was released recently.

## Description
Add FreeBSD 14 and remove FreeBSD 12

## Motivation and Context
FreeBSD 14 is released now. FreeBSD 12 is retired.

## How Has This Been Tested?
Will be tested via CI in pull request workflow.

## Types of changes

## Checklist:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -2,7 +2,6 @@ task:
     name: FreeBSD
     freebsd_instance:
         matrix:
-            image_family: freebsd-12-4
             image_family: freebsd-13-2
 
     install_script: pwd && ls -la && pkg update --force && pkg install -y cmake ninja qt6-base qt6-svg minizip


sykhro:
thanks for the PR! do you have any insight on those linker errors? should I raise that upstream?

mikelow:
Big ups for the PR, my dude.

emaste:
> do you have any insight on those linker errors?

Not yet, what I might do is make this pull req just remove 12.4, and submit a second one that adds 14.0 once those issues are sorted. FreeBSD 14 brings in a new LLVM linker (ld.lld) and it looks like its behaviour has changed.

emaste:
Some background:
- error originally introduced in lld in https://github.com/llvm/llvm-project/commit/b308cace8496a54e4674b305481f7e5886284306
- changed to a warning in 2017 in https://github.com/llvm/llvm-project/commit/fbe68a3584855871a2d99c57e53cc4572493daf9
- @MaskRay changed to errorOrWarn in 2022 in https://github.com/llvm/llvm-project/commit/3ac94280245415be66cb1b603367c5f4f6d498e7

emaste:
> FreeBSD freebsd_instance:family/freebsd-12-4 Expected — Waiting for status to be reported

I'm not sure how this (that freebsd-12-4 is required) is configured; it does not appear to be in the repo data. Maybe it is configured in GitHub settings?

sykhro:
> > FreeBSD freebsd_instance:family/freebsd-12-4 Expected — Waiting for status to be reported
> 
> I'm not sure how this (that freebsd-12-4 is required) is configured; it does not appear to be in the repo data. Maybe it is configured in GitHub settings?

The checks come from the GitHub App for cirrus, so it makes sense that they don't change based on the PR contents until it's merged

jrtc27:
It means that your vendored pre-built copy of libbassmidi.so violates the ELF spec and must have been produced by a broken linker.

Though you shouldn't be using that anyway for FreeBSD, that's built for x86 GNU/Linux and references all manner of @GLIBC_* symbols, which is a much bigger problem. That's never going to actually work even if the file weren't itself invalid.

mikelow:
We're pretty tight on options given that BASS is not open source. @skyhro I assume you confirmed the FreeBSD version running with libbassmidi.so at some point? If it's working fine despite @GLIBC_ symbols, we may as well use it for lack of alternative. @emaste is there a linker parameter to make that `errorOrWarn()` call issue a warning instead of an error?

FWIW, I'm working on a [branch](https://github.com/vgmtrans/vgmtrans/tree/juce2) that uses JUCE instead of BASS via a VST instrument. However, though it is open source, JUCE doesn't have official FreeBSD support either.

sykhro:
> Though you shouldn't be using that anyway for FreeBSD, that's built for x86 GNU/Linux and references all manner of @GLIBC_* symbols, which is a much bigger problem. That's never going to actually work even if the file weren't itself invalid.

Ah, makes sense. Honestly, freebsd was an accident - I was testing out circleci integration for macos, but that had limited build minutes, so I tried freebsd. It then remained there just because it was building.

> @sykhro I assume you confirmed the FreeBSD version is running with libbassmidi.so at some point? If it's working fine despite @GLIBC_ symbols, we may as well use it for lack of alternative.

See above. But I recall someone testing it out...

> FWIW, I'm working on a [branch](https://github.com/vgmtrans/vgmtrans/tree/juce2) that uses JUCE instead of BASS via a VST instrument. However, though it is open source, JUCE doesn't have official FreeBSD support either.

I'm guessing there are ways to build juce for freebsd, but I'm not 100% sure we should bother given that almost nobody uses the linux version, let alone this one. Sorry!

mikelow:
> I'm guessing there are ways to build juce for freebsd, but I'm not 100% sure we should bother given that almost nobody uses the linux version, let alone this one. Sorry!

I see a fork or two of JUCE for FreeBSD, but I'm not sure that any are up-to-date or maintained. At any rate, I totally agree that supporting FreeBSD is not a priority. No offense to the BSD users within this PR.

So, shall we just drop FreeBSD from the CI configs?

sykhro:
> So, shall we just drop FreeBSD from the CI configs?

I'm honestly in favour of this. I think FreeBSD users would probably compile the program on their own rather than downloading a binary anyway.

Sincere thanks for opening the PR though. It's always nice to receive contributions.

jrtc27:
If your dependencies don't support FreeBSD then it's a bit of a shame but I guess not much you can do, you're just SOL

emaste:
> FWIW, I'm working on a [branch](https://github.com/vgmtrans/vgmtrans/tree/juce2) that uses JUCE instead of BASS via a VST instrument. However, though it is open source, JUCE doesn't have official FreeBSD support either.

Interesting -- there is a growing community of people interested in music production using FreeBSD, and I see that JUCE is in the FreeBSD ports tree for a couple of years (and is kept up to date) thanks to @yurivict https://github.com/freebsd/freebsd-ports/commits/main/devel/juce. There are a couple of patches in the FreeBSD port that perhaps should be sent upstream.

> FreeBSD users would probably compile the program on their own

This is probably true, and in the interest of transparency I came across this project only because I was looking for .cirrus.yml files that reference FreeBSD 12.4, which will be EOL in the near future. The 12.4 jobs will start failing once packages are no longer available, so I wanted to make sure that those jobs are either removed or upgraded before then.

If vgmtrans does migrate to JUCE it seems that building and running on FreeBSD should be straightforward. Having a build job in a CI environment is great to quickly identify build failures so that they can be easily addressed right when they're introduced, rather than with more difficulty later on.

All of that said, my primary goal here was to avoid you having CI failures show up some time after Dec 31 as FreeBSD 12.4 infrastructure starts getting retired, and removing FreeBSD from your CI entirely will of course also solve that issue. I have updated the commit though to only remove 12.4 rather than adding 14.0, and I'll give the JUCE branch mentioned above a try on FreeBSD 14.

mikelow:
@emaste Thanks for all the help! Fair warning that the JUCE branch is not very friendly yet in terms of build process. I did update with some instructions in the README.md, though, in case you want to move forward.

sykhro:
Thank you for your time and contributions. We have decided to drop FreeBSD CI for now, we will re-enable it once the project has moved to JUCE.

PR #407, opened 2023-12-13
Improvement to HexView splitter behavior and font increase/decrease

mikelow:
Resizing the app window causes both hex view and tree view to resize 1:1. I've added logic to limit the maximum width of the hex view so that it doesn't expand beyond the ASCII text column, allowing extra space to be devoted to the tree view. 

I also added logic so that increasing/decreasing the font size will make the hex view proportionally grow/shrink, so if you've repositioned the splitter to, say, hide the ASCII view, you won't need to reposition it after a font size change.

Also, I found that increasing/decreasing the font size in increments of 0.5 was sometimes having no effect, as Qt seems inexact about font sizing. I added some logic to increment the font size until there is an actual change in character width.

I also decreased the default font size a bit.

## How Has This Been Tested?
Tested on Mac and Windows

### DIFF
--- a/src/ui/qt/workarea/VGMFileView.cpp
+++ b/src/ui/qt/workarea/VGMFileView.cpp
@@ -54,7 +54,12 @@ VGMFileView::VGMFileView(VGMFile *vgmfile)
           });
 
   connect(new QShortcut(QKeySequence::ZoomIn, this), &QShortcut::activated, [&] {
-    updateHexViewFont(0.5);
+    updateHexViewFont(+0.5);
+  });
+
+  QShortcut* shortcutEqual = new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_Equal), this);
+  connect(shortcutEqual, &QShortcut::activated, [&] {
+    updateHexViewFont(+0.5);
   });
 
   connect(new QShortcut(QKeySequence::ZoomOut, this), &QShortcut::activated, [&] {
--- a/src/ui/qt/workarea/VGMFileView.cpp
+++ b/src/ui/qt/workarea/VGMFileView.cpp
@@ -34,7 +34,12 @@ VGMFileView::VGMFileView(VGMFile *vgmfile)
 
   m_splitter->addWidget(m_hexScrollArea);
   m_splitter->addWidget(m_treeview);
-  m_splitter->setSizes(QList<int>() << 900 << 270);
+  m_splitter->setSizes(QList<int>{hexViewWidth(), treeViewMinimumWidth});
+  m_splitter->setStretchFactor(0, 0);
+  m_splitter->setStretchFactor(1, 1);
+  m_hexScrollArea->setMaximumWidth(hexViewWidth());
+  m_treeview->setMinimumWidth(treeViewMinimumWidth);
+
 
 
   connect(m_hexview, &HexView::selectionChanged, this, &VGMFileView::onSelectionChange);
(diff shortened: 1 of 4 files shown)


sykhro:
I'm wondering if the 0.5 thing is related to the font format used internally, but we don't really allow having that as a user setting so this is fine. Do you know what the effects are with an hi-dpi screen?

mikelow:
> I'm wondering if the 0.5 thing is related to the font format used internally, but we don't really allow having that as a user setting so this is fine. 

Could you elaborate? I'm confused.

> Do you know what the effects are with an hi-dpi screen?

Do you mean the splitter? The splitter logic works as expected on a hi-dpi screen.

mikelow:
Oh one thing I forgot to ask:

How do you feel about changing our PR ceremony to giving a "lgtm" instead of the reviewer actually merging? Sometimes I realize I want to make an additional change, and it gives me peace of mind to not have the race the change in before it can be reviewed.

sykhro:
> Could you elaborate? I'm confused.

I mean, I suspect that there's some sort of rounding in the internals of Qt which results in this particular font not being rendered with certain increments. Users can't change the font of the splitter though so this shouldn't break for anyone.

> Do you mean the splitter? The splitter logic works as expected on a hi-dpi screen.

I meant the font-resizing adjustment

sykhro:
> Oh one thing I forgot to ask:
> 
> How do you feel about changing our PR ceremony to giving a "lgtm" instead of the reviewer actually merging? Sometimes I realize I want to make an additional change, and it gives me peace of mind to not have to race the change in before it can be reviewed.

I was wondering about the same yesterday. It's the curse of being used to people contributing 1-line changes to this project with no follow up :p from now on, I'll leave a LGTM with the github approval thing. If we want to be extra sure, we could use draft PRs until the changes are stabilized

src/ui/qt/workarea/VGMFileView.cpp
```diff
@@ -54,7 +54,12 @@ VGMFileView::VGMFileView(VGMFile *vgmfile)
           });
 
   connect(new QShortcut(QKeySequence::ZoomIn, this), &QShortcut::activated, [&] {
-    updateHexViewFont(0.5);
+    updateHexViewFont(+0.5);
+  });
+
+  QShortcut* shortcutEqual = new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_Equal), this);
```
sykhro, line 60: Is there a particular reason we need this? Personal preference? (the change is tiny enough that I don't mind either way)

mikelow: Chrome and Firefox (and maybe others) use it. Better for discoverability.

mikelow:
> I mean, I suspect that there's some sort of rounding in the internals of Qt which results in this particular font not being rendered with certain increments. Users can't change the font of the splitter though so this shouldn't break for anyone.

Yes, I agree there seems to be some kind of rounding. I did some testing when I was working on HexView and I don't think it's specific to the font. Although they all respond a bit differently, it seemed that across the board fonts were fairly insensitive to decimal level size changes. It is strange.

Still confused what you mean about "the font of the splitter"

> I meant the font-resizing adjustment

I think I may be exclusively testing on high dpi devices. 🤷‍♂️

I'm working on two devices - a Macbook Air and a desktop with 4k monitor (Hackintosh running Mac or Windows), so I assume I'm always running in high dpi mode. I need to read up more on Qt's high dpi mode cause I don't understand its ramifications.

sykhro:
> Still confused what you mean about "the font of the splitter"

The font of the hexview

> I'm working on two devices - a Macbook Air and a desktop with 4k monitor (Hackintosh running Mac or Windows), so I assume I'm always running in high dpi mode.

Ah ok, perfect then.

> I need to read up more on Qt's high dpi mode cause I don't understand its ramifications.

You are in for a ride, especially if you start comparing platforms and pre-windows 10

mikelow:
> I was wondering about the same yesterday. It's the curse of being used to people contributing 1-line changes to this project with no follow up :p from now on, I'll leave a LGTM with the github approval thing. If we want to be extra sure, we could use draft PRs until the changes are stabilized

👍 great. Yeah, I usually don't put out a PR unless I think it's ready at the time. I wouldn't want to dump more code into a PR after it's been approved, but would create a new one. Sometimes I realize I've made a mistake after the fact, though, and having control over the merge gives me peace of mind.

mikelow:
> You are in for a ride, especially if you start comparing platforms and pre-windows 10

I will happily forego the pre-windows 10 ride 🙄. Yikes.

PR #409, opened 2023-12-13
Fix crashes on quit related to static destruction order

mikelow:
When quitting the app, if a file was ever loaded into the app, MacOS would report a crash after quitting. Looking into the crash logs, it became clear this is caused by the destruction order of static objects. Specifically, the VGMRoot object, the FormatMap registry object, and all of the Format objects are declared statically. The destructor of VGMRoot references these other static instances at some point down the stack trace, but they might be destroyed before VGMRoot is.

To fix the issue, I made the FormatMap registry and Format instances into static pointers, ensuring that they'll be destroyed after VGMRoot (upon app exit, or when we delete them ourselves).

It's unclear to me if best practice is to actually delete all of these instances on close of the app, as we're mostly doing, given that the OS presumably will do all of this for us sans the performance hit of manual deletion. Put another way, do we really need to do anything in ~VGMRoot? For consistency, I added a method to delete all of the formats from the map.

## How Has This Been Tested?
Ran this on MacOS.

### DIFF
--- a/src/main/Format.cpp
+++ b/src/main/Format.cpp
@@ -4,9 +4,9 @@
 
 using namespace std;
 
-FormatMap &Format::registry() {
-  static FormatMap registry;
-  return registry;
+FormatMap& Format::registry() {
+  static FormatMap* registry = new FormatMap();
+  return *registry;
 }
 
 Format::Format(const string &formatName) :
@@ -22,6 +22,13 @@ Format::~Format(void) {
     delete matcher;
 }
 
+void Format::DeleteFormatRegistry() {
+  for (auto& pair : registry()) {
+    delete pair.second;
+  }
+  registry().clear();
+}
+
 Format *Format::GetFormatFromName(const string &name) {
   FormatMap::iterator findIt = registry().find(name);
   if (findIt == registry().end())
@@ -33,15 +40,6 @@ bool Format::OnNewFile(VGMFile *file) {
   if (!matcher)
     return false;
   return matcher->OnNewFile(file);
-  //switch (file->GetFileType())
-  //{
-  //case FILETYPE_SEQ:
-  //	return OnNewSeq((VGMSeq*)file);
-  //case FILETYPE_INSTRSET:
-  //	return OnNewInstrSet((VGMInstrSet*)file);
-  //case FILETYPE_SAMPCOLL:
-  //	return OnNewSampColl((VGMSampColl*)file);
-  //}
 }
 
 VGMColl *Format::NewCollection() {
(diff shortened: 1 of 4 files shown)


sykhro:
The order of construction and destruction of globals is not defined, so one could argue that the best practice is avoid having ordering dependencies of this kind. Easier said than done though.

I’m fine with this change. It’s not necessary for the app itself, but it’s useful if anyone wants to reuse the code as a library (making vgmtranscore in an actual library was one of the goals of my refactor branch, but i put it off because it wasn’t compatible with the WTL GUI and i didn’t think the Qt one was ready yet)

sykhro (approved):

mikelow:
> The order of construction and destruction of globals is not defined, so one could argue that the best practice is avoid having ordering dependencies of this kind. Easier said than done though.

I agree. It would probably be better to remove singletons, especially the Format instances. Anyway, if you have specific ideas for how to refactor this, I'm all ears. I trust you have more C++ experience than me.

> I’m fine with this change. It’s not necessary for the app itself, but it’s useful if anyone wants to reuse the code as a library (making vgmtranscore in an actual library was one of the goals of my refactor branch, but i put it off because it wasn’t compatible with the WTL GUI and i didn’t think the Qt one was ready yet)

Just to be clear, the motivation for this change is that I get a crash report dialog every time I quit the app on MacOS.

What's the idea of vgmtranscore? To put all the non-ui code into a lib? I'm not opposed to this but I'm not sure I understand what the possible use cases are. Also, you're probably already on the same page, but at this point I think we should consider the WTL version dead and unsupported.

sykhro:
> I agree. It would probably be better to remove singletons, especially the Format instances. Anyway, if you have specific ideas for how to refactor this, I'm all ears. I trust you have more C++ experience than me.

I had started some of this in the `refactor` branch. I don't remember if it was `Format` or `Loader`, but I had done a self-registering template thing which helps in getting rid of the map singleton.

> What's the idea of vgmtranscore? To put all the non-ui code into a lib? I'm not opposed to this but I'm not sure I understand what the possible use cases are.

Our codebase is full of driver code which is sometimes useful when RE'ing new things. I found myself hacking away at it in the past to use just some parts. Besides that, honestly, I don't think there would be many users.

> Also, you're probably already on the same page, but at this point I think we should consider the WTL version dead and unsupported.

Yes, we have feature parity now so it's time to say goodbye. The DirectMIDI stuff is (basically) officially dead as of Windows 11 too

sykhro:
> Also, you're probably already on the same page, but at this point I think we should consider the WTL version dead and unsupported.

Speaking of this: you should delete the AppVeyor settings that are registered to your account. Otherwise it's going to try and keep building even after we delete the files from here

mikelow:
> Speaking of this: you should delete the AppVeyor settings that are registered to your account. Otherwise it's going to try and keep building even after we delete the files from here

It is done.

PR #410, opened 2023-12-14
Use system WindowText color for MarqueeLabel (respects light/dark mode)

mikelow:
Images attached:

Before:
Screenshot 2023-12-13 at 9 55 16 PM
Screenshot 2023-12-13 at 9 53 19 PM

After:
Screenshot 2023-12-13 at 9 45 12 PM
Screenshot 2023-12-13 at 9 45 33 PM

Update:
Tested on Windows and working fine. The app doesn't respect Windows dark mode in general, though.

### DIFF
--- a/src/ui/qt/util/MarqueeLabel.cpp
+++ b/src/ui/qt/util/MarqueeLabel.cpp
@@ -6,6 +6,7 @@
 
 #include "MarqueeLabel.h"
 #include <QPainter>
+#include <QApplication>
 
 MarqueeLabel::MarqueeLabel(QWidget *parent) : QWidget(parent) {
   m_static_text.setTextFormat(Qt::PlainText);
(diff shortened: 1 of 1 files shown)


sykhro:
Did you test with Qt5 or Qt6 on Windows? I recall this bug appearing with Qt6.

I should look into the dark mode thing. When I tried there were no Win32 hooks that could notify applications of the system switching to dark (21H1, i think?)

mikelow:
Tested with Qt5 on Windows 10, Qt6 on Mac.

Didn't dig too deeply into it, but it looks like Windows dark mode support may have improved with Qt6 and/or Windows 11. I'm fine if we can't utilize hooks to detect a change, so long as it respects the system setting at launch.

mikelow:
Unrelated: Isn't it about time you put your name on the contributors list in the README? I would do it except I'm not sure I would accurately summarize the scope of your work.

sykhro:
> Didn't dig too deeply into it, but it looks like Windows dark mode support may have improved with Qt6 and/or Windows 11. I'm fine if we can't utilize hooks to detect a change, so long as it respects the system setting at launch.

Ahh yes, I remember. Nowadays I have the compute power to compile Qt6 without rotting away in the meantime, so I'll send a PR to bump it up on Windows. And while I'm at it, drop Qt5. There really is no reason to keep using that, unless we care about Windows 7. I honestly don't. My rule of thumb is to support things that are officially not EOL

> Isn't it about time you put your name on the contributors list in the README? I would do it except I'm not sure I would accurately summarize the scope of your work.

Whoops. I will send a PR for that.

sykhro (approved):

PR #411, opened 2023-12-14
Improve VGMFileListView Headers

mikelow:
This improves the appearance and resizing behavior of the VGMFileListView. There are a few issues I want to fix:

1. The headers are spaced evenly and cannot be resized
2. Resizing should expand/contract the Name header instead of both headers evenly
3. There is a little splitter at the right edge of the Format section header
Screenshot 2023-12-13 at 11 40 35 PM

This fixes the above problems. I've also added logic so that the splitter resizes both sections so that their widths sum to the full header width.

The initial width of the Format section header is set by calculating the width of the "Format" text itself and adding some padding. This is necessary as the font will be different per platform.

I disable the horizontal scroll bar and override the scrollContentsBy() event so that horizontal scrolling is impossible. this is necessary to get rid of the extraneous splitter, as it could otherwise scroll into view.

Tested on MacOS and Windows.

Windows
Capture

Mac:
Screenshot 2023-12-13 at 11 31 05 PM


sykhro:
Thanks! LGTM. It would be nice to restore the header filtering/ordering thing someday

sykhro (approved):

mikelow:
> Thanks! LGTM. It would be nice to restore the header filtering/ordering thing someday

Oh yeah. That was a freebie in WTL

PR #414, opened 2023-12-14
mame_roms.xml and readme update

mikelow:
This does some cleanup to mame_roms.xml and adds the no-cd CPS3 definitions in supplement to the merged rom definitions. I could be clever and add logic so that merged sets can reference the no-cd sets and reduce some repetition, but it's not worth it as this will likely only ever apply to the 5 cps3 games. Plus, I'd sooner work on migrating to json from xml.

I've also updated the readme to reflect support for CPS3, and I've removed reference to the old WTL build in the "How to compile it" section.


sykhro:
LGTM. I recommend [this library](https://github.com/nlohmann/json) for whenever you want to switch to json

sykhro (approved):

PR #415, opened 2023-12-15
Use beginResetModel/endResetModel for batch deletion of collections

mikelow:
This is the counterpart to #405 for batch deletion of collections (when deleting a raw file, for example). Vastly improves performance for the same reasons discussed before.


sykhro (approved):
LGTM. Thanks!

PR #417, opened 2023-12-16
VGMFileListView multi select and deletion

mikelow:
Enable multiple selection in VGMFileListView, fix logic for deleting all selected files. 

The old for loop was broken because the file pointed to by `index` changed with every onClose() call.  What I wouldn't do for a simple map() function instead of reserving space for a vector and looping to populate it. I know about std::transform, but that is just unreadable.

## Motivation and Context
Might as well allow multi selection.

## How Has This Been Tested?
Tested on MacOS, selections of various size.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/workarea/VGMFileListView.cpp
+++ b/src/ui/qt/workarea/VGMFileListView.cpp
@@ -109,7 +109,7 @@ void VGMFileListModel::RemoveVGMFile() {
  */
 
 VGMFileListView::VGMFileListView(QWidget *parent) : QTableView(parent) {
-  setSelectionMode(QAbstractItemView::SingleSelection);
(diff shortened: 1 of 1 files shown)


sykhro:
what happens if you right-click files of different kind? e.g. seq and instrset

mikelow:
> what happens if you right-click files of different kind? e.g. seq and instrset

Right click ignores the selection and gives a context menu for whatever single item was right-clicked.

I find this acceptable, for now. I will work on adding context menus that allow batch conversion. I'm thinking I'll generalize the code a bit so we're not using the switch on GetFileType()

sykhro (approved):

PR #419, opened 2023-12-20
Add MenuManager and integrate it into VGMFileListView

mikelow:
As [promised](https://github.com/vgmtrans/vgmtrans/pull/417#issuecomment-1858931378), this PR adds batch conversion so that when multiple VGMFiles are selected, they can all be saved in a single command. But wait, there's more! I didn't just tack this on to VGMFileListView. I wrote a generalized solution that will hopefully make it easy to create menus for any type across the entire app.

## The core classes: <br>MenuManager, Command, CommandContext, and CommandContextFactory

This PR introduces a new "service" for menu generation called `MenuManager`. The idea is that you can register a set of "Commands" to any type you so desire, where a `Command` represents a menu action, and then pass an instance of any type to `MenuManager::GetCommands()` to retrieve the eligible `Command`s of the object. The Command object, in turn, can execute the action.

Commands require data, of course. For example, the command "Save as MIDI" needs to know what file(s) to save and what path to save them to. Hence, we need to know not only what Commands an object can execute but also the data required by them and a decoupled mechanism for passing data to them. Enter the `CommandContext` class, which provides data to the `Command`.

To actually run a command, the `Command` class provides the `Execute()` method which requires a `CommandContext` parameter. A `Command` subclass knows to expect a specific `CommandContext` subclass, and will dynamic_cast the parameter to the expected type. So to use a command, we need to create the appropriate CommandContext and populate it with data. To do so, the Command provides the `GetContextFactory()` method, which returns a subclassed `CommandContextFactory` instance. 

A `CommandContextFactory` provides both the information on the data required by a `CommandContext` and the means to create it. The method `CreateContext()` takes a `PropertyMap` - a key/value map - and returns the specialized CommandContext. The `GetPropertySpecifications()` method returns a list of `PropertySpecification`s that detail the keys and corresponding value types expected to be entered into the PropertyMap passed to `CreateContext()`.

Ok, so the process is:

1. At app startup, register Commands for types by calling `MenuManager::RegisterCommands()`
2. Get the Commands for an object by calling `MenuManager::GetCommands()`
3. To use a Command, first generate the CommandContextFactory by calling `Command::GetContextFactory()`
4. Get the specifications of required properties by calling `CommandContextFactory::GetPropertySpecifications()`
5. Populate a `PropertyMap` with the required properties.
6. Create a `CommandContext` by calling `CommandContextFactory::CreateContext()` with the populated `PropertyMap`
7. Execute the command by calling `Command::Execute()` with the `CommandContext`.

## CreateMenuForItems() and FindCommonCommands()

"Wait a second," you say, "I've been hoodwinked. Why, for all this devastatingly technical jibber jabber, you still haven't addressed the promised batch conversion feature." Now you just slow down there, and quit with those damned accusations. This PR does indeed address batch conversions. Why I wrote a couple functions for that very purpose.

There's the `MenuManager::FindCommonCommands()` function, which takes a vector of instance pointers and returns a vector of commands common to all of the instances. For example, if I pass a list of three VGMInstrSet instances, it will return commands including conversion to DLS and SF2, because they all offer that functionality. If I pass a VGMInstrSet and a VGMSeq, it will only give me the commands to close the files or save them in their original format.

I've also written the `MenuManager::CreateMenuForItems()`, which uses `FindCommonCommands()` to generate a QMenu. It also accepts a vector a instance pointers.

## Handling Polymorphism in Command Registration

One of the difficulties I ran into is that C++ offers limited RTTI (run-time type information). Specifically, while we can get the type name of an instance using typeid(), we cannot get its class hierarchy. In our use case, it's important that we be able to register commands to a common parent class, like VGMSeq, VGMInstrSet, and VGMSampColl. But if `typeid()` will only tell us that a VGMFile instance is a `RareSnesSeq` and not a `VGMSeq`, what are we to do?

My solution is to have two registries within MenuManager. The first is a simple map of type_index to a vector of Command which we will always check first. The second is a vector of pairs, where the first value is a function that tests whether an instance can be dynamic_cast to a type, and the second is the vector of Commands for that type. This is obviously slower than the map lookup, since we need to execute every test function until we get a match, but it works. The other downside is that we need to limit this functionality to a specific base class, as we can't use `dynamic_cast` on a void*; it must know the base class. For our needs, this functionality works on VGMItem*. Of course, we could expand this to more base types if needed.

## Services and Dependency Injection

You'll notice that I called MenuManager a "service" and put it in a new "services" directory. The idea here is to move away from using singletons - especially the VGMRoot singleton - and toward dependency injection. I'm not actually using dependency injection in this PR; the MenuManager is owned by VGMFileList because I didn't want to increase the scope of this PR. Howeve,r the way this is structured allows us to use MenuManager as an injected dependency. I'm calling these injected dependencies "services" but I'm open to using another term.

## Is this an example of overengineering?

I wrestled with this concern. I don't think so. A major problem with the app is that most functionality is hidden behind right-click context menus. This is poor design. Users will be confused and/or never discover functionality. We should be updating the application menu with the same contextual menu items as a right click. I think this will make it easier to do that across the app.

## How Has This Been Tested?
Tested on MacOS Sonoma x64/arm64 and Windows 10.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/services/MenuManager.h
+++ b/src/ui/qt/services/MenuManager.h
@@ -1,0 +1,292 @@
+/*
+* VGMTrans (c) 2002-2023
+* Licensed under the zlib license,
+* refer to the included LICENSE.txt file
+ */
+
+#pragma once
+
+#include "common.h"
+#include <typeindex>
+#include <typeinfo>
+#include <QMenu>
+#include "Root.h"
+#include "util/UIHelpers.h"
+
+#include <ghc/filesystem.hpp>
+
+namespace fs = ghc::filesystem;
+
+class VGMItem;
+class VGMFile;
+class VGMSeq;
+class VGMInstrSet;
+class VGMSampColl;
+class RawFile;
+class CommandSeparator;
+
+/**
+ * Represents the type of value a property holds in a CommandContext's PropertyMap.
+ */
+enum class PropertySpecValueType {
+  Path,           ///< Represents a file path or directory path if an ItemList property is present with > 1 items
+  DirPath,        ///< Represents a directory path
+  ItemList,       ///< Represents a list of items, such as selected files in the VGMFileListView
+};
+
+/**
+ * A variant type used to store different types of property values in a PropertyMap
+ */
+using PropertyValue = variant<
+    int,
+    float,
+    string,
+    shared_ptr<vector<VGMFile*>>,
+    shared_ptr<vector<VGMSeq*>>,
+    shared_ptr<vector<VGMInstrSet*>>,
+    shared_ptr<vector<VGMSampColl*>>,
+    shared_ptr<vector<RawFile*>>
+>;
+
+using PropertyMap = map<string, PropertyValue>;
+
+/**
+ * The base class for a command context. The command context provides an interface for commands
+ * to receive data. CommandContexts are created indirectly by CommandContextFactories via the CreateContext() method.
+ */
+class CommandContext {
+public:
+  virtual ~CommandContext() = default;
+};
+
+/**
+ * PropertySpecifications specify the keys and values types required for the PropertyMap that will
+ * be passed into a CommandContextFactory's CreateContext() method. A CommandContextFactory will
+ * expose these specifications via the GetPropertySpecifications() method
+ */
+struct PropertySpecification {
+  string key;
+  PropertySpecValueType valueType;
+  string description;
+  PropertyValue defaultValue;
+};
+
+using PropertySpecifications = vector<PropertySpecification>;
+
+/**
+ * The base class of a factory for creating CommandContext objects. Implementations of this class are responsible
+ * for creating a CommandContext that will be passed into a Command's Execute() method. It does this via the
+ * CreateContext() method, which expects a PropertyMap. The key and value types required to be set in the PropertyMap
+ * are given via the GetPropertySpecifications() method.
+ */
+class CommandContextFactory {
+public:
+  virtual ~CommandContextFactory() = default;
+
+  /**
+   * Creates a CommandContext object based on the provided properties.
+   * @param properties A map of properties to be used for context creation.
+   * @return A shared pointer to the created CommandContext object.
+   */
+  [[nodiscard]] virtual shared_ptr<CommandContext> CreateContext(const PropertyMap& properties) const = 0;
+
+  /**
+   * Retrieves the specifications for the properties required to be set in the PropertyMap passed to CreateContext().
+   * @return A list of property specifications.
+   */
+  [[nodiscard]] virtual PropertySpecifications GetPropertySpecifications() const = 0;
+};
+
+
+/**
+ * Base class for all commands. A Command is the same thing as a menu action. The GetContextFactory() method
+ * provides a factory that can be used to generate the CommandContext required by the Execute method.
+ */
+class Command {
+public:
+  virtual ~Command() = default;
+
+  /**
+   * Executes the command within the given context.
+   * @param context A reference to the CommandContext which will provide the data needed to execute the command.
+   */
+  virtual void Execute(CommandContext& context) = 0;
+
+  /**
+   * Retrieves the name of the command to appear in the UI. Commands must use unique Names, or they will
+   * break the behavior of batch command processing (an example of batch processing is
+   * selecting multiple sequences and executing the "Save to MIDI" command).
+   * @return The name of the command.
+   */
+  [[nodiscard]] virtual string Name() const = 0;
+
+  /**
+   * Gets the factory for creating a command context specific to this command.
+   * @return A shared pointer to the CommandContextFactory.
+   */
+  [[nodiscard]] virtual shared_ptr<CommandContextFactory> GetContextFactory() const = 0;
+};
+
+
+/**
+ * MenuManager is the service responsible for the registration and retrieval of commands.
+ */
+ class MenuManager {
+public:
+  using CheckFunc = function<bool(VGMItem*)>;
+  using CommandsList = vector<shared_ptr<Command>>;
+  using CheckedTypeEntry = pair<CheckFunc, CommandsList>;
+
+  MenuManager();
+
+  /**
+   * Registers a command for the template parameter type.
+   * @param command The command to be registered.
+   */
+  template<typename T>
+  void RegisterCommand(shared_ptr<Command> command);
+
+  /**
+   * Registers multiple commands at once for the template parameter type.
+   * @param commands A vector of commands to be registered.
+   */
+  template<typename T>
+  void RegisterCommands(const vector<shared_ptr<Command>>& commands);
+
+  /**
+   * Retrieves the list of commands suitable for a given item.
+   * @param item The VGMItem for which commands are requested.
+   * @return A vector of shared pointers to the commands.
+   */
+  template<typename T>
+  vector<std::shared_ptr<Command>> GetCommands(T* item) {
+    if (!item) {
+      return {}; // Return empty vector if item is nullptr
+    }
+
+    std::type_index typeIndex(typeid(*item));
+
+    // Check direct type associations first
+    auto it = commandsForType.find(typeIndex);
+    if (it != commandsForType.end()) {
+      return it->second;
+    }
+
+    // Special handling for VGMItem types
+    auto vgmItem = dynamic_cast<VGMItem*>(item);
+    if (vgmItem) {
+      // Check for matches using full class hierarchy
+      for (const auto& pair : commandsForCheckedType) {
+        if (pair.first(vgmItem)) {
+          return pair.second;
+        }
+      }
+    }
+
+    // Return empty vector if no commands are found
+    return {};
+  }
+
+  /**
+ * Given a list of objects, returns the list of Commands shared by all the instance types, as registered in the
+ * MenuManager.
+ * @param items The list of objects for which to find common Commands
+ * @param commandManager A reference to a MenuManager
+ * @return A list of commands shared by all items, ordered as they are in the first item's Command registry.
+   */
+  template <typename T>
+  vector<shared_ptr<Command>> FindCommonCommands(const vector<T*>& items) {
+    if (items.empty()) {
+      return {};
+    }
+
+    vector<shared_ptr<Command>> commonCommands;
+    // Get commands of the first file as the reference order
+    auto referenceCommands = GetCommands(items.front());
+
+    // Iterate over each command in the reference list
+    for (const auto& refCmd : referenceCommands) {
+      string refCmdName = refCmd->Name();
+      bool isCommon = true;
+
+      // Check if this command is present in all other files
+      for (auto* item : items) {
+        auto commands = GetCommands(item);
+        if (none_of(commands.begin(), commands.end(),
+                    [&refCmdName](const shared_ptr<Command>& cmd) { return cmd->Name() == refCmdName; })) {
+          isCommon = false;
+          break;
+        }
+      }
+
+      if (isCommon) {
+        commonCommands.push_back(refCmd);
+      }
+    }
+    return commonCommands;
+  }
+
+  template <typename T>
+  QMenu* CreateMenuForItems(shared_ptr<vector<T*>> selectedFiles) {
+
+    auto menu = new QMenu();
+    if (selectedFiles->empty()) {
+      return nullptr;
+    }
+    auto commands = FindCommonCommands(*selectedFiles);
+
+    for (const auto& command : commands) {
+
+      // Handle CommandSeparator
+      if (dynamic_cast<CommandSeparator*>(command.get()) != nullptr) {
+        menu->addSeparator();
+        continue;
+      }
+
+      auto contextFactory = command->GetContextFactory();
+      auto propSpecs = contextFactory->GetPropertySpecifications();
+
+      menu->addAction(command->Name().c_str(), [this, command, selectedFiles, propSpecs, contextFactory] {
+        PropertyMap propMap;
+        for (const auto& propSpec : propSpecs) {
+
+          bool isDirPath = false;
+          switch (propSpec.valueType) {
+            case PropertySpecValueType::DirPath:
+              isDirPath = true;
+            case PropertySpecValueType::Path:
+              if (isDirPath || selectedFiles->size() > 1) {
+                fs::path dirpath = fs::path(OpenSaveDirDialog());
+                if (dirpath.string().empty()) {
+                  return;
+                }
+                propMap.insert({ propSpec.key, dirpath });
+              } else {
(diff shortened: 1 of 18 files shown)


sykhro:
Great work, it's nice to see things moving again!
I will review this in more detail tomorrow but it looks good at first glance.

> One of the difficulties I ran into is that C++ offers limited RTTI (run-time type information). Specifically, while we can get the type name of an instance using typeid(), we cannot get its class hierarchy. In our use case, it's important that we be able to register commands to a common parent class, like VGMSeq, VGMInstrSet, and VGMSampColl. But if typeid() will only tell us that a VGMFile instance is a RareSnesSeq and not a VGMSeq, what are we to do?

With the current architecture of things I can't really think of simpler solutions. A clean-ish C++ way of doing this is through type-erasure (something like [this](https://gist.github.com/mrkline/114fa9963a6615231c40247e129c21b4)).
It would essentially be the same thing behind the scenes, it just makes the "figuring out the type" thing less explicit. 

>  The idea here is to move away from using singletons - especially the VGMRoot singleton - and toward dependency injection. ... I'm calling it a service, but I'm happy to call it something else.

This is the direction I wanted to take too (e.g. logging in the `refactor` branch, even if it's actually a singleton in disguise...). `services` sounds fine

mikelow:
> With the current architecture of things I can't really think of simpler solutions. A clean-ish C++ way of doing this is through type-erasure (something like [this](https://gist.github.com/mrkline/114fa9963a6615231c40247e129c21b4)).
> It would essentially be the same thing behind the scenes, it just makes the "figuring out the type" thing less explicit.

I'm still trying to grok this. I found the author's talk [Inheritance is the Base Class of Evil](https://www.youtube.com/watch?v=bIhUE5uUFOA). Think I'm going to need a good sleep before I can fully follow it.

If you know how to apply this pattern to improve the implementation, please let me know! Sounds like it's a powerful way to use polymorphism without class inheritance. My feeble brain hasn't grasped yet how to put it into practice.

> This is the direction I wanted to take too (e.g. logging in the refactor branch, even if it's actually a singleton in disguise...). services sounds fine

Excellent. Not surprised we're on the same page. If you know of a more common naming convention, let me know. If we go with "service", I'm debating renaming `MenuManager` to `MenuService`.

sykhro:
> I'm still trying to grok this. I found the author's talk [Inheritance is the Base Class of Evil](https://www.youtube.com/watch?v=bIhUE5uUFOA). Think I'm going to need a good sleep before I can fully follow it.

Hehe, yes, it's not straightforward; I hope I didn't send you down a rabbit hole! If you're open to considering c++-idiom alternatives, there's [double dispatch](https://eli.thegreenplace.net/2016/a-polyglots-guide-to-multiple-dispatch/) too (arguably less complex than the above)

> If you know how to apply this pattern to improve the implementation, please let me know! 

The problem is finding the common interface for the class to hide. If we have a `CommandTarget` as `object_t`, then the common function could be `CommandTarget::execute`, calling a free function e.g. `executeCommand`. So you could have `void executeCommand(VGMSeq *, ...` which should be picked automatically when you invoke `executeCommand` on a  `RareSnesSeq`. But then it becomes more complicated to.. actually run the command. Welp. I will think about this some more

> Excellent. Not surprised we're on the same page. If you know of a more common naming convention, let me know. If we go with "service", I'm debating renaming `MenuManager` to `MenuService`.

I honestly don't know. I have seen *Manager and *Service in codebases, they are the go-to "I don't know how to name this thing but it feels important". I have to say that *Service sounds like something that's running in the background, but I don't think we'll ever have anything of the sort.

mikelow:
One of my goals with the MenuManager class is to 

> Hehe, yes, it's not straightforward; I hope I didn't send you down a rabbit hole! If you're open to considering c++-idiom alternatives, there's [double dispatch](https://eli.thegreenplace.net/2016/a-polyglots-guide-to-multiple-dispatch/) too (arguably less complex than the above)

That is a cool pattern, but it's not clear to me how I would utilize it. If there were a way, it seems it would requiring updating the classes which register with the MenuManager, something I'm trying to avoid.

> The problem is finding the common interface for the class to hide. If we have a CommandTarget as object_t, then the common function could be CommandTarget::execute, calling a free function e.g. executeCommand. So you could have void executeCommand(VGMSeq *, ... which should be picked automatically when you invoke executeCommand on a RareSnesSeq. But then it becomes more complicated to.. actually run the command. Welp. I will think about this some more

I'll go over the material and try to make sense of all of this. The double dispatch pattern was easy to understand and explained well. The type-erasure stuff, not so much.

> I honestly don't know. I have seen *Manager and *Service in codebases, they are the go-to "I don't know how to name this thing but it feels important". I have to say that *Service sounds like something that's running in the background, but I don't think we'll ever have anything of the sort.

I've seen "service" used for dependency injected modules outside of C++. I suppose there isn't any standard. My qualm with "manager" is that it's too specific in its meaning. Granted, service does have that operating system "background service" association.

mikelow:
> The problem is finding the common interface for the class to hide. If we have a CommandTarget as object_t, then the common function could be CommandTarget::execute, calling a free function e.g. executeCommand. So you could have void executeCommand(VGMSeq *, ... which should be picked automatically when you invoke executeCommand on a RareSnesSeq. But then it becomes more complicated to.. actually run the command. Welp. I will think about this some more

Well the way I was thinking about it, the common interface would be a getCommands() function that returns a different list of Commands for each type. The problem I'd like to solve is the coupling of MenuManager with VGMItem; so that we can register commands for other parent classes and still have them work for their subclasses. Even so I can't wrap my head around this design pattern - I've been fiddling around, but the example code is too different from what I'm trying to do and I find it unintuitive.

sykhro:
> it would requiring updating the classes that register with the MenuManager, which is something I'm trying to avoid.

ah ok, fair enough
 
 > the common interface would be a getCommands() function that returns a different list of Commands for each type
 
 Juggling types like this in c++ is basically only possible as long as the information you need is known at compile time
 
> The problem I'd like to solve is the coupling of MenuManager with VGMItem; so that we can register commands for other parent classes and still have them work for their subclasses. Even so I can't wrap my head around this design pattern - I've been experimenting, but the example code is too different from what I'm trying to do and I find it unintuitive.

I have tried but didn't end up with anything useful. I'm finding it hard to find something else that would work with everything that we have in place right now, so this is ok for me

mikelow:
> it would requiring updating the classes that register with the MenuManager, which is something I'm trying to avoid.

Well, I'm not _completely_ opposed to this.  I'm trying to retain the separation between UI logic and the core logic. One easy way to solve this is to just make a dummy base class everything inherits from. Then dynamic_cast works with all of them. Doesn't seem like a great solution, but neither is hardcoding VGMItem.

I didn't look into whether we could register the parent class somehow instead of hardcoding it. That would be cleaner.

> Juggling types like this in c++ is basically only possible as long as the information you need is known at compile time

Yeah, it occurred to me the type-erasure design probably can't fix this problem. At that point, I think it would just get us constant time resolution of the polymorphic class, as opposed to O(n) where we iterate over a vector of functions. I'm all for it, if we can figure out a way, but it doesn't solve what I consider the bigger design issue.

Practically, I don't foresee the O(n) lookup being a hindrance, as all leaf subclasses are still resolved in constant time, and I don't expect there will be many more types that need to be resolved via the checkFunc vector. Also, C++ is fast.

> I have tried but didn't end up with anything useful. I'm finding it hard to find something else that would work with everything that we have in place right now, so this is ok for me

👍 Don't hold back your critiques.

mikelow:
> I didn't look into whether we could register the parent class somehow instead of hardcoding it. That would be cleaner.

I've just added a commit that does exactly this. Now RegisterCommands() takes an optional second template argument to specify the base class for which to register the commands (I've removed the singular RemoveCommand() method). The `commandsForCheckedType` property is now a map, associating a base class type to its former type - a vector of pairs, where the first item is a function to test if an instance belongs to class, and the second item is a vector of Commands for that class.

The only difference for client usage now is that CreateMenuForItems() expects a template argument to specify which base type to use for dynamic type lookup. This type will need to match whichever base type the commands were registered with. Of course, if the type was not registered with a base type, it will not matter.

I've also added some meta programming magic in the form of `has_getname`, which, when used with constexpr, let's us check if a template argument has the method GetName() at compile time. Now, `CreateMenuForItems()` will work for any type, whereas before it would have failed if the type did not define GetName(). If GetName() is missing, it will simply _not_ provide a suggested save file name when opening a save file dialog.

mikelow:
Waiting for approval. Are you still interested in exploring refactoring?

sykhro:
I was thinking about this earlier. imo this solution is okay, I feel like it's better to get the ball rolling rather than lacking features forever :)

src/ui/qt/services/commands/SaveCommands.h
```diff
@@ -0,0 +1,206 @@
...
+
+  void Execute(CommandContext& context) override {
+    auto& vgmContext = dynamic_cast<SaveCommandContext<TInContext>&>(context);
+    const auto& files = vgmContext.GetItems();
+    const auto& path = vgmContext.GetPath();
+
+    if (alwaysSaveToDir) {
+      // alwaysSaveToDir indicates we were given a directory path and Save() also expects a directory path
+      for (auto file : files) {
+        auto specificFile = dynamic_cast<TSavable*>(file);
```
sykhro, line 112: I only have a nitpick: `Savable` or `Saveable`?

src/ui/qt/workarea/VGMFileListView.cpp
```diff
@@ -26,7 +26,7 @@ VGMFileListModel::VGMFileListModel(QObject *parent) : QAbstractTableModel(parent
 
 QVariant VGMFileListModel::data(const QModelIndex &index, int role) const {
   if (!index.isValid()) {
-    return QVariant();
+    return {};
```
sykhro, line 29: This was `QVariant()` because codegen with `{}` used to be horrendous (something about copy construction in Qt, I don't remember). Did you observe a performance difference?

src/ui/qt/workarea/VGMFileListView.cpp
```diff
@@ -154,48 +154,24 @@ void VGMFileListView::onHeaderSectionResized(int index, int oldSize, int newSize
 }
 
 void VGMFileListView::itemMenu(const QPoint &pos) {
-  auto element = indexAt(pos);
-  if (!element.isValid())
-    return;
 
-  VGMFile *file = qtVGMRoot.vVGMFile[element.row()];
-  if (!file) {
+  auto selectedFiles = make_shared<vector<VGMFile*>>();
```
sykhro, line 158: Any reason why this can't be a `unique_ptr` (or better, just a `vector` value?)

src/ui/qt/services/MenuManager.h
```diff
@@ -0,0 +1,302 @@
...
+  template <typename Base>
+  struct has_getname<Base, decltype(std::declval<Base>().GetName(), void())> : std::true_type {};
+
+  /**
+ * Create a QMenu populated with actions for the commands common to all of the instances in the provided vector
+ * @param items The list of objects the actions of the QMenu should pertain to
+ * @return A QMenu populated with actions for the common commands, in the same order as they were registered for the first item
+   */
+  template <typename Base, typename T = Base>
+  QMenu* CreateMenuForItems(shared_ptr<vector<T*>> items) {
```
sykhro, line 241: since the vector isn't really owned by this,  `std::vector<T*> &` is more appropriate than `shared_ptr`

src/ui/qt/services/commands/SaveCommands.h
mikelow: Interesting lexical question. I had to research this. It seems both "savable" and "saveable" are acceptable spellings. I've seen some sources claim the meaning of the two words is different, but best I can tell that's not true. Savable seems to be more common in American English. The American English MacOS spell checker doesn't even like "saveable".

There are, however, twice as many usages of "saveable" in github code than "savable".  If "Savable" rubs you the wrong way, I will happily change it to "Saveable". 

image

sykhro: ah, interesting, it felt like a typo to my non-native brain. let's leave it like this then

src/ui/qt/workarea/VGMFileListView.cpp
mikelow: Not sure if I noticed a performance difference, but I'll switch it back if that's even potentially an issue. To be clear, I made this change because of a clang-tidy warning, which I assumed meant it would be functionally equivalent.

_"Clang-tidy: Void repeating the return type from the declaration; use a braced initializer list instead"_

src/ui/qt/workarea/VGMFileListView.cpp
mikelow: Because that would make too much sense. I need to needlessly complicate this.

😉 I'll change it.

src/ui/qt/services/MenuManager.h
mikelow: 💯

src/ui/qt/workarea/VGMFileListView.cpp
mikelow: Oops, I forgot that I did have a reason for making this a shared_ptr: CreateMenuForItems() is passing a closure to QMenu::addAction() which utilizes the vector of VGMFile*. This closure will be invoked beyond the lifetime of selectFiles` scope.

If you think it would be cleaner to copy the vector, though. I could do that.

src/ui/qt/services/MenuManager.h
mikelow: See above.

mikelow:
> I was thinking about this earlier. imo this solution is okay, I feel like it's better to get the ball rolling rather than lacking features forever :)

If you think this code smells, I don't mind taking it back to the drawing board. I just don't know how it could be refactored.

src/ui/qt/workarea/VGMFileListView.cpp
sykhro: I had a look at QVariant and it looks saner nowadays, looks like this is no longer an issue

src/ui/qt/workarea/VGMFileListView.cpp
sykhro: Ah, whoops, makes sense that it would go past the selectFiles lifetime. Then I'd say that a shared pointer is the proper abstraction for ownership here

sykhro:
> > I was thinking about this earlier. imo this solution is okay, I feel like it's better to get the ball rolling rather than lacking features forever :)
> 
> If you think this code smells, I don't mind taking it back to the drawing board. I just don't know how it could be refactored.



> > I was thinking about this earlier. imo this solution is okay, I feel like it's better to get the ball rolling rather than lacking features forever :)
> 
> If you think this code smells, I don't mind taking it back to the drawing board. I just don't know how it could be refactored.

It took me a bit to understand it but I wouldn't say it's bad. We are in a very dynamic situation so any other solution would be very similar imo. I'm happy to merge this

PR #420, opened 2023-12-21
use dataChanged() on raw file removal, for now

mikelow:
For some reason (probably related to fiddling with the installed build tools) I started seeing exceptions thrown on the Windows build in RawFileListViewModel::RemoveRawFile() upon loading a file that gets unpacked. It looks like we are trying to remove rows at an index that doesn't necessarily exist. The code in this function is odd in that there's no guarantee a removed RawFile will be at the last index, and besides this, it appears that we're supposed to:
1. Call beginRemoveRows()
2. Remove the rows from the data source
3. Call endRemoveRows()

At any rate, we have the same non-adherence in AddRawFile() and it's not causing a fuss, possibly because in that case, the last index is actually what is updated every time.

Anyway, for now, I've just added a dataChanged() in place of begin/endRemoveRows() and it makes the issue go away.

## How Has This Been Tested?
Ran it on Windows 10.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/workarea/RawFileListView.cpp
+++ b/src/ui/qt/workarea/RawFileListView.cpp
@@ -45,8 +45,7 @@ void RawFileListViewModel::AddRawFile() {
 void RawFileListViewModel::RemoveRawFile() {
   int position = static_cast<int>(qtVGMRoot.vRawFile.size());
   if (position >= 0) {
-    beginRemoveRows(QModelIndex(), position, position);
-    endRemoveRows();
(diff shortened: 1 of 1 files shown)


sykhro:
> The code in this function is odd in that there's no guarantee a removed RawFile will be at the last index.

yeah, I'm guessing it was problematic to get the index of the file that was being removed. or we just never bothered...

sykhro (approved):

PR #421, opened 2023-12-21
Remove everything related to portamento time mode

mikelow:
It's no longer needed.

### DIFF
--- a/src/main/MidiFile.h
+++ b/src/main/MidiFile.h
@@ -11,7 +11,6 @@ class MidiTrack;
 class MidiEvent;
 class DurNoteEvent;
 class NoteEvent;
-enum class PortamentoTimeMode: uint8_t;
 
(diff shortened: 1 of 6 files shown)


src/main/MidiFile.h
```diff
@@ -47,7 +46,6 @@ typedef enum {
   MIDIEVENT_PORTAMENTO,
   MIDIEVENT_PORTAMENTOTIME,
   MIDIEVENT_PORTAMENTOTIMEFINE,
-  MIDIEVENT_PORTAMENTOTIMEMODE,
```
sykhro, line 50: I hope that nothing depended on the internal value of this enum

sykhro (approved):

src/main/MidiFile.h
mikelow: Nah. Don't see any references to the internal value.

PR #422, opened 2023-12-21
Remove the old Menu macros and WTL headers

mikelow:
They're all unused now that the WTL version is out.


sykhro:
Could you also remove [dxAug07](https://github.com/vgmtrans/vgmtrans/tree/master/lib/dxAug07)?

sykhro (changes requested):
We also have [dxAug07](https://github.com/vgmtrans/vgmtrans/tree/master/lib/dxAug07) which is now unused and needs removal

mikelow:
Updated. Happy to get rid of it.

sykhro (approved):
That should be everything then

PR #423, opened 2023-12-21
Replace all instances of wide chars with normal ass chars

mikelow:
This replaces every `wchar_t`, every `wstring`, every `wostream`, every L""/L'' literal and every function call related to wide chars with their plain jane `char` equivalent. Some unnecessary conversion calls are removed.

I did not go through and clean up all of the string handling code. There is surely plenty of bad string-related code in here, but now it's at least all working with string/char instead of a gross mix of wide chars and normal chars.

## Motivation and Context
We don't need wide chars. They're gross. Get rid of  them.

## How Has This Been Tested?
Tested on MacOS and Windows 10.

Loaded psf, psf2, spc, 2sf, gsf files, including instances of "minixsf" files that recursively load another file. Tested saving various formats to disk. Tested adding unicode characters to the path of files being loaded. No issues.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/DLSFile.cpp
+++ b/src/main/DLSFile.cpp
@@ -137,7 +137,7 @@ int DLSFile::WriteDLSToBuffer(vector<uint8_t> &buf) {
 }
 
 // I should probably make this function part of a parent class for both Midi and DLS file
-bool DLSFile::SaveDLSFile(const std::wstring &filepath) {
+bool DLSFile::SaveDLSFile(const std::string &filepath) {
   vector<uint8_t> dlsBuf;
   WriteDLSToBuffer(dlsBuf);
   return pRoot->UI_WriteBufferToFile(filepath, dlsBuf.data(), static_cast<uint32_t>(dlsBuf.size()));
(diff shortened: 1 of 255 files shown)


sykhro (approved):
Did this (sloppily) in https://github.com/vgmtrans/vgmtrans/commit/7c3d5b8b07013bb9ba92c7e947939170e028d07d#diff-b3268da32433d18325dfd8e3039dcdaae1a2509133be04a40eeb32c9455bfc17 a few years ago. There were no issues then, so it should be ok :) we're bound to get some unicode bugs anyway

mikelow:
Don't assume mine is any less sloppy (although looks like you overwrote both L" and l" ;). is this from your refactor branch?

sykhro:
> Don't assume mine is any less sloppy (although looks like you overwrote both L" and l" ;). is this from your refactor branch?

Yes. A couple commits later there's one that replaces every stringstream with calls to libfmt, but I don't think we can just re-apply it (it's been.. 4 years?! 2019)

PR #424, opened 2023-12-26
Add RSNLoader

mikelow:
This provides decompression for the rsn format used for spc collections. Rsn files are just rar archives, so this also adds support for rar files. I'm using the open source LPGL [unarr library](https://github.com/selmf/unarr) and opted to directly include the relevant source files, removing as much non-rar related code as possible. Alternatively, we could just import the project as a gitsubmodule 🤷‍♂️.

In terms of licensing, I did some research and I think we're in compliance. I've included the original license and an explaination in the lib/unarr/README.md with a note that this is a modified version of the code, a link to the original repo and reference to the unarr license. Yes, the original rar format itself has a restrictive license, but so long as unarr was clean-room reverse-engineered, and I expect it is given that it's based on the source code of popular MacOS app - [The Unarchiver](https://code.google.com/archive/p/theunarchiver/), everything should be hunky dory.

## Motivation and Context
rsn is a popular format for spc archives

## How Has This Been Tested?
Tested on MacOS x86/arm64 and Windows 10.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/loaders/RSNLoader.cpp
+++ b/src/main/loaders/RSNLoader.cpp
@@ -1,0 +1,49 @@
+/*
+* VGMTrans (c) 2002-2023
+* Licensed under the zlib license,
+* refer to the included LICENSE.txt file
+ */
+
+#include "common.h"
+#include "RSNLoader.h"
+#include "Root.h"
+#include "unarr.h"
+
+using namespace std;
+
+#define FILE_SIGNATURE_SIZE 7
+
+PostLoadCommand RSNLoader::Apply(RawFile *file) {
+
+  // Read header
+  uint8_t rarHeader[FILE_SIGNATURE_SIZE];
+  file->GetBytes(0, FILE_SIGNATURE_SIZE, rarHeader);
+
+  if (memcmp(rarHeader, "Rar!\x1A\x07\x00", FILE_SIGNATURE_SIZE) != 0) {
+    return KEEP_IT;
+  }
+
+  auto path = file->GetFullPath();
+  auto stream = ar_open_file(path.c_str());
+  ar_archive *ar = ar_open_rar_archive(stream);
+
+  while (ar_parse_entry(ar)) {
+    size_t size = ar_entry_get_size(ar);
+    const char *raw_filename = ar_entry_get_name(ar);
+    if (raw_filename) {
+      pRoot->AddLogItem(new LogItem(string("Decompressing file from rar archive:" + string(raw_filename)),
+                      LOG_LEVEL_INFO, "RSNLoader"));
+    } else {
+      continue;
+    }
+    auto buffer = new uint8_t[size];
+    if (!ar_entry_uncompress(ar, buffer, size)) {
+      pRoot->AddLogItem(new LogItem(string("Error decompressing file from rar archive:" + string(raw_filename)),
+                                    LOG_LEVEL_ERR,"RSNLoader"));
+      delete[] buffer;
+      return KEEP_IT;
+    }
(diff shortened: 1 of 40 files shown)


src/main/loaders/RSNLoader.cpp
```diff
@@ -0,0 +1,49 @@
...
+  // Read header
+  uint8_t rarHeader[FILE_SIGNATURE_SIZE];
+  file->GetBytes(0, FILE_SIGNATURE_SIZE, rarHeader);
+
+  if (memcmp(rarHeader, "Rar!\x1A\x07\x00", FILE_SIGNATURE_SIZE) != 0) {
+    return KEEP_IT;
+  }
+
+  auto path = file->GetFullPath();
+  auto stream = ar_open_file(path.c_str());
```
sykhro, line 27: I'm guessing there should be a `ar_close_file()` somewhere?

sykhro:
> Alternatively, we could just import the project as a gitsubmodule 🤷‍♂️.

We won't update it often so I'd say it's okay to have it vendored.

I have to say, I'm not a fan of submodules for dependencies because they make life more complicated for CI (and sometimes for devs too, if you mix https and ssh in your environment). Ideally we'd want to use something like vcpkg, but we can't really afford a binary cache (you don't want to compile Qt locally) so it'd be more trouble than it's worth.

src/main/loaders/RSNLoader.cpp
mikelow: good catch 👍

sykhro:
The `/usr/bin/ld: src/main/unarr/libunarr.a: error adding symbols: file format not recognized` error is likely due to LTO being used to compile `unarr`.

mikelow:
> The /usr/bin/ld: src/main/unarr/libunarr.a: error adding symbols: file format not recognized error is likely due to LTO being used to compile unarr.

Oh, you're way ahead of me then. As you can see, I've been going down a CI rabbit hole. I thought maybe we weren't setting up the llvm 17 linker correctly. You'll have to forgive me that I'm unfamiliar with LTO. I'll look into the libunarr CMake scripts.

> p.s.: I would like to keep clang 17 on CI because currently it has better codegen than AppleClang

Yeah, I was planning to try to get clang 17 working again on Mac. Reverting was part of my debugging process. Still, I'm not sure why this solved the Mac build. Would LTO still explain it?

mikelow:
> The /usr/bin/ld: src/main/unarr/libunarr.a: error adding symbols: file format not recognized error is likely due to LTO being used to compile unarr.

Brilliant! Fixed it. I've reverted all changes to build.yml.

sykhro:
> Would LTO still explain it?

Yes, `lld` from AppleClang handles LLVM object files slightly differently.  

> Brilliant! Fixed it

Ahh, yes, it's been almost 10 years and this is still a bug around CMake+lld :D

btw, it would be nice to turn on LTO for the whole program, but I think we have some undefined behaviour somewhere. I will set up ASAN+UBSAN sometime soon, so that we can run the program with those and try to squish memory corruption bugs

mikelow:
> btw, it would be nice to turn on LTO for the whole program, but I think we have some undefined behaviour somewhere. I will set up ASAN+UBSAN sometime soon, so that we can run the program with those and try to squish memory corruption bugs

This is all immensely educational for me. Glad to have your C++ knowledge.

mikelow:
@sykhro sorry that I seemed to have mistakenly merged this before approval. Let me know if you want me to make any changes and/or revert.

PR #428, opened 2024-01-09
Various compile warning fixes

TheComputerGuy96:
This should fix the remaining compile warnings (check the commit messages for a better description) :frog: 

I wonder if I should split this PR though :thinking:


TheComputerGuy96:
Note: I just realized CI uses clang instead of gcc (which has better compiler warnings) so the warnings are only fully fixed for GCC :frog:

PR #433, opened 2024-02-04
Move CPS1 logic into separate files. Look away from the mess

mikelow:
Note that this PR is basing on top of the previous "okiadpcm" branch PR.

Adding sampled playback for CPS1 expanded the already bloated CPSScanner.cpp and CPSInstr.cpp files. Since the logic for CPS1 is unique, I've moved it into respective CPS1Scanner.cpp and CPS1Instr.cpp files and added a "CPS1" Format.

This leaves us in an awkward spot, as we would have both a "CPS1" format and a "CPS" format, not to mention a CPS1Scanner and a CPSScanner. Thus I've renamed the CPS format to CPS2, CPSScanner to CPS2Scanner and CPSinstr to CPS2Instr. Unfortunately, "CPS2" is partly a misnomer here as both files and formats cover CPS2/CPS3 and some ZN-1/ZN-2 games. Further muddying the waters, CPSSeq, at least for now, is still used by all formats above, and the CPSTrackV1/V2 classes do not align cleanly with any CPS version (there is a CPS2 game that uses the CPS3 sequence structure).

I throw my hands up in the air. I think eliminating the bloat is probably a net improvement.

## How Has This Been Tested?
Tested that CPS1/2/3 games are all loading and playing as expected.

### DIFF
--- a/bin/mame_roms.xml
+++ b/bin/mame_roms.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0" ?>
 <romlist ver="1">
-    <game name="1941" format="CPS" fmt_version="CPS1_2.00">
+    <game name="1941" format="CPS1" fmt_version="CPS1_2.00">
         <romgroup type="audiocpu" load_method="append" seq_table="0x1000" >
             <rom>41_9.12b</rom>
         </romgroup>
     </game>
-    <game name="1944" format="CPS" fmt_version="1.80">
+    <game name="1944" format="CPS2" fmt_version="1.80">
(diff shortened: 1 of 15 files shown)


sykhro (commented):
I don't have a strong opinion on this, I don't know the formats enough to propose a naming scheme. Happy to see the bloat gone.

I did a similar thing in the past, unifying all of the .*sf loaders while also getting rid of recursion. Wondering if it's worth rebasing on the current tree.

mikelow:
> I did a similar thing in the past, unifying all of the .*sf loaders while also getting rid of recursion. Wondering if it's worth rebasing on the current tree.

Definitely sounds like a worthy refactor.

PR #444, opened 2024-04-08
ci: add macos arm64 target

mikelow:
## Description
Add MacOS arm64 as a CI target.

Note that Github seems to be severely throttling their arm64 machines. I advise we allow ourselves to merge future pull requests regardless of whether the arm64 build is pending.

## Types of changes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,12 +6,15 @@ on:
   pull_request:
   workflow_dispatch:
 jobs:
-  x86_64-apple-darwin:
-    name: macOS 11.x+, Intel
-    runs-on: macos-13
+  macos:
+    name: macOS 11.x+
+    runs-on: ${{ (matrix.target == 'x86_64' && 'macos-13') || 'macos-14' }}
+    strategy:
+      matrix:
+        target: [ "x86_64", "arm64" ]
     env:
       osx_min_ver: "11.0"
-      osx_arch: "x86_64"
+      HOMEBREW_PREFIX: ${{ (matrix.target == 'x86_64' && '/usr/local') || '/opt/homebrew' }}
     steps:
       - uses: actions/checkout@v3
       - name: "Install dependencies"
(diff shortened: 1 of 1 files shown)


sykhro:
Note that with apple silicon a signature must be present: the app will crash on start otherwise. The user can do it locally with the codesign command.

mikelow:
I looked back at the discussion here #381 and did some testing against MacOS Sonoma. The situation appears to be even more restrictive than it was with past MacOS versions. There are two layers of "security" used to enforce Apple's $100 shakedown.

### 1) Quarantine.
If the app is downloaded from a browser, it will be marked quarantined (this doesn't happen if we download the app with wget or curl). MacOS will lie to the user that the app is corrupt and encourage the user to trash it... unless we pay $100 for the privilege to "notarize" our free and open source app.

image

On Sonoma, the only workaround seems to be:

`xattr -d com.apple.quarantine VGMTrans.app`


### 2) Certification
Once you get past the quarantine, the user is hit with yet another extremely misleading error:

image

To get past this one, the user is forced to self-sign the app with the command:

`codesign --force --deep -s - VGMTrans.app`

However, I suspect this isn't possible if the user doesn't have a Apple Developer certificate installed in their keychain, and they probably must also change the settings of their certificate to "Always Trust". I will test this later.

The following commands which used to provide a means for bypassing security checks no longer seem to have any effect on Sonoma:

`sudo spctl --master-enable`
`sudo spctl --global-enable`

It is possible to remove the failing certificate from the app via:

`codesign --remove-signature VGMTrans.app`

For Intel, this apparently helps provide a better error dialog that suggests the user can bypass the security check. But on an arm64 Sonoma machine we get the following message, no matter how we open the app, and no - we cannot bypass it:

image

### Where does this leave us?

Though it seems impractical to release an arm64 version unless we provide heavy-handed instructions, the arm64 build may still be informative for continuous integration purposes. Since it's takes over 20 minutes for arm builds to finish, I'd advise we simply ignore them for the sake of pull requests (unless we notice a failed build).

Alternatively, we could configure it to only run on the master branch.

mikelow:
Let me know what your thoughts are on adding this target to the CI.

We may not want to distribute the arm64 version - maybe if we provide instructions. But this question is separate from whether we want to include arm64 in our CI pipeline for testing.

sykhro:
If it's low maintenance, why not. I wouldn't make it available though, I can already see issues being opened because "the program doesn't start"...

mikelow:
Alright, going to merge this, with dash of hesitance.

PR #445, opened 2024-04-09
Change collection list double-click behavior to play/pause

mikelow:
## Description
Currently, double clicking a collection puts the collection name into an edit mode. I don't think this provides much value to the user.

I can think of two more-helpful responses to a double click:
1) play/pause the sequence player for the clicked collection
2) open up the hex view for the associated sequence file of the clicked collection

I went with 1.

I also changed the selection behavior to only allow selection of one collection at a time, since that is the paradigm we use with the VGMCollView (to the left of the VGMCollListView).

## How Has This Been Tested?
Played around with it. Haven't been able to break it.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/workarea/VGMCollListView.cpp
+++ b/src/ui/qt/workarea/VGMCollListView.cpp
@@ -102,12 +102,14 @@ VGMCollListView::VGMCollListView(QWidget *parent) : QListView(parent) {
   setItemDelegate(new VGMCollNameEditor);
 
   setContextMenuPolicy(Qt::CustomContextMenu);
-  setSelectionMode(QAbstractItemView::ExtendedSelection);
-  setEditTriggers(QAbstractItemView::DoubleClicked);
+  setSelectionMode(QAbstractItemView::SingleSelection);
+  setEditTriggers(QAbstractItemView::NoEditTriggers);
   setResizeMode(QListView::Adjust);
   setIconSize(QSize(16, 16));
   setWrapping(true);
 
+  connect(this, &QListView::doubleClicked, this,
+          &VGMCollListView::handlePlaybackRequest);
   connect(this, &QAbstractItemView::customContextMenuRequested, this,
           &VGMCollListView::collectionMenu);
   connect(model, &VGMCollListViewModel::dataChanged, [=]() {


sykhro:
> Currently, double clicking a collection puts the collection name into an edit mode. I don't think this provides much value to the user.

This is a standard UX mechanism: a slow double click triggers a different action than a regular double click. You can try it in the Windows shell: double click slowly on a file and it will let you edit its name, double click normally and it will open it in the default application.

I can agree with the argument that this is a relic of the past nobody knows about, I also agree that most people don't care about the hex view so option 1 is sensible.

sykhro (approved):

mikelow:
> edit: I wrote this comment before looking at the patch, I didn't recall edit being straight up wired to double click..

If your assumption was that this is default behavior, you are indeed correct. If we comment out the `setEditTriggers()` line entirely, we still see that double clicking leads to name editing.

Also, even if we comment out the setEditTriggers() line, Qt isn't making a distinction between a fast and slow double-click.

My point wasn't that edit-name-on-double-click is non-standard UX behavior, but that renaming collections serves little purpose for the user, therefore we should dedicate this common UX action to something that _is_ useful.

Let me know if you disagree.

mikelow:
Pretty sure we're on the same page, given your approval the PR. Merging.

apollolux:
> > Currently, double clicking a collection puts the collection name into an edit mode. I don't think this provides much value to the user.
> 
> This is a standard UX mechanism: a slow double click triggers a different action than a regular double click. You can try it in the Windows shell: double click slowly on a file and it will let you edit its name, double click normally and it will open it in the default application.
> 
> I can agree with the argument that this is a relic of the past nobody knows about and it's easier to remove it, I also agree that most people don't care about the hex view so option 1 is sensible.
> 
> edit: I wrote this comment before looking at the patch, I didn't recall edit being straight up wired to double click..

"Slow double click" You mean two single clicks, one to select a thing and another literally any time after it's selected beyond the (system's) double-click timing threshold? This works on Mac and Linux, too, BTW, when you don't have that weird "single-click activates things" behavior turned on like recent OS versions of everything keep trying to make happen to simulate web browsing.

mikelow:
> "Slow double click" You mean two single clicks, one to select a thing and another literally any time after it's selected beyond the (system's) double-click timing threshold? This works on Mac and Linux, too, BTW, when you don't have that weird "single-click activates things" behavior turned on like recent OS versions of everything keep trying to make happen to simulate web browsing.

I know what you're referring to, but in this case, the edit mode is triggered with a double-click (without a slow/fast distinction ).

PR #446, opened 2024-04-09
Fix mdi window bug, add unified vgm file selection behavior

mikelow:
**Update: I discovered that this PR is vulnerable to a separate dangling pointer bug that can cause a crash. It's fixed here: #447**

## Description

I would have split this into two PRs, but there's overlap.

1) There's a bug whereby closing an MDI window causes every other open MDI window to become windowed instead of maximized. To fix this, I've added a new ensureMaximizedSubWindow() function that gets called every time the subWindowActivated signal is fired by MdiArea.

2) The WTL version had a UI behavior whereby selecting a file in the vgm file list or focusing its MDI window caused both to be selected/focused. I've added this behavior back in and extended it to the VGMCollView. 

## How Has This Been Tested?
Thorough clicking and clacking.

The MDI window bug. Closing a window often results in something like this:
image

## Types of changes

## Checklist:

### DIFF
--- a/src/main/VGMColl.cpp
+++ b/src/main/VGMColl.cpp
@@ -611,3 +611,24 @@ bool VGMColl::OnSaveAllSF2() {
   }
   return true;
 }
+
+// A helper lambda function to search a vector of VGMFile* for the file
+template<typename T>
+bool contains(const std::vector<T*>& vec, const VGMFile* file) {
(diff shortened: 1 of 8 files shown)


mikelow:
Good point. I'll test it.

mikelow:
Confirmed this fixes the issue on Windows.

PR #447, opened 2024-04-09
Fix VGMCollView not updating its model immediately after VGMColl deletion

mikelow:
## Description
Turns out there is a crashing bug in #446. The issue is a dangling pointer in VGMCollViewModel that #446 accesses. When a user deletes a collection, VGMCollViewModel does not check if the deleted collection is the one pointed to by its m_coll property.

In #446, if the user selects a collection, then deletes it by removing any of its files, then attempts to open the MDI window of any file, the VGMCollViewModel::containsVGMFile() may cause a crash when it dereferences m_coll.

This fixes the bug by handling the UI_RemoveVGMColl() callback in VGMCollView to check if the removed collection is selected. If so, it updates the model accordingly. Yes, we should probably be using smart pointers here and everywhere.

Note that I removed `UI_RemovedVGMColl()` and upgraded `UI_RemoveVGMColl(VGMColl*)` to a signal in its place.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/Root.h
+++ b/src/main/Root.h
@@ -41,7 +41,7 @@ class VGMRoot {
   void AddVGMFile(VGMFile *theFile);
   void RemoveVGMFile(VGMFile *theFile, bool bRemoveFromRaw = true);
   void AddVGMColl(VGMColl *theColl);
-  void RemoveVGMColl(VGMColl *theFile);
+  void RemoveVGMColl(VGMColl *theColl);
   void AddLogItem(LogItem *theLog);
   void AddScanner(const std::string &formatname);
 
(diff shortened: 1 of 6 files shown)


sykhro:
>  Yes, we should probably be using smart pointers here and everywhere

Heh, Qt has its own memory model and I've always found that mixing those led to insanely complicated bugs. I know that they embraced the STL some more with Qt6 so perhaps is worth looking into. 100% agree on using smart pointers in the "backend"

sykhro (approved):

mikelow:
> Heh, Qt has its own memory model and I've always found that mixing those led to insanely complicated bugs. I know that they embraced the STL some more with Qt6 so perhaps is worth looking into. 100% agree on using smart pointers in the "backend"

Hmm, does Qt discourage the user of smart pointers for properties of subclassed widgets pointing to non-Qt classes (in this case, a VGMColl)?

PR #448, opened 2024-04-10
Fix a Windows bug causing new subwindows to flash

mikelow:
## Description
There is a Windows-specific bug that occurs when a VGMFile is opened into the MdiArea. A new window will flash and then disappear before the sub window is created.

The culprit is an unnecessary call to `m_hexScrollArea->show()` in `VGMFileView.cpp`.

I've removed this call, and I've also changed the call in MdiArea from `tab->show()` to `tab->showMaximized()`. While this change isn't necessary for the window to display maximized, it seems more reflective of our intentions. I'm confused why windows are appearing maximized even when we call `tab->show()`. Unfortunately, this change doesn't obviate the need for my last fix with `ensureMaximizedSubWindow()`.

## How Has This Been Tested?
Tested on Windows 10

## Types of changes

## Checklist:


sykhro:
> I'm confused why windows are appearing maximized even when we call tab->show().

MDI is an area of Qt that doesn't receive much love

sykhro (approved):

PR #451, opened 2024-04-14
Add SnappingSplitter for HexView

mikelow:
## Description
This adds a new QSplitter subclass to which you can apply "SnapRanges" - upper and lower bounds which the splitter will snap to, preventing sizing to any value in between. It's basically the same behavior as when a QSplitter nears collapse of one of its views.

The use case is to improve how the HexView responds to window resizing and manual splitter resizing.

Relatedly, I've added logic for HexView to hide the address prefix when it cannot fit on the screen.

## Motivation and Context
Making the HexView sizing less awkward.

## How Has This Been Tested?
Tested on Windows and MacOS.

https://github.com/vgmtrans/vgmtrans/assets/1659535/829430c9-4e37-47a9-be64-949849675f3c

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/widgets/SnappingSplitter.cpp
+++ b/src/ui/qt/widgets/SnappingSplitter.cpp
@@ -1,0 +1,82 @@
+/*
+* VGMTrans (c) 2002-2024
+* Licensed under the zlib license,
+* refer to the included LICENSE.txt file
+ */
+
+#pragma once
+
+#include "SnappingSplitter.h"
+#include <QMouseEvent>
+
+SnappingSplitter::SnappingSplitter(Qt::Orientation orientation, QWidget* parent)
+    : QSplitter(orientation, parent) {
+  connect(this, &SnappingSplitter::splitterMoved, this, &SnappingSplitter::onSplitterMoved);
+}
+
+void SnappingSplitter::onSplitterMoved() {
+  printf("splitterMoved\n");
+  enforceConfigurations();
+  state = saveState();
+}
+
+void SnappingSplitter::resizeEvent(QResizeEvent* event) {
+  QSplitter::resizeEvent(event);
+  restoreState(state);
+  enforceConfigurationsOnResize();
+}
+
+void SnappingSplitter::enforceConfigurations() {
+  for (const CollapseRange& range : collapseRanges) {
+    QList<int> sizes = this->sizes();
+
+    int halfway = range.upperBound + (range.lowerBound - range.upperBound) / 2;
+    if (sizes[range.index] <= range.upperBound && sizes[range.index] > range.lowerBound) {
+      if (sizes[range.index] > halfway) {
+        // Hold the size at upperBound until halfway point is crossed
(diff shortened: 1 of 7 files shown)


src/ui/qt/widgets/SnappingSplitter.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <QMouseEvent>
+#include <QScrollArea>
+
+SnappingSplitter::SnappingSplitter(Qt::Orientation orientation, QWidget* parent)
+    : QSplitter(orientation, parent) {
+  connect(this, &SnappingSplitter::splitterMoved, this, &SnappingSplitter::onSplitterMoved);
+}
+
+void SnappingSplitter::onSplitterMoved() {
+  printf("splitterMoved\n");
```
sykhro, line 19: leftover from debug?

src/ui/qt/widgets/SnappingSplitter.cpp
```diff
@@ -0,0 +1,107 @@
...
+  persistState();
+}
+
+void SnappingSplitter::resizeEvent(QResizeEvent* event) {
+//  QSplitter::resizeEvent(event);
+//  if (bDoRestore) {
+    restoreState(state);
+    bDoRestore = false;
+//  }
+//  printf("sizes[0]: %d\n", sizes()[0]);
```
sykhro, line 31: ditto?

sykhro:
> It's basically the same behavior as when a QSplitter nears collapse of one of its views.

I wonder if it's possible to achieve the same using QSplitter::setSizes() and the collapsible property directly, but I still see usefulness in this.

mikelow:
Huh, sorry about that. I addressed this commented code and made some minor changes, but somehow my change was lost. I'll fix it up.

mikelow:
> I wonder if it's possible to achieve the same using QSplitter::setSizes() and the collapsible property directly, but I still see usefulness in this.

Not sure I entirely follow, but the collapsible flag seems to be only for [fully collapsing a view to size 0](https://doc.qt.io/qt-6/qsplitter.html#setCollapsible). In effect, with a vanilla QSplitter, you get a "SnapRange" where the lower bound is 0 and upper bound the widget's minimumSize.

The idea behind SnappingSplitter was to allow more snappable areas like that.

sykhro:
> > I wonder if it's possible to achieve the same using QSplitter::setSizes() and the collapsible property directly, but I still see usefulness in this.
> 
> Not sure I entirely follow, but the collapsible flag seems to be only for [fully collapsing a view to size 0](https://doc.qt.io/qt-6/qsplitter.html#setCollapsible). In effect, with a vanilla QSplitter, you get a "SnapRange" where the lower bound is 0 and upper bound the widget's minimumSize.
> 
> The idea behind SnappingSplitter was to allow more snappable areas like that.

My bad, I must be misremembering. I'm sure I have seen this functionality somewhere before

sykhro (approved):

PR #455, opened 2024-04-18
Collection list polish

mikelow:
## Description
Just a couple small changes:

- First, it turns out the unwanted name editing behavior in the collection list view is due to setting the `ItemIsEditable` flag, so I removed it.
- Second, on MacOS, the collection list is reserving space for the horizontal scrollbar when it shouldn't. I fixed this. This is probably a Qt bug affecting wrapping QListViews that scroll horizontally, since Qt otherwise behaves in the standard fashion of not reserving space for MacOS's disappearing scrollbars. 

## How Has This Been Tested?
Run on MacOS Sonoma.

Minimum height required to show 6 rows:

Before
image

After
image

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/workarea/VGMCollListView.cpp
+++ b/src/ui/qt/workarea/VGMCollListView.cpp
@@ -71,7 +71,7 @@ Qt::ItemFlags VGMCollListViewModel::flags(const QModelIndex &index) const {
     return Qt::ItemIsEnabled;
   }
 
-  return QAbstractListModel::flags(index) | Qt::ItemIsEditable;
+  return QAbstractListModel::flags(index);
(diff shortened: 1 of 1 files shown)


sykhro (approved):
Not sure how future-proof this fix is, but it looks better like this. I think Qt5 behaves the same

mikelow:
Note that I just updated the code to use setViewportMargin() instead of the padding attribute in a stylesheet, since this seems cleaner.

> Not sure how future-proof this fix is

I've tried, but I don't think there's any reasonable way to determine the area Qt is reserving to draw scrollbars in the QListView.

PR #456, opened 2024-04-20
Add a status bar, simplify tree view items

mikelow:
## ⚠️ _Updated. See Description section_

The commit message sums up most of the changes:
- add a status bar on the main window and its logic to update it in VGMFileListView, VGMFileView, and VGMFileTreeView
- simplify VGMFileTreeView items to just their name for easier navigation/visualization
- improve logic for focus and item selection between VGMFileListView, VGMFileView, VGMFileTreeView, and RawFileListView.
- remove redundant VGMFile `name` property and add some info in VGMFile::GetDescription()
- VGMItem::GetDescription() should now return only supplemental descriptive information about an item, and not the item name as a prefix
- Update the HexView tooltip formatting to be multiline

## Description
This introduces two new classes `StatusBarContent`, and ~~`StatusManager`~~. `StatusBarContent` is the widget we use for statusBar()->setWidget() in MainWindow, so it's basically the status bar. It contains the necessary QLabels and a public `setStatus()` method. ~~`StatusManager` is a singleton used for sending a status bar update. It exposes two public methods, which emit a signal to StatusBarContent, one of which is just a wrapper for setStatus() and the other which takes a `VGMItem` and sets the status with its properties.~~

⚠️ **Update:** I've renamed and repurposed `StatusManager` to a more generic singleton called `NotificationCenter`. In addition to handling the status bar updates, this class now also handles the signal when VGMFile selection changes. All of the views that respond to vgmfile selection change do so by connecting to the `NotificationCenter::vgmFileSelected` signal, and instead of emitting directly when vgmfile selection changes, they instead call `NotificationCenter::selectVGMFile()`.

I've also added RawFileListView as a subscriber to `vgmFileSelected`. When a VGMFile is selected, it selects the parent RawFile as the current item.

## Motivation and Context
Two things:
- I'd like the tree view to fit more information on the screen at once. Making each item one line long and moving the extra information into a status bar achieves that.
- I find it useful to be able to get offset/size information on VGMFiles.

## TODO
- ~~Add integration with the raw file and collection list views.~~
- Lighten the status bar text color in light mode. Played around with QPalette and the like, but had some difficulty getting it to work.

I understand it may be useful to pack more information into each tree item. I'd like to explore adding a toggle button to display more information per tree item, possibly as part of the "File Structure" header, which would need to be enlarged.

Another (not mutually exclusive) idea would be to have a mix between a tree view and a table view, to structure the data of each item more clearly and allow it to take less vertical space. Not sure if Qt provides a view like this.

## How Has This Been Tested?
Run on MacOS and Windows. Tested long name/description text and verified it scales and truncates appropriately.

MacOS
Screenshot 2024-04-19 at 3 25 50 PM
Screenshot 2024-04-19 at 3 29 04 PM

Windows
Capture
CaptureLight

Ubuntu
CaptureLight

## Types of changes

## Checklist:


sykhro:
ACK, will look at all of this later today. Sorry, I'm juggling a lot of things lately (but I still deeply care about this project :) )

mikelow:
> ACK, will look at all of this later today. Sorry, I'm juggling a lot of things lately (but I still deeply care about this project :) )

Thank you, but there's no rush! Take all the time you need.

sykhro:
> I understand it may be useful to pack more information into each tree item. I'd like to explore adding a toggle button to display more information per tree item, possibly as part of the "File Structure" header, which would need to be enlarged.

I find that being able to see details such as length of other items at a glance is very useful when implementing new/debugging parsers (which was the reason I added it). We should also probably drop the bold font since it's not a header anymore.

> Another (not mutually exclusive) idea would be to have a mix between a tree view and a table view, to structure the data of each item more clearly and allow it to take less vertical space. Not sure if Qt provides a view like this.

I can think of two ways to achieve this. If you could add it to this one PR it would be great win-win. No worries if you'd rather leave it for the future.

1. Following [this Qt example](https://doc.qt.io/qt-6/qtwidgets-itemviews-simpletreemodel-example.html). Unsure if this has the best performance.
2. Using a delegate (extending the `QItemDelegate::paint()` method to draw a grid, then `setItemDelegate()` on the `QTreeView` using the new class). This is definitely faster but also more hacky.

sykhro:
As for the status bar, that's okay by me. How would you feel about having it at the top of the window? This way it's to the hex view, I find that it gives less eye ping-pong.

mikelow:
> I find that being able to see details such as length of other items at a glance is very useful when implementing new/debugging parsers (which was the reason I added it). We should also probably drop the bold font since it's not a header anymore.

Can you explain this use-case little more? I'm surprised because I consider offset/size to be the least useful information in the tree, I suppose because this information is conveyed in the HexView.

> 1. Following [this Qt example](https://doc.qt.io/qt-6/qtwidgets-itemviews-simpletreemodel-example.html). Unsure if this has the best performance.

This looks very promising. Are you aware of performance issues, or just speculating?

> 2. Using a delegate (extending the QItemDelegate::paint() method to draw a grid, then setItemDelegate() on the QTreeView using the new class). This is definitely faster but also more hacky.

But wouldn't the other complexity be in implementing the column headers?

mikelow:
> As for the status bar, that's okay by me. How would you feel about having it at the top of the window? This way it's to the hex view, I find that it gives less eye ping-pong.

I get the issue with visual ping-pong, but having the status bar at the top would really go against UI convention. I also don't think it would solve the ping-pong, just marginally improve it.

I think a hybrid tree/table view would solve the issue. I'm envisioning it being like a MIDI event list view in a DAW. Ideally, it wouldn't be just a second column with the description string, but something more structured. That will take some work.

I might just try to immediately implement solution 1 - the toggle switch, since that will be far less work. But I'd like to do this in a separate PR as this one is pretty big as-is.

mikelow:
> We should also probably drop the bold font since it's not a header anymore.

Good point. Done.

PR #458, opened 2024-04-20
Add enter and return key behavior for list views

mikelow:
## Description
This adds keypress handling for the enter and return keys (which it treats the same). In the file list view and the collection view, these keys open the current file just a double click would. In the collection list view, it toggles playback just like the space bar does.

## Motivation and Context
Getting the app to conform to standard UI input behavior.

## How Has This Been Tested?
Tested on Mac.

## Types of changes

## Checklist:


sykhro (approved):
On the topic of keyboard nav: does tab-ing between widgets currently work? Haven't tried in a while.

mikelow:
> On the topic of keyboard nav: does tab-ing between widgets currently work? Haven't tried in a while.

Don't think so. Do you mean between the different UI views? Did this ever work?

PR #462, opened 2024-04-23
Add Settings service, add "Show Details" button to VGMFileTreeView

mikelow:
## Description
Two additions here:

### Settings Service,
First, is the introduction of a `Settings` class, a singleton that accesses QSettings ~~and emits signals when settings have been changed~~ (⚠️ Update: I'm now routing through NotificationCenter as I believe it's more intuitive). I'm attempting to use nested structures to group the settings and make code more readable. For example, the first component of the app to use Settings is the VGMFileTreeView, so I've added a nested struct called VGMFileTreeViewSettings with an instance that is simply the class name: VGMFileTreeView. Thus, the following lines access the showDetails setting:

`Settings::the()->VGMFileTreeView.showDetails()`
`Settings::the()->VGMFileTreeView.setShowDetails()`

~~QObject signals have some limitations. They cannot be static, and they cannot be used within nested classes or structs. Therefore, I had to declare them within the Settings scope. Because of this, I prefixed the name of the signal with the group: `VGMFileTreeView_showDetailsChanged()`.~~ 
⚠️ Update: As mentioned, the signals are now routed through NotificationCenter. They still use prefixes.

Let me know if you have idea on this. I'm sure this isn't perfect.

### Show Details button
Second, as mentioned in the status bar PR, I went ahead and added a checkbox button to the VGMFileTreeView header to enable/disable the more detailed tree view items. I've linked this to the Settings service so that it will persist across app runs and between sub windows, so you don't have to keep checking the box.

## How Has This Been Tested?
I ran this on MacOS, Windows and Ubuntu. Settings successfully persist across app runs, and updates other sub windows when the checkbox is activated/deactivated on one of them.

MacOS
Screenshot 2024-04-23 at 12 42 13 AMScreenshot 2024-04-23 at 12 41 58 AM

Windows
Capture2Capture

Ubuntu
imageimage

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/CMakeLists.txt
+++ b/src/ui/qt/CMakeLists.txt
@@ -64,6 +64,8 @@ add_executable(
   main_ui.cpp
   services/MenuManager.cpp
   services/MenuManager.h
+  services/Settings.cpp
+  services/Settings.h
   services/NotificationCenter.cpp
   services/NotificationCenter.h
   services/commands/SaveCommands.h
--- a/src/ui/qt/main_ui.cpp
+++ b/src/ui/qt/main_ui.cpp
@@ -15,6 +15,7 @@
 
 int main(int argc, char *argv[]) {
   QCoreApplication::setOrganizationName("VGMTrans");
+  QCoreApplication::setOrganizationDomain("vgmtrans.com");
   QCoreApplication::setApplicationName("VGMTrans");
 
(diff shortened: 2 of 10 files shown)


sykhro (approved):
Thanks! This is a nice solution for now

PR #468, opened 2024-04-29
Code cleanup stuff

TheComputerGuy96:
This PR compiles on Linux (with both GCC and Clang); not sure about Windows though :frog:


sykhro:
Thanks!

PR #474, opened 2024-05-07
Respecting compiler warnings, part 1

mikelow:
I'm going down the list of code files roughly alphabetically, addressing as many compiler warnings as I can. This does not include clang-tidy warnings. I am trying to leave the logic as-is to avoid potentially breaking changes, though there are a couple or so places where I had to change logic.

## Motivation and Context
Clean up the the code and make it easier to read compiler logs.

### DIFF
--- a/src/main/Root.cpp
+++ b/src/main/Root.cpp
@@ -57,8 +57,8 @@ bool VGMRoot::OpenRawFile(const std::string &filename) {
 }
 
 /* Creates a new file backed by RAM */
-bool VGMRoot::CreateVirtFile(uint8_t *databuf, uint32_t fileSize, const std::string &filename,
-                             const std::string &parRawFileFullPath, VGMTag tag) {
+bool VGMRoot::CreateVirtFile(const uint8_t *databuf, uint32_t fileSize, const std::string& filename,
+                             const std::string &parRawFileFullPath, const VGMTag& tag) {
     assert(fileSize != 0);
 
     auto newVirtFile = new VirtFile(databuf, fileSize, filename,
@@ -118,7 +118,7 @@ bool VGMRoot::CloseRawFile(RawFile *targFile) {
     return false;
   }
 
-  auto file = std::find(vRawFile.begin(), vRawFile.end(), targFile);
+  auto file = std::ranges::find(vRawFile, targFile);
   if (file != vRawFile.end()) {
     auto &vgmfiles = (*file)->containedVGMFiles();
     UI_BeginRemoveVGMFiles();
@@ -151,12 +151,11 @@ void VGMRoot::RemoveVGMFile(std::variant<VGMSeq *, VGMInstrSet *, VGMSampColl *,
   auto targFile = variantToVGMFile(file);
   // First we should call the format's onClose handler in case it needs to use
   // the RawFile before we close it (FilenameMatcher, for ex)
-  Format *fmt = targFile->GetFormat();
-  if (fmt) {
+  if (Format *fmt = targFile->GetFormat()) {
     fmt->OnCloseFile(file);
   }
 
-  auto iter = std::find(vVGMFile.begin(), vVGMFile.end(), file);
+  auto iter = std::ranges::find(vVGMFile, file);
 
   if (iter != vVGMFile.end()) {
     UI_RemoveVGMFile(targFile);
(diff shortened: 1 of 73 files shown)


src/main/Root.cpp
```diff
@@ -57,8 +57,8 @@ bool VGMRoot::OpenRawFile(const std::string &filename) {
 }
 
 /* Creates a new file backed by RAM */
-bool VGMRoot::CreateVirtFile(uint8_t *databuf, uint32_t fileSize, const std::string &filename,
-                             const std::string &parRawFileFullPath, VGMTag tag) {
+bool VGMRoot::CreateVirtFile(const uint8_t *databuf, uint32_t fileSize, const std::string& filename,
```
sykhro, line 60: jsyk, unrelated to this line in particular, std::string and other flat collections use "small buffer optimization". If we have a ton of small strings (including the empty string), it would make sense to pass them by value instead of by reference.

sykhro (approved):

src/main/Root.cpp
mikelow: Duly noted! I was unaware of small buffer optimization, but I was reading up about the optimization trade-offs between passing strings by value and using std::move() for member initialization vs passing by reference. From what I gleaned, it sounds like pass by value/move is best for rvalues, and pass by reference is better for lvalues.

Do you have any opinion on this?

sykhro: > From what I gleaned, it sounds like pass by value/move is best for rvalues, and pass by reference is better for lvalues.

Essentially, yes. However, value categories got complicated in recent years so sometimes it's not as simple (see [here](https://en.cppreference.com/w/cpp/language/value_category)).

In my opinion, it really depends on the idioms of the codebase. We come from older c++ with raw pointers, the next best things is switching to unique_ptr/shared_ptr where we can to avoid memory corruption bugs. And with unique_ptr, we should also be able to clean up the hierarchy of classes we have.

Modern c++ gives the opportunity to work with value semantics. Converting the codebase, however, is probably more trouble than it's worth. At least for the core stuff.
I would say to keep things as they are for now, these strings are definitely long (filesystem paths).

mikelow: Got it.

> I would say to keep things as they are for now, these strings are definitely long (filesystem paths).

Am I right in my understanding that we should try to use value string params for functions that mostly receive rvalues (like LogManager::log()) and use references when the parameter is usually passed as an lvalue (the code snippet above, I assume)?

Or am I overthinking this and we should just aim for pass by reference across the board?

sykhro: > Am I right in my understanding that we should try to use value string params for functions that mostly receive rvalues (like LogManager::log()) and use references when the parameter is usually passed as an lvalue (the code snippet above, I assume)?

Yes, I would say all "trivial" cases like that are good for `std::move`. The dangers come with mixing value and reference semantics (that is, having some functions return pointers and some values) and with potentially accessing something after you have `move`'d it. So this is mostly around function parameters for utility things.

With regards to all the rest, a good goal given the idioms of our project would be:
- `const` correctness
- smart pointers where ownership is involved
    - one should never call new/delete directly but use `std::make_unique` and `std:::make_shared`
    - raw pointers are still cool for passing objects around when something needs to access them without owning them
    - std::unique_ptr is easier to reason about than std::shared_ptr, so we should strive for that

mikelow: > The dangers come with mixing value and reference semantics (that is, having some functions work with pointers and some work with values) and with potentially accessing something after you have move'd it. 

But do you think it's reasonable that we allow for both pass by reference and pass by value so long as we exclusively use pass by value / std::move() params for simple constructors that move() the value in an initializer list? There seems to be significant [performance gains](https://lemire.me/blog/2023/01/30/move-or-copy-your-strings-possible-performance-impacts) from using std::move with rvalues, and unless I misunderstand, we pass around a lot of rvalues with VGMItem names, for example.

sykhro: > > The dangers come with mixing value and reference semantics (that is, having some functions work with pointers and some work with values) and with potentially accessing something after you have move'd it.
> 
> But do you think it's reasonable that we allow for both pass by reference and pass by value so long as we exclusively use pass by value / std::move() params for simple constructors that move() the value in an initializer list? There seems to be significant [performance gains](https://lemire.me/blog/2023/01/30/move-or-copy-your-strings-possible-performance-impacts) from using std::move with rvalues, and unless I misunderstand, we pass around a lot of rvalues with VGMItem names, for example.

Yeah, that's what I meant. Small-ish instances where it's obvious why something gets moved

PR #475, opened 2024-05-09
Clang warning fixes

TheComputerGuy96:
This should be enough to finally enable `-Werror` in non-MSVC cases (MSVC uses a `/WX` flag for that instead) :frog: 

For the zlib issue I could've just fixed the problematic prototype part but updating zlib probably brings some security fixes too so I think that's a better option


src/main/formats/CPS/CPS2Scanner.cpp
```diff
@@ -104,6 +104,8 @@ void CPS2Scanner::Scan(RawFile *file, void *info) {
       if (fmt_ver >= VER_130 && !seqRomGroupEntry->GetHexAttribute("artic_table", &artic_table_offset))
         return;
       break;
+    default:
+      break;
   }
```
mikelow, line 109: Instead of default: break; here, we should handle the missing enum cases. In the event of VER_UNDEFINED, we should log an error and return;

src/main/formats/CPS/CPS1Scanner.cpp
```diff
@@ -103,6 +103,8 @@ void CPS1Scanner::LoadCPS1(MAMEGame *gameentry, CPSFormatVer fmt_ver) {
     case VER_CPS1_502:
       seq_table_length = (uint32_t) (programFile->GetShortBE(seq_table_offset) * 2) + ptrsStart;
       break;
+    default:
+      break;
   }
```
mikelow, line 108: This is an error state - the version is non-CPS1 or undefined. We should log an error and return.

src/main/formats/MoriSnes/MoriSnesSeq.cpp
```diff
@@ -732,7 +732,7 @@ void MoriSnesTrack::ParseInstrumentEvents(uint16_t offset, uint8_t instrNum, boo
 
   while (bContinue) {
     uint16_t beginOffset = curOffset;
-    if (curOffset >= 0x10000) {
```
mikelow, line 735: Why this change?

src/main/formats/MoriSnes/MoriSnesSeq.cpp
```diff
@@ -762,7 +762,7 @@ void MoriSnesTrack::ParseInstrumentEvents(uint16_t offset, uint8_t instrNum, boo
       }
 
       beginOffset = curOffset;
-      if (curOffset >= 0x10000) {
```
mikelow, line 765: Ditto, why are you changing the value?

mikelow:
Did you do any testing? You made a number of logic changes in format code.

src/main/formats/FFT/FFTSeq.cpp
```diff
@@ -188,9 +188,11 @@ bool FFTTrack::ReadEvent(void) {
         //Octave Event
 
         //Set octave
-      case 0x94:
-        AddSetOctave(curOffset - 2, 2, GetByte(curOffset++));
+      case 0x94: {
+        uint8_t newOctave = GetByte(curOffset++);
+        AddSetOctave(curOffset - 2, 2, newOctave);
```
mikelow, line 193: This should be:
`AddSetOctave(beginOffset, 2, newOctave);`

Using curOffset was probably a dumb mistake of mine.

src/main/formats/FFT/FFTSeq.cpp
```diff
@@ -477,9 +479,11 @@ bool FFTTrack::ReadEvent(void) {
         //Volume Event
 
         // Volume
-      case 0xE0:
-        AddVol(curOffset - 2, 2, GetByte(curOffset++));
+      case 0xE0: {
+        uint8_t volByte = GetByte(curOffset++);
+        AddVol(curOffset - 2, 2, volByte);
```
mikelow, line 484: Likewise, should be:
AddVol(beginOffset, curOffset - beginOffset, volByte);

src/main/formats/FFT/FFTSeq.cpp
```diff
@@ -528,9 +532,11 @@ bool FFTTrack::ReadEvent(void) {
         //Panpot Event
 
         // Panpot
-      case 0xE8:
-        AddPan(curOffset - 2, 2, GetByte(curOffset++));
+      case 0xE8: {
+        uint8_t panpotByte = GetByte(curOffset++);
+        AddPan(curOffset - 2, 2, panpotByte);
```
mikelow, line 537: AddPan(beginOffset, curOffset - beginOffset, panpotByte);

src/main/formats/KonamiSnes/KonamiSnesSeq.cpp
```diff
@@ -234,6 +234,9 @@ void KonamiSnesSeq::LoadEventMap() {
       EventMap[0xfb] = EVENT_ADSR2;
       EventMap[0xfc] = EVENT_PROGCHANGEVOL;
       break;
+
+    default:
+      break;
   }
```
mikelow, line 240: I think we should put a warning about the format version being unknown:
`L_WARN("Unknown version of Konami SNES format.");`

mikelow:
@TheComputerGuy96 from where did you get this version of zlib?
@sykhro should we make zlib a git submodule given that we're doing that for spdlog?

sykhro:
> @TheComputerGuy96 from where did you get this version of zlib? @sykhro should we make zlib a git submodule given that we're doing that for spdlog?

Yes, at that point we could even switch to https://github.com/zlib-ng/zlib-ng/

mikelow:
@TheComputerGuy96 I'm happy to make a new branch that brings in your commits sans the vendored zlib, fixes the issues I mentioned, and introduces zlib-ng via git submodule.

TheComputerGuy96:
> from where did you get this version of zlib?

I unpacked the zlib 1.3.1 tarball (specifically https://github.com/madler/zlib/releases/download/v1.3.1/zlib-1.3.1.tar.gz) and started removing files from it to reduce the space of the module and modified CMakeLists,txt to remove the shared library stuff and build the minizip library together too

> At that point we could even switch to https://github.com/zlib-ng/zlib-ng/

I'm fine with this because Arch has packaged both zlib-ng and minizip-ng (so a submodule won't be needed there)

src/main/formats/MoriSnes/MoriSnesSeq.cpp
TheComputerGuy96: Because Clang was complaining about that (integer overflow?) check being always false

TheComputerGuy96:
> Did you do any testing? You made a number of logic changes in format code.

I would have to get music of multiple different formats to test this PR properly

mikelow:
Closing this now that it has been integrated into #477 and merged. 

I am working on replacing zlib/minizip with zlib-ng/minizip-ng and including them as git submodules. I will put up that PR when it is ready.

PR #476, opened 2024-05-10
Warning cleanup 2

mikelow:
More code cleanup, mostly to get rid of warnings.

Changes in SynthFile.h are actually minor - i moved one of the classes because it is used as a member for classes below it and removed an unused constructor or two Only logical change was adding default initial values for some members in constructors.

A lot of changes address "weak warnings." You'll see I've made a number of methods const or static. Let me know if you think any of this is overkill. 

## How Has This Been Tested?
Basic testing. Files are loading and playing as expected.

## Types of changes

- [x] Code cleanup

## Checklist:

### DIFF
--- a/src/main/components/VGMColl.cpp
+++ b/src/main/components/VGMColl.cpp
@@ -545,6 +545,8 @@ SynthFile *VGMColl::CreateSynthFile() {
         else
           attenuation = 0;
 
+        sampInfo->SetPitchInfo(realUnityKey, realFineTune, attenuation);
+
         double sustainLevAttenDb;
         if (rgn->sustain_level == -1)
           sustainLevAttenDb = 0.0;
@@ -556,8 +558,6 @@ SynthFile *VGMColl::CreateSynthFile() {
         newArt->AddADSR(rgn->attack_time, static_cast<Transform>(rgn->attack_transform),
           rgn->hold_time, rgn->decay_time, sustainLevAttenDb, rgn->sustain_time, rgn->release_time,
           static_cast<Transform>(rgn->release_transform));
-
-        sampInfo->SetPitchInfo(realUnityKey, realFineTune, attenuation);
       }
     }
   }
--- a/src/main/conversion/SynthFile.h
+++ b/src/main/conversion/SynthFile.h
@@ -81,11 +81,9 @@ public:
 
 class SynthRgn {
  public:
-  SynthRgn() : usKeyLow(0), usKeyHigh(0x7F), usVelLow(0), usVelHigh(0x7F),
-        fusOptions(0), usPhaseGroup(0), channel(1), tableIndex(0), sampinfo(nullptr), art(nullptr) { }
+  SynthRgn() = default;
   SynthRgn(uint16_t keyLow, uint16_t keyHigh, uint16_t velLow, uint16_t velHigh)
-      : usKeyLow(keyLow), usKeyHigh(keyHigh), usVelLow(velLow), usVelHigh(velHigh),
-        fusOptions(0), usPhaseGroup(0), channel(1), tableIndex(0), sampinfo(nullptr), art(nullptr) { }
+      : usKeyLow(keyLow), usKeyHigh(keyHigh), usVelLow(velLow), usVelHigh(velHigh) {}
   ~SynthRgn();
 
   SynthArt *AddArt();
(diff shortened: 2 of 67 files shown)


src/main/conversion/SynthFile.h
```diff
@@ -58,30 +51,56 @@ class SynthInstr {
...
 class SynthRgn {
  public:
-  SynthRgn() : sampinfo(nullptr), art(nullptr) { }
+  SynthRgn() : usKeyLow(0), usKeyHigh(0x7F), usVelLow(0), usVelHigh(0x7F),
+        fusOptions(0), usPhaseGroup(0), channel(1), tableIndex(0), sampinfo(nullptr), art(nullptr) { }
   SynthRgn(uint16_t keyLow, uint16_t keyHigh, uint16_t velLow, uint16_t velHigh)
-      : usKeyLow(keyLow), usKeyHigh(keyHigh), usVelLow(velLow), usVelHigh(velHigh), sampinfo(nullptr), art(nullptr) { }
-  SynthRgn(uint16_t keyLow, uint16_t keyHigh, uint16_t velLow, uint16_t velHigh, SynthArt &art);
+      : usKeyLow(keyLow), usKeyHigh(keyHigh), usVelLow(velLow), usVelHigh(velHigh),
+        fusOptions(0), usPhaseGroup(0), channel(1), tableIndex(0), sampinfo(nullptr), art(nullptr) { }
```
sykhro, line 88: Since we're here, let's reduce complexity: it's a good idea to move the default values to member initializers. Rationale [here](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c48-prefer-default-member-initializers-to-member-initializers-in-constructors-for-constant-initializers)

src/main/components/VGMColl.cpp
```diff
@@ -545,6 +545,8 @@ SynthFile *VGMColl::CreateSynthFile() {
         else
           attenuation = 0;
 
+        sampInfo->SetPitchInfo(realUnityKey, realFineTune, attenuation);
```
sykhro, line 548: Sneaky bugfix? :)

src/ui/qt/services/MenuManager.h
```diff
@@ -6,13 +6,15 @@
 
 #pragma once
 
+#include <sys/syslog.h>
```
sykhro, line 9: Will break on windows (why is this even needed?)

src/ui/qt/services/MenuManager.h
```diff
@@ -287,6 +289,9 @@ class MenuManager {
             case PropertySpecValueType::ItemList:
               propMap.insert({ propSpec.key, items });
               break;
+            default:
+              L_ERROR("Unknown PropertySpecValueType value: {}", static_cast<int>(propSpec.valueType));
```
sykhro, line 293: just fyi, spdlog uses libfmt/std::format. with both of those, we can define custom formatters for our objects, so in cases like this you could have a nicer message than just a number. it goes really well when paired with [magic_enum](https://github.com/Neargye/magic_enum)

sykhro (changes requested):
`const`-correctness is never overkill, this was much needed :)

src/main/components/VGMColl.cpp
mikelow: Actually, no. This can just be called earlier in the function. Was part of trying to change the interface of AddSampInfo and other AddX functions in SynthFile to accept a reference or value of the type, rather than the weird behavior of instantiating an instance and returning it. 

Decided such a refactor would be better suited for a different PR.

src/main/conversion/SynthFile.h
mikelow: Much cleaner. Will change

src/ui/qt/services/MenuManager.h
mikelow: Ok, something I'll have to research. Do you think we should add magic_enum as a dependency? I actually bookmarked that project while researching stuff for LogManager (wanted to log the name of the calling class - no great solutions).

sykhro: The closest thing to that in standard c++20 is `std::source_location` + `std::filesystem::path` to extract the filename of the calling code.
magic_enum might be useful, can't think of a specific usecase in this project though atm

mikelow: Ah, I see. It sounds like this might present a better alternative to my current compromise solution, which is to use __FILE_NAME__ in compilers that support it. One of my goals was to exclude the full path from the compiled code.

I see now how one can define a fmt::formatter<SomeClass> struct to create a custom formatter. Seemsl ike overkill for PropertySpecValueType, though.

mikelow:
I believe the issues have been addressed. Let me know if this is good to go.

sykhro (approved):
LGTM! Sorry, didn't notice the push

PR #477, opened 2024-05-12
Clang warnings fixup

mikelow:
@TheComputerGuy96, and FYI @sykhro 
This pulls in the commits from from #475 but excludes the vendored zlib update and enabling of -Werror. Instead, we will switch to importing zlib-ng as a git submodule in a future PR, at which point we can probably also enable -Werror. Note that I also added a commit that addresses some of the issues in MoriSnesSeq.cpp.

Given that I already reviewed #475, I will leave this up for a couple days for comment and then merge it if there is no objection or discussion.

## Motivation and Context
In order to keep our git history clean, I've moved the #475 commits into a new branch to avoid the unnecessary vendored update to zlib.

## How Has This Been Tested?
Tested that MoriSnes and FFT drivers are still working as expected.

## Types of changes

- [x] Code cleanup

## Checklist:

### DIFF
--- a/src/main/formats/MoriSnes/MoriSnesSeq.cpp
+++ b/src/main/formats/MoriSnes/MoriSnesSeq.cpp
@@ -13,8 +13,8 @@ DECLARE_FORMAT(MoriSnes);
 #define MAX_TRACKS  10
 #define SEQ_PPQN    48
 
-MoriSnesSeq::MoriSnesSeq(RawFile *file, MoriSnesVersion ver, uint32_t seqdataOffset, std::string newName)
-    : VGMSeq(MoriSnesFormat::name, file, seqdataOffset, 0, newName),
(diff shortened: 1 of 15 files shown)


src/main/formats/MoriSnes/MoriSnesSeq.cpp
```diff
@@ -26,10 +26,10 @@ MoriSnesSeq::MoriSnesSeq(RawFile *file, MoriSnesVersion ver, uint32_t seqdataOff
   LoadEventMap();
 }
 
-MoriSnesSeq::~MoriSnesSeq(void) {
+MoriSnesSeq::~MoriSnesSeq() {
```
TheComputerGuy96, line 29: The `void` removals seem unnecessary to me (and aren't a good idea for C compatibility)

sykhro: `void` as a parameter is how you express "this function takes no parameters" in C. It's reduntant in C++ and we make no effort to be C-compatible

PR #479, opened 2024-05-14
Switch to zlib-ng/minizip-ng (and use -Werror by default)

TheComputerGuy96:
This is my (bad) attempt to add zlib-ng/minizip-ng support to VGMTrans (hopefully it won't cause any runtime regressions)
which also enables `-Werror` in non-MSVC cases

This does build on Arch Linux with Clang (so macOS should work too); I didn't test Windows yet (so that's likely to blow up)


mikelow:
I have submitted my attempt to add zlib-ng and minizip-ng here: #480.

Thanks for your work on this. Here's my feedback:

I ran into a permission error when running your CMake script at the point it attempts to create a symbolic link. I'm confused about the purpose of the symbolic link. Perhaps minizip-ng's `FindZLIBNG.cmake` file does not detect `zlibstatic-ng.lib` but does detect `z-ng.lib`?

I believe the script will have issues when building a Debug config on Windows (filename "zlibstatic-ngd.lib") and also when using the Visual Studio toolchain, which puts the resulting zlib lib files in Debug and Release subdirectories.

I am confused about the inclusion of add_library() calls. I am wondering if this is influencing the order zlib-ng and minizip-ng are built, because I am otherwise confused how you guarantee zlib-ng is built before minizip-ng.

I believe there is a misunderstanding that this is building successfully with the -Werror flag. You are applying to Debug builds but not Release builds. Github Actions is building Release versions, so it is not surfacing warnings as errors.

mikelow:
Now that #480 is merged, I will close this PR. Thanks again, and sorry that this was redundant with the work I had already started on this.

TheComputerGuy96:
@mikelow

> I ran into a permission error when running your CMake script at the point it attempts to create a symbolic link. I'm confused about the purpose of the symbolic link.

This is because of how minizip-ng detects the zlib-ng library: https://github.com/zlib-ng/minizip-ng/blob/9184ba1216589077803b3e4e85f2b226f861ab2d/cmake/FindZLIBNG.cmake#L16 (and Windows having a different zlib-ng static library name so I had to do that hack to avoid modifying the submodules)

> I believe the script will have issues when building a Debug config on Windows (filename "zlibstatic-ngd.lib") and also when using the Visual Studio toolchain, which puts the resulting zlib lib files in Debug and Release subdirectories.

GitHub Actions uses MSVC for Windows right now (which is a part of VS) and it builds fine (Debug builds might be problematic though); I can't properly test Windows builds right now because of no VS in my Wine prefix

> I am confused about the inclusion of add_library() calls.

This is for matching library names with the system package builds of minizip-ng/zlib-ng (otherwise you would have 2 different code paths in target_link_libraries)

> You are applying to Debug builds but not Release builds.

I see the entire warnings section for non-MSVC only applies to non-Release builds (I should try removing the check and see what happens)

mikelow:
> This is because of how minizip-ng detects the zlib-ng library: https://github.com/zlib-ng/minizip-ng/blob/9184ba1216589077803b3e4e85f2b226f861ab2d/cmake/FindZLIBNG.cmake#L16 (and Windows having a different zlib-ng static library name so I had to do that hack to avoid modifying the submodules)

The other way to do this, which I employed in #480, is to to directly set the vars that FindZLIBNG is handling: `ZLIBNG_FOUND`, `ZLIBNG_INCLUDE_DIRS`, and `ZLIBNG_LIBRARY`.

> GitHub Actions uses MSVC for Windows right now (which is a part of VS) and it builds fine (Debug builds might be problematic though); I can't properly test Windows builds right now because of no VS in my Wine prefix

Perhaps the file path difference comes from using nmake instead of ninja. I found that if I built the project in a basic way using `cmake -B build` on my windows setup, the lib output paths were different, being placed in Debug and Release subdirectories. 

There's probably an option to force the output directory, but I realize now we don't have to predict or direct where the libs will be built or where FindZLIBNG will search for them if we use sykhro's suggestion: `set(ZLIBNG_LIBRARY $<TARGET_FILE:zlibstatic>)`.

> I see the entire warnings section for non-MSVC only applies to non-Release builds (I should try removing the check and see what happens)

Lots and lots of errors, unfortunately.

PR #480, opened 2024-05-20
Update to zlib-ng and minizip-ng using git submodules

mikelow:
This replaces the vendored versions of zlib and minizip with [zlib-ng](https://github.com/zlib-ng/zlib-ng) and [minizip-ng](https://github.com/zlib-ng/minizip-ng) via git submodules.

I ran into some difficulty getting minizip-ng to build with a sibling zlib-ng dependency. Minizip-ng's CMake script uses a find_package() script to find an existing zlib-ng lib at configuration time. However, since the zlib-ng dependency will not have been built yet, this poses a problem.

My solution is to set all of the vars that would result from a successful find_package() call were zlib-ng already built - effectively, trick the minizip-ng CMake script. However, the other complication is that zlib-ng must be built before minizip-ng. To ensure this, I add zlib-ng within the root CMake script, and add minizip-ng in a child CMake script (src/main/CMakeLists.txt).

If there is a cleaner way to do all of this, I would love to be corrected.

## Motivation and Context
[This PR](https://github.com/vgmtrans/vgmtrans/pull/475) updated the vendored zlib version in an effort to remove warnings, but we decided a better approach would be to migrate to zlib-ng and include it as a git submodule, hence my work here. [This PR](https://github.com/vgmtrans/vgmtrans/pull/479) also attempts to add zlib-ng and minizip-ng, however, I encountered a number of problems with that branch, which I will discuss there shortly.

## How Has This Been Tested?
I tested building the app on MacOS, Windows 10, and Ubuntu 22.04. I verified that minizip-ng is utilizing zlib-ng in the CMake configuration step. I also made sure that the CMake script remains compatible with the Visual Studio toolchain.

## Types of changes

- [x] Clean up / Update dependencies

## Checklist:


sykhro:
> If there is a cleaner way to do all of this, I would love to be corrected.

Your approach is close.
We need `USE_VENDORED_ZLIBNG` CMake options for friendliness with Linux distros.
At that point, after `add_subdirectory(minizipng)`:

```cmake
if (USE_VENDORED_ZLIBNG)
  add_dependencies(minizipng zlibstatic)
endif()
```

also, for paths to the compiled libraries, prefer `$<TARGET_FILE:tgt>` [docs](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_FILE) to grab the path directly from what cmake will generate. So you'd have `$<TARGET_FILE:zlibstatic>` , which also works in case of IMPORTED targets (when not using the vendored option)

sykhro (changes requested):
See comment in PR page

mikelow:
Excellent, thank you for the feedback. I've updated the CMake scripts, hopefully in line with your suggestions. I've moved the zlib-ng stuff out of the root CMake script and into /src/main/CMakeLists.txt file, preceding minizip-ng. I assume you believe it is preferable to include both libraries here instead of in the root CMakeLists.txt file.

Another change I made is to move the lib output directory path from:
${CMAKE_CURRENT_BINARY_DIR}/somelib

to:
${CMAKE_BINARY_DIR}/lib/somelib

I am curious if you have any thoughts on or if you are familiar with [this so-called super project pattern](https://twdev.blog/2024/02/cmake_external_project/) for CMake, which utilizes ExternProject_Add to simplify dependency management.

sykhro (approved):

sykhro:
> I am curious if you have any thoughts on or if you are familiar with [this so-called super project pattern](https://twdev.blog/2024/02/cmake_external_project/) for CMake, which utilizes ExternProject_Add to simplify dependency management.

Never had good experiences with `ExternalProject_Add`: there aren't great CMake resources around, many projects have subpar trees. Also the cloning being part of the build process can lead to unexpected pains (CI lacking permissions, more chance for random firewalls rules/AV to interfere, ...).

I remember a project where failures in`ExternalProject_Add` resulted in the whole git tree being reset. Wasn't fun losing changes like that :)

If this is about third-party dependencies, the comfiest thing (excluding vendoring) is probably `vcpkg` or `conan`. If this is about managing codependent projects, dunno. I've been trying `git-subrepo` with mixed success

PR #481, opened 2024-05-20
Linux install fix

mikelow:
This is an attempt to improve the Linux installation. Linux convention isn't my forte, so hopefully I didn't mess this up.

The changes:
* install mame_roms.xml into `${CMAKE_INSTALL_PREFIX}/share/vgmtrans` (ie. /usr/local/share/vgmtrans).
* install libbass.so and libbassmidi.so into `${CMAKE_INSTALL_PREFIX}/lib` (ie. /usr/local/lib)
* move responsibility for defining the resource path to the Qt CMake script. It now passes a RESOURCE_DIR string macro into the project and QtVGMRoot::UI_GetResourceDirPath() simply returns it.

## How Has This Been Tested?
Loaded mame_roms.xml successfully on Ubuntu 22.04, Windows 10, and MacOS Sonoma. libbass loaded successfully on Ubuntu.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/CMakeLists.txt
+++ b/src/ui/qt/CMakeLists.txt
@@ -30,6 +30,7 @@ find_package(Qt6 REQUIRED COMPONENTS Core Gui Widgets Svg)
 message(STATUS "Using Qt ${Qt6_VERSION}")
 
 if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+  set(BUNDLE_PATH "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/VGMTrans.app")
   set(RESOURCE_DIR "${BUNDLE_PATH}/Contents/Resources")
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
   set(RESOURCE_DIR "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}")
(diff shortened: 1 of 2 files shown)


sykhro (commented):
I've been out of the Linux game for a while so I'm not sure about current practices in packaging. third-party deps like BASS when not packaged by the distro might need to live in `/opt/vgmtrans/lib`. no clue how widely accepted this is however (my gentoo days are long gone)

mikelow:
Ok. I thought that might be the right place for path for the libs, but wasn't sure. By "keep it as-is" do you mean ditch the PR or move forward with it? I'm cool with either.

TheComputerGuy96:
@mikelow I think the BASS libraries should be in a separate vgmtrans subdirectory inside of /usr/lib (to prevent conflicts with any version of system BASS)

This requires a rpath to be set though 🐸

sykhro:
>  By "keep it as-is" do you mean ditch the PR or move forward with it? I'm cool with either.

I did some research and it seems like the place for this stuff is indeed `/opt`, we should avoid putting our deps into system lib directories imo

mikelow:
I didn't understand how our linux packaging worked when I started on this, and now I'm no longer convinced that fixing the install command is adding much value. I'll close the PR.

PR #483, opened 2024-05-22
Warning cleanup 4

mikelow:
This progresses through more format code.
I also added a describeUnknownEvent() method and fixed a couple values in .clang-format.

In a couple files, I commented out code that was inducing warnings rather than delete it, since the code could be relevant to future work on the format.

## Types of changes

- [x] Code cleanup

## Checklist:

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -28,11 +28,11 @@ MaxEmptyLinesToKeep: 1
 NamespaceIndentation: None
 PointerAlignment: Left
 ReflowComments:  true
-SortIncludes:    "Never"
+SortIncludes:    false
 SpaceBeforeParens: ControlStatements
 SpaceInEmptyParentheses: false
 SpacesBeforeTrailingComments: 2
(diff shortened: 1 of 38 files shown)


sykhro (approved):
Thank you again for the effort. Note that clang-format output and config format isn't stable between different releases. What version are you using? It should be documented

mikelow:
Hm, I wasn't aware that clang-format is unstable. I'm using the version bundled with CLion, which I would guess is the latest. I can just revert the changes to .clang-format. Hopefully newer versions retain some degree of backward compatability.

sykhro:
> Hm, I wasn't aware that clang-format is unstable. I'm using the version bundled with CLion, which I would guess is the latest (they don't make it clear). I can just revert the changes to .clang-format. Hopefully newer versions retain some degree of backward compatability.

not always :c clang-format also uses very little of clang internals and none of the AST, which is why it can get confused as to what the statements actually are

PR #485, opened 2024-05-23
Warning cleanup part 6

mikelow:
This one is a bit more random as I noticed my IDE wasn't surfacing many of clang's compile-time warnings in the editor.

A couple more meaningful changes:
- removed `key` and `vel` members from SeqTrack. Made many others protected instead of public.
- replaced some constant #defines in format files with static constexprs (example: kMaxTracks). 

These warnings are surfacing a couple awkward things:
1. -Wshadow warnings remind me that there's no clear variable naming convention. It's annoying that we can't have a `name` member and also a constructor with a `name` parameter without eliciting a -Wshadow warning. So do we adopt a m_ prefix for member variables? That would mean a ton of changes.
2. There are a lot of warnings concerning type narrowing, particularly around the use of size_t. It strikes me as overkill to move all VGMItem offset and length members to size_t or uint64_t, but I'm not sure what the alternative is other than keeping uint32_t and static_casting size_t to it (or changing size_t usage to uint32_t in the first place). How much do we care about supporting 4gb+ files?

## Types of changes

- [x] Code cleanup

## Checklist:


sykhro:
> This one is a bit more random as I noticed my IDE wasn't surfacing many of clang's compile-time warnings in the editor.

This makes sense. I'm preparing a PR that solves this problem and improves our handling of compiler settings

sykhro:
Err, I don't know where the rest of my comment went.

> -Wshadow warnings remind me that there's no clear variable naming convention. It's annoying that we can't have a name member and also a constructor with a name parameter without eliciting a -Wshadow warning. So do we adopt a m_ prefix for member variables? That would mean a ton of changes.

This is a job for an automated refactor using libtooling/libastmatcher. I can give it a shot if you'd like

> There are a lot of warnings concerning type narrowing, particularly around the use of size_t. It strikes me as overkill to move all VGMItem offset and length members to size_t or uint64_t, but I'm not sure what the alternative is other than keeping uint32_t and static_casting size_t to it (or changing size_t usage to uint32_t in the first place). How much do we care about supporting 4gb+ files?

I find the casts to be annoying to read, I wouldn't be opposed to migrating everything to `size_t` for lenths and `std::ptrdiff_t` for offsets

sykhro (approved):
That being said, I'm all for removing warnings

sykhro:
> That sounds great, but we should discuss style guidelines first. I'm fine with radical change and moving to a fairly standard C++ convention, snake_case and all.

I suggest having a look at [Serenity OS's](https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md)

> VGMFiles, on the other hand [...]

True, I was mostly thinking in terms of RawFile. Let's go with unit32_t then, when  I made that proposal I was misremembering `dwOffset` as signed. Perhaps it might also be useful to have a strong-type alias to that to distinguish between relative/absolute offsets

mikelow:
> I suggest having a look at [Serenity OS's](https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md)

I just read through it and I _really_ like it. Two things I'm not sure I'm sold on:
- East const over West const.
- including both `virtual` and `override` (or `final`) keywords.

I don't feel super strongly about either. So I'm fine to roll with it if it's your preference.

sykhro:
> > I suggest having a look at [Serenity OS's](https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md)
> 
> I just read through it and I _really_ like it. Two things I'm not sure I'm sold on:
> 
> * East const over West const.
> * including both `virtual` and `override` (or `final`) keywords.
> 
> I don't feel super strongly about either. So I'm fine to roll with it if it's your preference.

Right, as much as I get East const I find it confuses potential contributors.
About the second one, I guess that it's done so that it's easier to distinguish virtual functions introduced at some point of the inheritance hierarchy easier? Think C inherits B inherits A and defining a virtual function in B in addition to those in A

mikelow:
Actually, there's one major thing I don't love about the Serenity OS style guide: function names are snake_case.

I think if I had to choose, I would go for lowerCamelCase a la the [Swift guidelines](https://google.github.io/swift/), but I expect this is very non-standard for C++. Otherwise I would prefer UpperCamelCase, which seems to be more common than snake_case.

I think either of the camel cases make code more readable, as they create a visual to snake_cased variables and members.

What are your thoughts?

mikelow:
Oh, more significantly, Qt uses lowerCamelCase. There's that precedent, at least.

mikelow:
> About the second one, I guess that it's done so that it's easier to distinguish virtual functions introduced at some point of the inheritance hierarchy easier? Think C inherits B inherits A and defining a virtual function in B in addition to those in A

I would think the opposite. I'd expect that excluding `virtual` except on its first declaration (meaning no override) would make this distinction clearer.

Not an issue I feel too strongly on, though.

sykhro:
> I think if I had to choose, I would go for lowerCamelCase a la the [Swift guidelines](https://google.github.io/swift/), but I expect this is very non-standard for C++. Otherwise I would prefer UpperCamelCase, which seems to be more common than snake_case.

I think either of the camel cases make code more readable, as they create a visual contrast to snake_cased variables and members.

This is actually pretty standard in C++. A common mix I see is:
- type names in CamelCase
- function names in lowerCamelCase
- variable names in snake_case

in the end, the STL is snake_case + CamelCase; a ton of libraries (like Qt) are CamelCase

mikelow:
That combination sounds ideal to me. How do you feel about it?

sykhro:
Sounds good to me, let's go with that.

PR #487, opened 2024-05-24
Update .clang-format, update VGMRoot fields

mikelow:
Consider this dipping a toe in the water for code reformatting.

There aren't actually that many changes here:

- Updated .clang-format
    - Removed `BasedOnStyle: Google`
    - Added: `BreakBeforeBraces: WebKit` and `ReferenceAlignment: Left`
- Renamed VGMRoot's vRawFile, vVGMFile, vVGMColl to m_rawfiles, m_vgmfiles, m_vgmcolls
- Made the above fields private and created getter functions: rawFiles(), vgmFiles(), vgmColls()
- Renamed pRoot to s_root
- Applied formatting to Root.h and Root.cpp to test the waters.

UPDATE:
I also needed to update CLIVGMRoot slightly. It was clearing the root vgmfile list, which is no longer possible as the getter returns a const and the field itself is private (but also, it was kind of a hack approach in the first place).

## Description

The `BreakBeforeBraces: WebKit` rule just adds a new line before the opening brace of a function, ie:
```
void myFunc()
{
  ...
}
```

instead of:
```
void myFunc() {
  ...
}
```

I went with this since it seems very common, but I don't feel strongly about it. The clang format style option doc is [here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).

I was tempted to use `BasedOnStyle: WebKit', but it resulted in some minor formatting issues that I didn't know how to override, like empty braces being separated with a space (`{ }` instead of `{}`), and putting the `= 0` of a multiline abstract method declaration on its own line. 🤷‍♂️ I'm concerned that we might lose good baseline rules by not using a BasedOnStyle definition.

## Types of changes

- [x] Code cleanup

## Checklist:

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -14,7 +14,7 @@ AlwaysBreakAfterDefinitionReturnType: None
 AlwaysBreakAfterReturnType: None
 AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: "Yes"
-BasedOnStyle: Google
+BreakBeforeBraces: WebKit
 ColumnLimit:     100
 ConstructorInitializerAllOnOneLineOrOnePerLine: false
 ConstructorInitializerIndentWidth: 4
@@ -27,6 +27,7 @@ Language:        Cpp
 MaxEmptyLinesToKeep: 1
 NamespaceIndentation: None
 PointerAlignment: Left
+ReferenceAlignment: Left
 ReflowComments:  true
 SortIncludes:    false
 SpaceBeforeParens: ControlStatements
--- a/src/main/LogManager.h
+++ b/src/main/LogManager.h
@@ -21,7 +21,7 @@ protected:
   void sink_it_(const spdlog::details::log_msg& msg) override {
     auto level = convertSpdlogLevel(msg.level);
     auto logItem = new LogItem( fmt::to_string(msg.payload), level, msg.source.filename);
-    pRoot->Log(logItem);
+    g_root->Log(logItem);
   }
 
   void flush_() override {}
--- a/src/main/Root.cpp
+++ b/src/main/Root.cpp
@@ -23,15 +23,17 @@
 
 #include <filesystem>
 
-VGMRoot *pRoot;
+VGMRoot* g_root;
 
-VGMFile* variantToVGMFile(VGMFileVariant variant) {
-  VGMFile *vgmFilePtr = nullptr;
-  std::visit([&vgmFilePtr](auto *vgm) { vgmFilePtr = static_cast<VGMFile *>(vgm); }, variant);
+VGMFile* variantToVGMFile(VGMFileVariant variant)
+{
+  VGMFile* vgmFilePtr = nullptr;
+  std::visit([&vgmFilePtr](auto* vgm) { vgmFilePtr = static_cast<VGMFile*>(vgm); }, variant);
   return vgmFilePtr;
 }
 
-VGMFileVariant vgmFileToVariant(VGMFile* file) {
+VGMFileVariant vgmFileToVariant(VGMFile* file)
+{
   if (auto seq = dynamic_cast<VGMSeq*>(file)) {
     return seq;
   } else if (auto instrSet = dynamic_cast<VGMInstrSet*>(file)) {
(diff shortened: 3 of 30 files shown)


sykhro:
> I went with this since it seems very common, but I don't feel strongly about it.

Since we neither feel strongly about anything, at this point it's probably better to use a `clang-format` style that ends up making the smallest amount of changes to the existing code to avoid losing too much in `git blame`. `clang-power-tools` offers a `clang-format` detector, you feed it a bunch of source files and it spits out a `.clang-format`.

Also, commits that do nothing but rename variables should be added to `.git-blame-ignore-revs` too

sykhro (changes requested):
See PR comment

mikelow:
> Since we neither feel strongly about anything, at this point it's probably better to use a clang-format style that ends up making the smallest amount of changes to the existing code to avoid losing too much in git blame. clang-power-tools offers a clang-format detector, you feed it a bunch of source files and it spits out a .clang-format.

I'm fine with making minimal changes. Granted, the variable, parameter, and function naming conventions alone will spur a lot of change. The one style rule I feel strongly about - and it's an oversight I didn't bring it up before - is left-aligning pointers and references. ie:

`int* myVar` over `int *myVar`.

I'm hoping you either agree with this or don't feel strongly on it 🤞. Oddly, our .clang-format file specified left-aligned pointers before this PR, but the line `BasedOnStyle: Google` somehow overrode it.

Changing all the files to conform to any new style guide is not a high priority in my view, though I wouldn't see it as a bad thing.

> Also, commits that do nothing but rename variables should be added to .git-blame-ignore-revs too

What about when we're adding getter functions? Same principle?

sykhro:
> I'm hoping you either agree with this or don't feel strongly on it 🤞

I really don't mind what the code looks like as long as I can press ctrl+s and autoformatting takes care of it :)

> What about when we're adding getter functions? Same principle?

Adding getters shouldn't alter semantics but sometimes it does in unexpected ways. I would keep those commits

sykhro:
The style detector came up with this btw:
```
BasedOnStyle: Google
AccessModifierOffset: -2
AlignAfterOpenBracket: DontAlign
AlignOperands: DontAlign
AllowAllArgumentsOnNextLine: false
AllowAllConstructorInitializersOnNextLine: false
AllowShortEnumsOnASingleLine: false
AllowShortFunctionsOnASingleLine: None
AllowShortIfStatementsOnASingleLine: Never
AlwaysBreakAfterReturnType: All
BinPackArguments: false
BinPackParameters: false
BreakBeforeBinaryOperators: NonAssignment
BreakConstructorInitializersBeforeComma: true
BreakConstructorInitializers: AfterColon
ColumnLimit: 200
ConstructorInitializerAllOnOneLineOrOnePerLine: false
ContinuationIndentWidth: 2
Cpp11BracedListStyle: false
EmptyLineBeforeAccessModifier: Leave
FixNamespaceComments: false
IncludeBlocks: Preserve
KeepEmptyLinesAtTheStartOfBlocks: true
MaxEmptyLinesToKeep: 3
ReflowComments: false
SortIncludes: false
SpaceBeforeCpp11BracedList: true
SpaceBeforeCtorInitializerColon: false
SpaceBeforeInheritanceColon: false
SpaceInEmptyBlock: true
SpacesBeforeTrailingComments: 3
```

doesn't look too minimal though /shrug, we can keep your config then

mikelow:
> doesn't look too minimal though /shrug, we can keep your config then

It's still shorter than what we currently have. I agree it we should strive for a set of rules that minimize change we deem unimportant.

sykhro:
> > doesn't look too minimal though /shrug, we can keep your config then
> 
> It's still shorter than what we currently have. I agree that we should strive for a set of rules that minimize change we deem unimportant.

It's a double-edged sword. Since clang-format isn't stable, the more options you have the less likely it is for updates in the future to change what the formatting looks like when new options are added or modified

mikelow:
I've created a [separate PR](https://github.com/vgmtrans/vgmtrans/pull/491) that changes the root field names and adds the getter methods, but removes the .clang-format stuff for now. We can come back to that later, and it should be a separate change anyway.

PR #488, opened 2024-05-25
More CPS3 fixes

mikelow:
This fixes a bug affecting CPS3 audio quality. Interlaced CPS3 sample roms were loaded in the wrong order, creating scratchy audio. It also implements the region fine_tune value in CPS versions >= 1.3 and < 2.0 (most titles do not utilize this, but strider2 is an exception).

## Description
- add LoadOrder enum to MAMELoader, and logic for each LoadMethod to load roms in reverse order
- fix CPS romgroup entries that need to load in reverse order
- CPS: Add instrument region fine_tune byte for CPS versions >= 1.3 and < 2.0

## How Has This Been Tested?
I examined the samples under a magnifying glass (Audacity).

## Types of changes

## Checklist:

### DIFF
--- a/src/main/loaders/MAMELoader.cpp
+++ b/src/main/loaders/MAMELoader.cpp
@@ -7,12 +7,14 @@
 #include <tinyxml.h>
 #include <spdlog/fmt/fmt.h>
 #include <cstdlib>
+#include <ranges>
 
 #include "MAMELoader.h"
 #include "Format.h"
 #include "KabukiDecrypt.h"
 #include "CPS3Decrypt.h"
 #include "LoaderManager.h"
+#include "LogManager.h"
 #include "Root.h"
 #include "Scanner.h"
 #include "helper.h"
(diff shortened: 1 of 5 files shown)


src/main/loaders/MAMELoader.h
```diff
@@ -30,7 +31,7 @@ void FromString(const std::string &temp, T *out) {
 }
 
 struct MAMERomGroup {
-    MAMERomGroup() : loadmethod(LM_APPEND), file(nullptr) {}
+    MAMERomGroup() {}
```
sykhro, line 34: Compilers may output worse code when given empty constructors/destructors. See [rule of zero](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#cdefop-default-operations)

src/main/loaders/MAMELoader.cpp
```diff
@@ -21,339 +23,405 @@ namespace vgmtrans::loaders {
...
+void MAMELoader::apply(const RawFile* file)
+{
+  if (!bLoadedXml || file->extension() != "zip")
+    return;
+
+  std::string fullfilename = file->name();
+  size_t endoffilename = fullfilename.rfind('.');
+  if (endoffilename == std::string::npos)
+    return;  // no '.' found in filename, so don't do anything
+  std::string filename = fullfilename.substr(0, endoffilename);
```
sykhro, line 190: Seeing this code made me realize that we really could use adding a `path()` getter to RawFile that returns `std::filesystem::path`, it would eliminate a lot of these instances of "try to find the extension" with better performance

sykhro (changes requested):
No deal-breakers but worth addressing since the goal is improving code quality

src/main/loaders/MAMELoader.h
mikelow: I'm confused about the relation between the linked guideline rule and this code.

Is the idea here that it would be better to assign default values to `loadmethod` and `file` in the initialier list than to assign at declaration via `loadmethod{LM_APPEND}`, `file{nullptr}`?

What would you recommend here?

sykhro: Ah, what I'm saying is that `MAMERomGroup() {}` can be removed entirely.

src/main/loaders/MAMELoader.cpp
```diff
@@ -21,339 +23,405 @@ namespace vgmtrans::loaders {
...
+MAMERomGroup* MAMEGame::GetRomGroupOfType(const std::string& strType)
+{
+  for (std::list<MAMERomGroup>::iterator it = romgroupentries.begin(); it != romgroupentries.end();
+       ++it) {
+    if (it->type.compare(strType) == 0)
+      return &(*it);
+  }
 
-    return nullptr;
+  return nullptr;
```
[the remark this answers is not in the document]
sykhro: (the reason for this suggestion is that the compiler does a better job at vectorizing this loop)

src/main/loaders/MAMELoader.cpp
mikelow: Good call. I can make another PR to add this.

src/main/loaders/MAMELoader.cpp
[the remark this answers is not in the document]
mikelow: Awesome. Makes sense.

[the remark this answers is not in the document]
mikelow: There was a syntax error (regarding semicolon placement) and a couple warnings with that snippet. Let me know if this revision is good:

```
if (auto group = std::ranges::find_if(
        romgroupentries, [&strType](const auto& group) { return group.type == strType; });
    group != romgroupentries.end()) {
  return std::addressof(*group);
}
```

mikelow:
Updated. Left a comment detailing the changes I made to that std::ranges::find_if() call.

sykhro (approved):
LGTM, thanks!

PR #490, opened 2024-05-25
Use CMakePresets, misc CI improvements

sykhro:
## Description
- Switch to using CMakePresets to manage CMake settings instead of baking them inside the root CMakeLists
- Add Windows ARM64 job

### CMake
All CMake projects should be pluggable and should not define properties of the compiler. Flags and other settings are to be put in external setting files (toolchain files) to prevent potential clashes.
`cmake-presets` is a "new" way of managing CMake settings, acting more or less like VS solution configurations.  It was introduced with CMake 3.19 but we require functionality that was added with CMake 3.21.

With this change, CMakeLists is cleaned up and compiling is simpler. For example, on Windows:
```
cmake --preset win-x64
cmake --build --preset win-x64-release
```
will set up & compile for Windows x64 in Release mode.
Many IDEs speak cmake-presets too (VS, vscode, CLion, ...), prompting devs with handy selection menus:
image

All of the flags have been places in the presets file since we don't have that many. Older flags have been dropped since those compilers are too old for c++20 anyway.

Other misc notes:
- ASAN support is to be left for a future PR. However, if needed, users can easily add `-fsanitize=address` to the CXXFLAGS in CMakeCache.txt and recompile.
- The default build directory now specifies the configuration name to make it easier to switch between different trees, e.g. toggling between compiling for Intel or Apple Silicon.
- `ninja` is not meant to be mandatory but **strongly recommended**, especially as of 1.12: build speed is unparalleled. so, pass `-G "Ninja Multi-Config"` at the configure stage
  - Windows has "Ninja Multi-Config" hardcoded until I can do more testing
- Windows ARM64 is CLI-only for now, need to (cross-)compile Qt first (sigh)
- We also have warnings in release-mode. It makes no sense to hide them

### CI
- Don't install recommended packages on Linux, makes CI boot up faster
- Compile with Clang 18 on Linux
- Stop using deprecated Node v16 actions

## How Has This Been Tested?
Tested on macOS with XCode 15 Command Line Tools. 

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,27 +112,28 @@ jobs:
     runs-on: windows-2022
     strategy:
       matrix:
-        target: ["x64"]
+        target: ['x64', 'arm64']
     name: Windows, ${{ matrix.target }}
     steps:
        - uses: actions/checkout@v4
          with:
           submodules: recursive
           show-progress: false
-       - uses: ilammy/msvc-dev-cmd@v1
+       - name: Source vcvars for ${{ matrix.target }}
+         uses: ilammy/msvc-dev-cmd@v1
          with:
-            arch: ${{ matrix.target }}
+            arch: ${{ matrix.target != 'x64' && (format('{0}_{1}', 'amd64', matrix.target)) || matrix.target }}
        - name: "Prepare build"
          run: |
           echo "BRANCH=$(git rev-parse --abbrev-ref HEAD)" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-          cmake --preset win-x64
+          cmake --preset win-${{ matrix.target }}
        - name: "Build project"
-         run: cmake --build --preset win-x64-release
+         run: cmake --build --preset win-${{ matrix.target }}-release
        - name: "Package artifact"
-         run: 7z a "VGMTrans.zip" -r .\build\win-x64\bin*
+         run: 7z a "VGMTrans.zip" -r .\build\win-${{ matrix.target }}\bin*
        - name: "Upload artifact"
          uses: actions/upload-artifact@v4
          with:
-           name: VGMTrans-${{ github.sha }}-${{ env.BRANCH }}-x86_64-${{ runner.os }}
+           name: VGMTrans-${{ github.sha }}-${{ env.BRANCH }}-${{ matrix.target }}-${{ runner.os }}
            path: VGMTrans.zip
        
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,13 +8,12 @@ on:
 jobs:
   macos:
     name: macOS 11.x+
-    runs-on: ${{ (matrix.target == 'x86_64' && 'macos-13') || 'macos-14' }}
+    runs-on: ${{ (matrix.target == 'intel' && 'macos-13') || 'macos-14' }}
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -26,24 +25,22 @@ jobs:
...
-        run: |
-          cmake -DCMAKE_OSX_DEPLOYMENT_TARGET=${{ env.osx_min_ver }} \
-                -DCMAKE_OSX_ARCHITECTURES=${{ matrix.target }} \
-                -B build
+        run: cmake --preset "macos-${{ matrix.target }}"
       - name: "Build project"
-        run: cmake --build build --config "Release" --parallel
-      - working-directory: "build"
-        name: "Package DMG (macOS)"
+        # $(sysctl -n hw.logicalcpu) is required to not make XCode spawn infinite threads...
```
mikelow, line 30: Lol, what?  Was this happening before? Is it an issue with the GA virtual machine? Never had this issue, and I'm using the latest Xcode toolchain.

mikelow (approved):
Wow, this is immensely cleaner. Excellent work! 🎉

.github/workflows/build.yml
sykhro: I've been having this issue on macOS since 2020.
Invoking `cmake --build build ... --parallel` if the generator is XCode results in unlimited jobs spawning, the system eventually goes oom and slows down to a crawl. I've never tried diagnosing why but it happens across projects so I know for sure it's not our issue...

sykhro: That being said, I still recommend `-G Ninja`. It's multithreaded by default (you don't need --parallel) and with the newer releases the graph model has been greatly improved, it's a lot faster/caches better

mikelow: Ah, I didn't realize this applies only to the Xcode generator. When I've used it, I build with the Xcode IDE without issue, but never tried by command line.

sykhro:
Whoops: I accidentally click "Rebase and merge" because that's what I use at work and it was the first selection. I've noticed you've been using "Squash and merge", I will try to use that from now on. We can disable other merging strategies in settings

mikelow:
We don't have to disable the other options. Sometimes, I prefer to rebase without the squash if I think preserving the commit history will be clarifying, or may retain some change I want to refer back to.

Also, I always rebase before merging a PR. Funny I never looked into whether there's a squash and rebase option. I'd prefer that.

.github/workflows/build.yml
mikelow: Actually, I'm confused - I don't see any reference to the Xcode generator in the build.yml or CMake.presets file. Do I misunderstand? Are you having this issue when using ninja as well?

sykhro: CMake on macos auto-picks the XCode generator - I haven't actually checked that it's also happening on CI tbh, but github actions only gives us 2 cores anyway - can't hurt to cap it at that

mikelow: That's odd. I definitely don't see it using Xcode by defualt when I run `cmake -B build`.

sykhro: ah, that's interesting.. maybe it's due to the fact I have the command line tools

Issue #151, opened 2019-07-07
Individual tracks ripping


Wanny:
I'm testing VGMTrans with CPS2 songs and I noticed that some instruments seem to fall out of the regular converted 16 tracks.
For example, try converting the intro music from SFA2 and you'll notice that the bassline disappears after some secs.

So, for debugging purposes would be really welcome to have the option to export each individual original track separatedly, something like...
![image](https://user-images.githubusercontent.com/569307/60765631-6a7c8100-a06b-11e9-956b-1d3c1902bf56.png)
(mockup screenshot!)

sykhro:
PRs are welcome, the WTL ui is currently unmantained. I do plan to add this to the Qt version

PR #495, opened 2024-05-29
Rename VGMItem / VGMFile fields

mikelow:
- remove `VGMFile::name()` and `VGMFile::m_name` - they are redundant with VGMItem's name field
- rename `VGMItem::name` to `m_name`, make it private and add `name()` / `setName()`
- rename `VGMItem::vgmfile` to `m_vgmfile` and add `vgmFile()`
- rename `VGMItem::GetDescription()` to `description()`
- rename `VGMFile::GetFormat()` to `format()`
- rename `VGMFile::GetFormatName()` to `formatName()`
- add `VGMFile::setId()`
- make most VGMFile fields private

Plan is to squash this PR and add it to the .git-blame-ignore-revs list in the next PR.

## Motivation and Context
Code cleanup. Plus, it's necessary for VGMFile / RawFile to use a name() method in order to generalize the SaveAsOriginal menu command so that it can work with both VGMFile and RawFile. (see upcoming PR).

## How Has This Been Tested?
Files seem to be loading correctly and custom VGMFile/VGMItem names are still applied.

## Types of changes

- [x] Code cleanup

## Checklist:

### DIFF
--- a/src/main/Root.cpp
+++ b/src/main/Root.cpp
@@ -151,7 +151,7 @@ void VGMRoot::RemoveVGMFile(std::variant<VGMSeq *, VGMInstrSet *, VGMSampColl *,
   auto targFile = variantToVGMFile(file);
   // First we should call the format's onClose handler in case it needs to use
   // the RawFile before we close it (FilenameMatcher, for ex)
-  if (Format *fmt = targFile->GetFormat()) {
+  if (Format *fmt = targFile->format()) {
     fmt->OnCloseFile(file);
   }
 
--- a/src/main/Root.cpp
+++ b/src/main/Root.cpp
@@ -141,7 +141,7 @@ bool VGMRoot::CloseRawFile(RawFile *targFile) {
 void VGMRoot::AddVGMFile(
   std::variant<VGMSeq *, VGMInstrSet *, VGMSampColl *, VGMMiscFile *> file) {
   m_vgmfiles.push_back(file);
-  L_INFO("Loaded {} successfully.", *variantToVGMFile(file)->GetName());
+  L_INFO("Loaded {} successfully.", variantToVGMFile(file)->name());
   UI_AddVGMFile(file);
 }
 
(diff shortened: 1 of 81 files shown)


sykhro:
> Plus, it's necessary for VGMFile / RawFile to use a name() method in order to generalize the SaveAsOriginal menu command so that it can work with both VGMFile and RawFile. 

I was looking at this just a couple days ago to implement #151

sykhro (approved):

mikelow:
> I was looking at this just a couple days ago to implement https://github.com/vgmtrans/vgmtrans/issues/151, gave up because at the moment it's impossible to copy a VGMSeq reasonably (thinking about refactoring those bits)

@sykhro
Hm, I wonder if we couldn't just create a new VGMSeq and copy the SeqTrack into it. Another thought: add an optional param to SequencePlayer::playCollection() and MidiFile::ConvertToMidi() (maybe some struct with conversion options) to solo a track.

sykhro:
> > I was looking at this just a couple days ago to implement #151, gave up because at the moment it's impossible to copy a VGMSeq reasonably (thinking about refactoring those bits)
> 
> @sykhro Hm, I wonder if we couldn't just create a new VGMSeq and copy the SeqTrack into it. Another thought: add an optional param to SequencePlayer::playCollection() and MidiFile::ConvertToMidi() (maybe some struct with conversion options) to solo a track.

The problem is copying in general; we don't have any copy constructors defined and due to the myriad of raw pointers and destructors which call things like `DeleteVect` it felt very easy to end up in a use-after-free situation

mikelow:
> The problem is copying in general; we don't have any copy constructors defined and due to the myriad of raw pointers and destructors which call things like DeleteVect it felt very easy to end up in a use-after-free situation

Yeah, the pointer situation is a mess 🙁.

Coincidentally, I was thinking about this functionality recently myself, but I was focused on the UI implications - whether we should move the play controls outside of the collection list view to somewhere more global.

sykhro:
> Yeah, the pointer situation is a mess 🙁.

I will try a small-scale refactor with `unique_ptr`, that's what we should be aiming for ideally

mikelow:
> I will try a small-scale refactor with unique_ptr, that's what we should be aiming for ideally

That would be amazeballs!

PR #507, opened 2024-06-04
tinyxml2 port

TheComputerGuy96:
This was originally intended for fixing the old-style cast warnings (but it looks like tinyxml2 still uses them) so that warning either has to be disabled or the library needs to be fixed upstream :frog: 

I tested this change a bit and found a few segfaults (which I fixed)


sykhro:
> This was originally intended for fixing the old-style cast warnings (but it looks like tinyxml2 still uses them) so that warning either has to be disabled or the library needs to be fixed upstream 🐸

ah, the joy of vendoring dependencies. I would suggest testing it some more with address sanitizer (edit CMakeCache.txt to add `-fsanitize=address` to `CMAKE_{C, CXX}_FLAGS` and recompile)

PR #509, opened 2024-06-05
Remove VGMContainerItem

mikelow:
This is a refactor that removes the VGMContainerItem class. In its place, VGMItem is given a vector of child VGMItems (`std::vector<VGMItem*> m_children`), effectively making it a tree.

There are other changes made as a consequence of removing VGMContainerItem:
- a new addChildren() method is now used in place of AddContainer(). Discussion on this further down.
- VGMItem is now the only class implementing GetItemFromOffset(). The `includeContainer` parameter was removed as it was unused.
- VGMInstrSet::aRgns was renamed to m_regions and made private. Classes derived from VGMInstrSet now need to use AddRgn() instead of pushing onto a vector.

## Motivation and Context
VGMContainerItem is confusing and unnecessary.

## Discussion

*Pros and Cons*

Pros:
- Simplicity
  * old: VGMContainer is a mess. It stores a vector of headers, a vector of vectors (containers), and a vector of "localitems". It coexists with VGMItem, which it also inherits from. Only VGMContainerItems may contain children, and they use different rules for GetItemFromOffset(), so leaf nodes should not be VGMContainerItems.
  *  new: VGMItem is a simple tree. Any VGMItem can have child items. There is no VGMContainerItem.
- Flexibility
  * old: VGMContainer's item ordering is inflexible and ambiguous. Headers precede local items. Local items precede any additionally added containers.
  * new: VGMItem has a single vector of children and doesn't manipulate their order. 
- Memory cleanup
  * old: Component classes must delete their container vectors (aTracks, aInstrs, etc).
  * new: ~VGMItem() deletes all children. The common container vectors are automatically added as children in the load logic of the component base classes.

Cons:
- Duplicate pointers
  * old: AddContainer() adds a reference to a container vector, like `aTracks`. The contents of the vector then exist in the item hierarchy while remaining solely contained in the vector.
  * new: addChildren() copies a vector of pointers into the m_children vector. Pointers to the same object exist in two places: in m_children to represent their presence in the item hierarchy, and in their original component containers like `aTracks` where they can be used for conversion.
- Containers are no longer added declaratively
  * old: AddContainer() is called in a component class's constructor.
  * new: addChildren() is called in a component class's load logic after a container is populated.

*A note on handling complex formats*

Some formats don't conform to our component paradigms. For example, the SonyPS2Instr class does not have a perfect analogue for VGMRgn. In this case, we want to prevent the m_regions vector from being added to the item hierarchy. Whereas before, we called the `RemoveContainer()` method to achieve this, I've added a `disableAutoAddRegionsAsChildren()` method which performs the same function. Since the regions aren't added to the item hierarchy, we also need to delete them in ~SonyPS2Instr(), hence the addition of `VGMInstr::deleteRegions()`. I expect that smart pointers would obviate this last step.

*GetItemFromOffset()*

GetItemFromOffset is now a simple recursive function. 
  * Base case: If an item has no children, return itself if it exists at the address, otherwise return nullptr. 
  * If an item has children, recursively call GetItemFromOffset() on each child until one returns an instance. If none return an instance, return nullptr. 
  * Note that if a child has children but none are contained at the offset, we always return nullptr instead of checking whether the item itself is contained at the offset. This resolves unwanted matching when there are overlapping items and some are sparsely populated.

## How Has This Been Tested?
I tested against every format and made fixes where necessary. There should be no regression.

## Types of changes

- [x] Refactor

## Checklist:


sykhro:
Great work! Happy to see this gone. I will review it asap, some thoughts below.

---

> addChildren() copies a vector of pointers into the m_children vector. Pointers to the same object exist in two places: in m_children to represent their presence in the item hierarchy, and in their original component containers like aTracks where they can be used for conversion.

This is something that definitely benefits being modeled with smart pointers since it makes the relationship clear.
The owner of the pointer (the original component) should be `unique_ptr`. The one in the VGMItem can be raw pointers

> Containers are no longer added declaratively

Not a terrible loss I'd say given that none of our classes are RAII. That'd be a big refactor though.

> Since the regions aren't added to the item hierarchy, we also need to delete them in ~SonyPS2Instr(), hence the addition of VGMInstr::deleteRegions(). I expect that smart pointers would obviate this last step.

Yes, as I was saying above

> GetItemFromOffset 

I've been wondering for a while about an alternative data structure to make this faster, but it only seemed to be problematic on Windows with the debug CRT

> is now a simple recursive function

I admire the elegancy - how is it in debuggability terms?

src/main/components/seq/VGMSeqNoTrks.cpp
```diff
@@ -27,17 +26,23 @@ void VGMSeqNoTrks::ResetVars() {
 
 // LoadMain() - loads all sequence data into the class
 bool VGMSeqNoTrks::LoadMain() {
-  this->SeqTrack::readMode = this->VGMSeq::readMode = READMODE_ADD_TO_UI;
+  this->SeqTrack::readMode = READMODE_ADD_TO_UI;
+  this->VGMSeq::readMode = READMODE_ADD_TO_UI;
```
sykhro, line 30: I never noticed these were there. My mental model of "readmode" is probably wrong

src/main/formats/Akao/AkaoInstr.cpp
```diff
@@ -66,11 +66,11 @@ AkaoInstrSet::AkaoInstrSet(RawFile *file, uint32_t offset,
 
 bool AkaoInstrSet::GetInstrPointers() {
   if (bMelInstrs) {
-    VGMHeader *SSEQHdr = AddHeader(instrSetOff, 0x10, "Instr Ptr Table");
+    VGMHeader *SSEQHdr = addHeader(instrSetOff, 0x10, "Instr Ptr Table");
     int i = 0;
     //-1 aka 0xFFFF if signed or 0 and past the first pointer value
     for (int j = instrSetOff; (GetShort(j) != static_cast<uint16_t>(-1)) && ((GetShort(j) != 0) || i == 0) && i < 16; j += 2) {
-      SSEQHdr->AddSimpleItem(j, 2, "Instr Pointer");
+      SSEQHdr->addSimpleChild(j, 2, "Instr Pointer");
```
sykhro, line 73: How strong do you feel about "Simple" here? I've always found it a bit confusing

mikelow: Let's change it then! What do you think? addItem()?

src/main/components/seq/VGMSeqNoTrks.cpp
mikelow: 🤦‍♂️ readmode was there from the start, meaning it's not well-designed. The idea is to reuse the load methods for different passes: first to create VGMItems (ie the elements the UI is based on), and subsequent passes to perform MIDI conversion. Somewhere along the line I added READMODE_FIND_DELTA_LENGTH (a misnomer) to determine the duration of the sequence in an initial pass.

VGMSeqNoTrks is a small abomination. It's using multiple inheritance in a diamond pattern without virtual inheritance. 🫣

src/main/formats/Akao/AkaoInstr.cpp
sykhro: yes, `addItem()` works fine, unless you think it might get confusing with other things. I've seen about half of the files so far btw, I'll get to the rest soon

src/main/components/seq/VGMSeqNoTrks.cpp
sykhro: > (a misnomer)

ah, that explains it!

> VGMSeqNoTrks is a small abomination. It's using multiple inheritance in a diamond pattern without virtual inheritance. 🫣

Is there a reason why it exists in the first place? What's the difference with a `VGMSeq` with empty `aTracks`?

src/main/formats/Akao/AkaoInstr.cpp
mikelow: Oops, I actually think keeping "child" in the name is more clarifying. Perhaps an override of addChild()? addBasicChild()? I think I like the former most.

sykhro: yes, I like `addChild()` better too. it makes it very clear what's the relationship with the item

src/main/components/seq/VGMSeqNoTrks.cpp
mikelow: It's been a _long_ time since I thought about this class. I'd need to study it a bit more, but I'm pretty sure it makes sense it to exist, it's just not designed well. No matter what, we need a SeqTrack, since that provides us all of the methods for parsing events. Conceptually these sequences are like MIDI Type 0, which is to say that they technically consist of a single track.

I suppose the key addition is the SetCurTrack() method, which swaps which MIDI track we're writing to during conversion.  I opted to convert these single track / multiple channel sequences into multiple tracks that use a single channel (ie MIDI Type 1). I'd still make that decision today, since the output is cleaner and there's no downside.

I might be forgetting something, but I think the only real difference between this class and VGMSeq is that we need it to add the events of the single track as root children of the sequence and provide a method like SetCurTrack(). The excessive Load method overriding and multiple inheritance is likely unnecessary.

src/main/formats/Akao/AkaoInstr.cpp
mikelow: 100% agree. I went ahead and renamed it to addChild()

sykhro (approved):
LGTM! I hope I didn't miss anything this time - I will figure out a testing pipeline, someday...

mikelow:
> I admire the elegancy - how is it in debuggability terms?

Honestly, didn't have to debug it, but if I did, I don't think it will be an issue. This is typically only going to recurse 2 or 3 times.

mikelow:
> LGTM! I hope I didn't miss anything this time - I will figure out a testing pipeline, someday...

If we were to add tests, what library(ies) would you recommend for this purpose?

sykhro:
> > LGTM! I hope I didn't miss anything this time - I will figure out a testing pipeline, someday...
> 
> If we were to add tests, what library(ies) would you recommend for this purpose?

Catch2 or gtest. I like the former better, the latter has been around longer though

PR #514, opened 2024-06-09
HexView optimizations

mikelow:
A lot of little performance optimizations for HexView:
- hack fix for paintEvents being called for all open subwindows of the MDI, not just the active one
- draw the selected item pixmap by copying segments of cached line pixmaps when possible (virtually, always) instead of calling the various print() methods to draw from scratch
- optimize paintEvent for happy path - construct a QPixmap and QPainter only when a line isn't in cache.
- memoize getVirtualWidth() methods calculated values. They are invalidated on font change
- memoize a new charHalfWidth() value, also invalidated on font change
- only redraw the overlay when window height expands beyond its current height (don't redraw when window height shrinks)
- skip drawing lines that are totally eclipsed by the selected item
- MacOS: detect paint events intended to draw the scroll bar which don't overlap with our draw area. Don't redraw lines when this happens.
- check that event type is Paint before calling `handleOverlayPaintEvent()` and `handleSelectedItemPaintEvent()` instead of checking in the methods.

Also:
- brighten the selection overlay a wee bit (the dark background)
- define drop shadow constants and tweak them to give the selected item a little more pop

I may run some benchmarks on `handleSelectedItemPaintEvent()` to see if the new method of drawing the selected item is truly improving performance. Best I can tell, it does, marginally.

## How Has This Been Tested?
Tried my best to break it.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/workarea/HexView.cpp
+++ b/src/ui/qt/workarea/HexView.cpp
@@ -422,14 +422,22 @@ bool HexView::handleSelectedItemPaintEvent(QObject* obj, QEvent* event) {
         pixmapPainter.save();
         pixmapPainter.translate(0, line * lineHeight);
 
-        auto linePixmap = lineCache.object(startLine + line);
         int bytesToPrint = std::min(
-                              std::min(static_cast<int>(selectedItem->unLength) - offsetIntoEvent,BYTES_PER_LINE - col),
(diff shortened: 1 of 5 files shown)


sykhro (approved):
Thanks! It definitely adds up. Logic seems okay to me, can't test at the moment

PR #516, opened 2024-06-12
HexView optimizations part 2

mikelow:
This is a follow up to the previous PR #514. I will merge them together, because...

#514 added a regression. One thing I tacked on was brightening the selection overlay by reducing the final opacity of the animated QGraphicsOpacityEffect.  It turns out that an overlaid widget using a QGraphicsOpacityEffect with an opacity value between 0 and 1 (exclusive) hurts draw performance. The solution is to animate to a full 1.0 opacity (as before), and instead reduce the alpha of the fill color that we use to paint the overlay. This is the cleaner solution anyway.

Other changes:

- Solved a bug that was causing the full QScrollArea viewport to be invalidated on a scroll - ie. HexView was redrawing every visible line (albeit with pixmap caches) on each draw pass. The bug is that we're setting the width of HexView greater than its QScrollArea's viewport. A little odd that it results in this behavior.
- The overlay is now sized (arbitrarily) 5 times taller than the viewport and moved only when it is scrolled out of view. The reason is to reduce the need to constantly move it on scroll. Before, when an item was selected and the overlay was therefore visible, we would move the overlay on every scroll event. However, this invalidates the entire area that it is moved to, and thus every line needed to be redrawn on every frame. This problem was shadowed by the previous issue.
- The endLine value, used to determine which lines to redraw, was over-calculated. We were always drawing an extra line on each draw pass.

Happy to say that this has all adds up to noticeably more fluid HexView.

## How Has This Been Tested?
Thoroughly tested on Mac. Will test it on Windows soon.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/util/UIHelpers.cpp
+++ b/src/ui/qt/util/UIHelpers.cpp
@@ -14,7 +14,7 @@
 #include <QStandardPaths>
 #include <QApplication>
 
-QScrollArea* getContainingScrollArea(QWidget* widget) {
+QScrollArea* getContainingScrollArea(const QWidget* widget) {
   QWidget* viewport = widget->parentWidget();
   if (viewport) {
     auto* scrollArea = qobject_cast<QScrollArea*>(viewport->parentWidget());
--- a/src/ui/qt/util/UIHelpers.h
+++ b/src/ui/qt/util/UIHelpers.h
@@ -14,7 +14,7 @@ class QPixmap;
 class QGraphicsEffect;
 class VGMItem;
 
-QScrollArea* getContainingScrollArea(QWidget* widget);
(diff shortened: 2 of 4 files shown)


sykhro (approved):
> One thing I tacked on was brightening the selection overlay by reducing the final opacity of the animated QGraphicsOpacityEffect. It turns out that an overlaid widget using a QGraphicsOpacityEffect with an opacity value between 0 and 1 (exclusive) hurts draw performance.

I'm surprised by this, but the solution is much better

sykhro:
Does this work across different DPIs?

mikelow:
> Does this work across different DPIs?

Good question. Just gave it a test on DPI of 1, and it's working well. Been running it on DPI 2 otherwise. So all good.

PR #519, opened 2024-06-15
Hexview: improve animation and performance

mikelow:
On the animation side:
- Animate the `blurRadius` and `offset` properties of the selectionView's QGraphicsDropShadowEffect, which is now applied to the widget instead of exclusively used at draw time
- On deselection, instead of immediately hiding the selected item view (which can cause a sort of flash effect), keep it visible and reverse animate its shadow effect alongside the overlay fade in, then hide it afterward.
- I also deepened the shadow by increasing the offset.

On the performance side:
- I refactored the animation code. It's much simpler now. Instead of deleting and recreating animations, it now uses pause() and setDirection() for animation interruption and reversal. The three QPropertyAnimations are bucketed into a QParallelAnimationGroup.
-  Both the `overlay` and `selectionView` have added logic to only draw the portions of their views that were invalidated (the event->rect() property). Before, if the selectionView was at all visible, it would be drawn in entirety. Likewise, now that I've expanded the size of the overlay to be larger than the viewport (for performance reasons explained in #516), resize events and the fade animations were causing it to redraw itself in entirety. Now we just redraw the portion inside the viewport.

## Details
You may be wondering why I didn't have the paintEvent handlers behave this way in the first place. I was confused. When I first designed my implementation, I assumed Qt would render like a compositor where each widget would retain its own bitmap buffer - thus we would draw the widgets (selectionView, overlay) once, move them where they need to be on top of other widgets, and expect Qt to composite them efficiently. Turns out, it doesn't keep a buffer, and everything needs to be redrawn constantly. It's unfortunate, because this paradigm is less intuitive and would render much more efficiently if it utilized the native compositor of MacOS (can't speak for Windows or Linux).

Back to the nitty gritty:  In order to apply an animation to the selectionView's shadow, we now must apply a QGraphicsDropShadowEffect to the selectionView, whereas before we would only create the shadow effect within the paintEvent rendering itself (which was then saved to a pixmap cache). To prevent a degradation in performance, we now cache two QPixmap's: one for the raw text rendering, and one for the text rendering with full shadow effect (as we did before). During the fade in and fade out animations, we enable the shadow effect on the view and draw the raw pixmap. In the "finished" animation callback, we disable the effect and draw the shadowed pixmap from thereon, as this saves performance over keeping the effect enabled.

## How Has This Been Tested?
Trying to break it as best I can on MacOS, soon Windows as well.

## Video:
Old:

https://github.com/vgmtrans/vgmtrans/assets/1659535/31483ddc-6502-4e12-a5c7-43be3dcdffa1

New:

https://github.com/vgmtrans/vgmtrans/assets/1659535/c7b5b040-935e-4ec7-a841-d8b9a757a48d

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/main_ui.cpp
+++ b/src/ui/qt/main_ui.cpp
@@ -16,6 +16,10 @@ int main(int argc, char *argv[]) {
   QCoreApplication::setOrganizationDomain("vgmtrans.com");
   QCoreApplication::setApplicationName("VGMTrans");
 
+#ifdef Q_OS_LINUX
+  QGuiApplication::setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::RoundPreferFloor);
+#endif
+
   QApplication app(argc, argv);
   #ifdef _WIN32
   app.setStyle(QStyleFactory::create("fusion"));
--- a/src/ui/qt/main_ui.cpp
+++ b/src/ui/qt/main_ui.cpp
@@ -16,9 +16,7 @@ int main(int argc, char *argv[]) {
   QCoreApplication::setOrganizationDomain("vgmtrans.com");
   QCoreApplication::setApplicationName("VGMTrans");
 
-#ifdef Q_OS_LINUX
-  QGuiApplication::setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::RoundPreferFloor);
-#endif
+  QGuiApplication::setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::Round);
 
   QApplication app(argc, argv);
   #ifdef _WIN32
--- a/src/ui/qt/workarea/HexView.cpp
+++ b/src/ui/qt/workarea/HexView.cpp
@@ -201,7 +201,6 @@ void HexView::setSelectedItem(VGMItem *item) {
 
   if (selectedItem == nullptr) {
     showOverlay(false, true);
-    // selectionView->hide();
     prevSelectedItem = nullptr;
     return;
   }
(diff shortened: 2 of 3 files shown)


sykhro:
> You may be wondering why I didn't have the paintEvent handlers behave this way in the first place. I was confused. 

I made the same assumption (welp?)

>  would render much more efficiently if it utilized the native compositor of MacOS (can't speak for Windows or Linux)

iirc there's a "qt extras" of sorts with some OS-specific things (I recall being able to do Aero Glass with it back in the day), no clue if it's even supported as of Qt6

sykhro (approved):

mikelow:
@sykhro I made one important change that I want to run by you: I've added a call to setHighDpiScaleFactorRoundingPolicy() to force the scaling to an integer value. In the course of testing on Windows and Ubuntu, I discovered rendering artifacts related to floating pointing scale factors. 

On Windows, when the OS is set to scale at 150%, which results in a device pixel ratio (DPR) of 1.5, there was jitter with the newly added shadow animations. I tried many different things to fix this but found no good solution - the best was forcing the verticalScrollBar position values to be odd, but this is probably affected by the arbitrary font line height. Outside of the animations, there are very minor artifacts with selected items sometimes rendering a pixel off.  At 125% or 175%, the aforementioned issues get worse, and there are gaps between some of the lines of text.

On Ubuntu, the situation is worse. On my 4k monitor, setting the OS display settings to 100% results in a 1.25 DPR, and it remains 1.25 for all scaling sizes until we get to 200% (2.5 dpi). Artifacting is as bad as Windows at 1.25.

There is no artifacting when the DPR is an integer, 1 or 2. On MacOS, no matter the system scaling setting, the DPR is an of either 1 or 2.

Rounding the DPR to an integer with setHighDpiScaleFactorRoundingPolicy() totally fixes the problem, and I know of no other solution.

sykhro:
> Rounding the DPR to an integer with setHighDpiScaleFactorRoundingPolicy() totally fixes the problem, and I know of no other solution.

Can't think of anything else too, I had similar issues in the past on Linux and I just gave up (X vs Wayland is another minefield). This is not the kind of software where scaling the ui precisely is vital, so it's fine by me

PR #525, opened 2024-07-30
Fix FileNameMatcher and VirtFile::extension()

mikelow:
A couple issues addressed here:
- FileNameMatcher never sets the id for any of its file types, and is therefore totally broken. 
- VirtFile::extension() returns the extension of its parent raw file by default. This is undesirable, as we would usually prefer to get the extension of the filename of the VirtFile. For example, when extracting a psf2, the SonyPS2 scanner needs to know the extension of the sample collection file.

I've fixed FileNameMatcher and reversed the VirtFile::extension() logic so that it returns the VirtFile name extension first, and if none exists, the parent RawFile's extension.

This fixes detection of SonyPS2 collections.

## How Has This Been Tested?
I tested that every format is still detected correctly.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/components/Matcher.h
+++ b/src/main/components/Matcher.h
@@ -297,19 +297,19 @@ class FilenameMatcher : public SimpleMatcher<std::string> {
 
   bool seqId(VGMSeq *seq, std::string &id) override {
     RawFile *rawfile = seq->rawFile();
-    id = rawfile->path();
+    id = rawfile->path().string();
     return !id.empty();
   }
 
   bool instrSetId(VGMInstrSet *instrset, std::string &id) override {
     RawFile *rawfile = instrset->rawFile();
-    id = rawfile->path();
+    id = rawfile->path().string();
     return !id.empty();
   }
 
   bool sampCollId(VGMSampColl *sampcoll, std::string &id) override {
     RawFile *rawfile = sampcoll->rawFile();
-    id = rawfile->path();
+    id = rawfile->path().string();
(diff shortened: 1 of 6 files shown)


sykhro (approved):
Sounds reasonable. I would like to remove the Raw/Virt distinction eventually and use memory-mapped files. I think I had that in my branch?

mikelow:
> Sounds reasonable. I would like to remove the Raw/Virt distinction eventually and use memory-mapped files. I think I had that in my branch?

I think I pulled in your work, but correct me if I missed something. The code uses [mio](https://github.com/vimpunk/mio) for memory mapping. RawFile was made an abstract class with, DiskFile and VirtFile its non-abstract implementations.

sykhro:
> > Sounds reasonable. I would like to remove the Raw/Virt distinction eventually and use memory-mapped files. I think I had that in my branch?
> 
> I think I pulled in your work, but correct me if I missed something. The code uses [mio](https://github.com/vimpunk/mio) for memory mapping. RawFile was made an abstract class, with DiskFile and VirtFile its non-abstract implementations.

Oh yes I remember this, I get the feeling I might never have pushed the refactor where I abstracted both of those away

mikelow:
> Oh yes I remember this, I get the feeling I might never have pushed the refactor where I abstracted both of those away

Oh dang. Hopefully you didn't lose the work. If you're able to push it in some form, I'll gladly work on integrating it.

PR #526, opened 2024-08-02
Akao Format improvements

mikelow:
This provides three improvements to the Akao format:

1. Adds support for creating collections that use more than one sample collection (Vagrant Story, FF8, others?)
2. Adds a new AkaoMatcher class that better handles the complicated task of determining what files belong in an Akao collection
3. Combines instrument region ADSR with articulation data ADSR, as opposed to only utilizing the articulation data ADSR. This fixes a lot of tracks with broken-sounding instruments.

## Description
Heads up that the Akao format itself is a complete mess. For the sake of matching, the relevant facts are:

1. Sequences and instrument sets are bundled together effectively as one file in their original format, so they are given the same id when we create them in the scanner.
2. Sequences may contain a byte that indicates the id of an associated sample collection, depending on the version. However, a sequence may utilize a second common sample collection for which no id is provided. Final Fantasy VIII and Vagrant Story are examples of games that do this.
3. PSF sets sometimes exclude the aforementioned sample id association byte or the id byte of the sample collection itself. This is because removing all unread bytes is an optional feature of the PSF creation software, and these bytes are never actually read by the sound driver during playback.
4. Sample collections contain a list of "articulations." Each have an id, point to a sample, and provide ADSR data.
5. Each instrument region, defined in the instrument set, points to an articulation ID.

Therefore, we are limited in terms of guaranteed information. The goal is to ensure that the articulation IDs used in the regions of the instrument set are all present and accounted for, and there may be multiple sample collections that define the used articulation ids even if their articulations are unique from each other - so there is some unavoidable guessing. The implemented strategy is to utilize the associated sample collection if it is detected, but otherwise use any sample collection that covers any referenced articulation ids. 

However, we could get better results if the AkaoMatcher class had a callback informing it that a RawFile has finished loading.  Consider what happens when a PSF is loaded. A sequence may point to an associated sample collection ID, but we have no idea if the sample collection will have its id byte present. We also don't know when we have finished loading all the files of the PSF, so we must assume that each file is the last and check if it can be used in the collection. 

If instead the Matcher were informed when RawFile has finished loading, it could first evaluate if a sample collection with the correct ID was loaded. If not, it could check if the RawFile was a PSF, in which case it would make its best attempt to find the correct sample collection(s) exclusively from the files just loaded.

To solve this, I created #527 to provide the Scanner class with a pointer to its Format. This way, a Scanner can get the Format's Matcher instance and inform the Matcher when it has finished loading files. I have also now posted #528, which puts this into practice in AkaoMatcher, building off of this PR and #527.

## How Has This Been Tested?
I've tested this on a large number of Akao sets. In terms of matching results, it will sometimes perform better or worse compared to FilegroupMatcher, ~but the cons should be addressed in a future PR once we provide the Matcher with a new callback~, but #528 basically fixes everything.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/formats/Akao/AkaoFormat.cpp
+++ b/src/main/formats/Akao/AkaoFormat.cpp
@@ -12,34 +12,75 @@
 
 bool AkaoColl::loadMain() {
   AkaoInstrSet *instrset = reinterpret_cast<AkaoInstrSet *>(instrSets()[0]);
-  AkaoSampColl *sampcoll = reinterpret_cast<AkaoSampColl *>(sampColls()[0]);
+  std::vector<AkaoSampColl*> orderedSampColls;
+  orderedSampColls.reserve(sampColls().size());
+  std::transform(sampColls().begin(), sampColls().end(), std::back_inserter(orderedSampColls),
+                     [](VGMSampColl* vgm) {
+                         return static_cast<AkaoSampColl*>(vgm);
+                     });
 
-  //Set the sample numbers of each region using the articulation data references of each region
-  for (uint32_t i = 0; i < instrset->aInstrs.size(); i++) {
-    auto instr = dynamic_cast<AkaoInstr*>(instrset->aInstrs[i]);
+  // Sort the sample collections by starting_art_id
+  std::sort(orderedSampColls.begin(), orderedSampColls.end(), [](AkaoSampColl *a, AkaoSampColl *b) {
+    return a->starting_art_id < b->starting_art_id;
+  });
 
-    for (const auto vgmregion : instr->regions()) {
-      const auto rgn = dynamic_cast<AkaoRgn*>(vgmregion);
-      AkaoArt *art;
+  // Maps for articulation ids
+  std::map<int, AkaoArt*> artIdToArtMap;
+  std::map<int, int> artIdToSampleNumMap;
+  std::map<int, AkaoSampColl*> artIdToSampCollMap;
 
-      if (!(static_cast<int32_t>(rgn->artNum) - static_cast<int32_t>(sampcoll->starting_art_id) >= 0 &&
-          rgn->artNum - sampcoll->starting_art_id < 200)) {
+  int cumulativeSamples = 0;
 
-        L_ERROR("Articulation #{:d} does not exist in the samp collection", rgn->artNum);
-        art = &sampcoll->akArts.front();
-      } else {
-        if (rgn->artNum - sampcoll->starting_art_id >= sampcoll->akArts.size()) {
-          L_ERROR("Articulation #{:d} referenced but not loaded", rgn->artNum);
-          art = &sampcoll->akArts.back();
+  for (auto *sampcoll : orderedSampColls) {
+    for (int i = 0; i < sampcoll->nNumArts; ++i) {
+      int artId = sampcoll->starting_art_id + i;
+      artIdToArtMap[artId] = &sampcoll->akArts[i];
+      artIdToSampleNumMap[artId] = sampcoll->akArts[i].sample_num + cumulativeSamples;
+      artIdToSampCollMap[artId] = sampcoll;
+    }
+    cumulativeSamples += sampcoll->samples.size();
+  }
+
+  for (auto *vgminstr : instrset->aInstrs) {
+    auto* instr = dynamic_cast<AkaoInstr*>(vgminstr);
+    auto regions = instr->regions();
+    for (auto *vgmregion : regions) {
+      auto *rgn = dynamic_cast<AkaoRgn*>(vgmregion);
+      auto itArt = artIdToArtMap.find(rgn->artNum);
+
+      if (itArt == artIdToArtMap.end()) {
+        L_ERROR("Articulation #{:d} does not exist in the sample collection", rgn->artNum);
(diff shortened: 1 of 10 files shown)


src/main/formats/Akao/AkaoFormat.cpp
```diff
@@ -5,41 +5,89 @@
...
   AkaoInstrSet *instrset = reinterpret_cast<AkaoInstrSet *>(instrSets()[0]);
-  AkaoSampColl *sampcoll = reinterpret_cast<AkaoSampColl *>(sampColls()[0]);
 
-  //Set the sample numbers of each region using the articulation data references of each region
-  for (uint32_t i = 0; i < instrset->aInstrs.size(); i++) {
-    auto instr = dynamic_cast<AkaoInstr*>(instrset->aInstrs[i]);
+  auto [artIdToArtMap, artIdToSampleNumMap, artIdToSampCollMap] = mapSampleCollections();
+
+  for (auto *vgminstr : instrset->aInstrs) {
+    auto* instr = dynamic_cast<AkaoInstr*>(vgminstr);
```
sykhro, line 52: this might be null if the `dynamic_cast` fails. probably impossible in this codepath but worth keeping in mind

src/main/formats/Akao/AkaoMatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        seqs.erase(seq->seq_id);
+
+        if (!coll->load()) {
+          delete coll;
+          return false;
+        }
+
+        std::cout << "Collection created successfully!" << std::endl;
```
sykhro, line 144: debug leftover?

src/main/formats/Akao/AkaoFormat.cpp
```diff
@@ -5,41 +5,89 @@
...
+    auto regions = instr->regions();
```
sykhro, line 53: is this copy intentional?

src/main/formats/Akao/AkaoFormat.h
```diff
@@ -13,14 +14,17 @@ class AkaoInstrSet;
...
-  explicit AkaoColl(std::string name = "Unnamed Collection") : VGMColl(std::move(name)), origInstrSet(nullptr), numAddedInstrs(0) {}
+  explicit AkaoColl(std::string name = "Unnamed Collection") : VGMColl(std::move(name)), origInstrSet(nullptr), numInstrsToAdd(0) {}
 
   bool loadMain() override;
   void preSynthFileCreation() override;
   void postSynthFileCreation() override;
 
   AkaoInstrSet *origInstrSet;
-  uint32_t numAddedInstrs;
+  uint32_t numInstrsToAdd;
```
sykhro, line 24: I would take the opportunity to move the initialization for `origInstrSet` and `numInstrsToAdd` here

sykhro (changes requested):

src/main/formats/Akao/AkaoFormat.cpp
```diff
@@ -5,41 +5,89 @@
...
 
+// Generate maps to help route the relationships to and from articulations, especially for games
+// that load more than one sample collection at a time.
+std::tuple<std::map<int, AkaoArt*>, std::map<int, int>, std::map<int, AkaoSampColl*>> AkaoColl::mapSampleCollections() {
+  std::vector<AkaoSampColl*> orderedSampColls;
+  orderedSampColls.reserve(sampColls().size());
+
+  std::map<int, AkaoArt*> artIdToArtMap;
+  std::map<int, int> artIdToSampleNumMap;
+  std::map<int, AkaoSampColl*> artIdToSampCollMap;
```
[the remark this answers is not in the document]
mikelow: Yes indeed.

src/main/formats/Akao/AkaoFormat.cpp
```diff
@@ -5,41 +5,89 @@
...
+    for (int i = 0; i < sampcoll->nNumArts; ++i) {
+      int artId = sampcoll->starting_art_id + i;
+      artIdToArtMap[artId] = &sampcoll->akArts[i];
+      artIdToSampleNumMap[artId] = sampcoll->akArts[i].sample_num + cumulativeSamples;
+      artIdToSampCollMap[artId] = sampcoll;
+    }
+    cumulativeSamples += sampcoll->samples.size();
+  }
+
+  return std::make_tuple(artIdToArtMap, artIdToSampleNumMap, artIdToSampCollMap);
```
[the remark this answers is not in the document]
mikelow: Thank you!

src/main/formats/Akao/AkaoMatcher.cpp
mikelow: 💯

src/main/formats/Akao/AkaoFormat.cpp
mikelow: Nope. Must be some leftover from earlier logic. I've removed it.

src/main/formats/Akao/AkaoFormat.h
mikelow: Done and done.

sykhro (approved):

PR #527, opened 2024-08-02
Add format property to VGMScanner

mikelow:
As discussed in #526, this provides the VGMScanner class with a reference to its Format. This will allow a VGMScanner to communicate with its Matcher if necessary.

## Description
The VGMScanner constructor now takes a single `Format*` param. The ScannerManager `add()` methods now reinterpret the string passed to them not as an arbitrary id but as the name of the Format. They use this string within the scannerSpawner closure to lookup the Format in the Format registry (via the renamed method `Format::formatFromName()`). The Format is then passed to the VGMScanner.

## Motivation and Context
See #526. In short, it would be of great help for some Matcher implementations be informed when a Scanner has finished loading files. This will enable that. 

## How Has This Been Tested?
No apparent regressions. If the name of any format is wrong, it would mean passing a nullptr as the Format* arg, but no Scanners utilize the format field yet so nothing would break.

## Types of changes

## Checklist:


sykhro (approved):
LGTM

PR #530, opened 2024-08-04
Fix PSFLoader tag handling

mikelow:
- add VGMTag::fromPsfTags
- call fromPsfTags in PSFLoader and pass it to the VirtFiles it creates
- add an additional search for the "_Lib" tag because there's a PSF set that uses this key instead of the spec's "_lib"

## Description
The new PSFLoader parses PSF tags into a map, but doesn't pass them along to the VirtFiles it creates. This PR addresses that by by adding a method to construct a VGMTag from the PSF tag map and passing the result to each VirtFile. I considered replacing the VGMTag class altogether with a simple map, but realized it wouldn't be so simple as VGMTag is also being used to house extra data (like SPC dsp data) needed by one of the formats (KonamiSnes).

I don't love the partial reliance on VGMTag by AkaoScanner, but that's a separate issue.

## How Has This Been Tested?
Akao PSF sets that did not load correctly now do. The version of the Akao driver is extrapolated by inspecting the "game" tag.

## Types of changes

## Checklist:


sykhro (approved):
if you think this same pattern might reappear with other formats, I would consider a free function like `ConvertToVGMTag` that takes a PSFFile. it can live in the global scope and have overloads.
if this is the only instance, I will gladly take it as-is as a bugfix :)

mikelow:
> if you think this same pattern might reappear with other formats, I would consider a free function like `ConvertToVGMTag` that takes a PSFFile. it can live in the global scope and have overloads. if this is the only instance, I will gladly take it as-is as a bugfix :)

Hm, why a free function instead of a static psuedo-constructor for VGMTag? 

I think making the parameter a PSFFile instead of a map<string, string> would make sense if we refactor the SPCLoader code to use an equivalent SPCFile class. Otherwise, I don't see why we wouldn't stick with the map. But I'm happy to do this refactor.

sykhro:
> > if you think this same pattern might reappear with other formats, I would consider a free function like `ConvertToVGMTag` that takes a PSFFile. it can live in the global scope and have overloads. if this is the only instance, I will gladly take it as-is as a bugfix :)
> 
> Hm, why a free function instead of a static psuedo-constructor for VGMTag? Actually, if we were to change the parameter to a distinct file class and not a generic map, why not just use a plain old constructor: VGMTag(PSFFile*)?
> 
> I think making the parameter a PSFFile instead of a map<string, string> would make sense if we refactor the SPCLoader code to use an equivalent SPCFile class. Otherwise, I don't see why we wouldn't stick with the map. But I'm happy to do this refactor.

oh, my mistake, I got confused because I thought of "psftags" as "tags you find in playstation dumps". totally slipped my mind that it's used by psffloader regardless of format. 
I would've used `PSFFile` to make it clear. I tend to prefer free functions to avoid introducing dependencies into the class 

disregard what I said :)

mikelow:
Closing this and opening #531.

PR #531, opened 2024-08-05
Add SPCFile, fix tags for PSFFile

mikelow:
In light of feedback on #530, I decided to refactor SPCLoader so that it's more consistent with PSFLoader. SPC files are now loaded into an SPCFile class. Thereafter, we generate the VGMTag via tagFromSPCFile(). Likewise, I've added a tagFromPSFFile() function, and per my changes in #530, I've fixed the VirtFile constructor to actually set the RawFile::tag field.

I've also added a RawFile::readNullTerminatedString() method, as this logic was needed for reading the id666 tags, and I figure it might be generally useful.

I still wonder if it would be cleaner to make tagFromXFile() static methods of the PSF/SPCFile classes.

Unrelated: I've updated the Akao note velocity from 100 to 127. 

## How Has This Been Tested?
Verified the PSF tags are being set. Tested on SPC files with regular id666 tags and SPC files with extended id666 tags of all three types and verified it was iterating over subchunks correctly.

## Types of changes

- [x] Code cleanup

## Checklist:

### DIFF
--- a/src/main/components/PSFFile.cpp
+++ b/src/main/components/PSFFile.cpp
@@ -12,7 +12,7 @@
 constexpr auto PSF_TAG_SIG = "[TAG]";
 constexpr auto PSF_TAG_SIG_LEN = 5;
 
-VGMTag tagFromPSFFile(const PSFFile& psf) {
+VGMTag PSFFile::tagFromPSFFile(const PSFFile& psf) {
   auto& psfTags = psf.tags();
   return VGMTag(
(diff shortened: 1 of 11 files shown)


sykhro (approved):
LGTM

sykhro:
> I still wonder if it would be cleaner to make tagFromXFile() static methods of the PSF/SPCFile classes.

not a bad idea given that we have no namespaces

PR #196, opened 2021-03-02
PS1 AKAO Format Overall Improvements

loveemu:
This PR replaces most of the current AkaoFormat code to support more driver versions accurately.

The updated code still does not emulate all driver behavior but is expected to parse real-world music sequences correctly.

### AKAO Format

* Add support for early AKAO sequences and instruments of Parasite Eve and before
* Change to use MIDI bank for custom instruments and drum kits for code maintainability
* Change matcher to FilegroupMatcher (previously GetIdMatcher was used)
* Fix the incorrect tuning of instruments whose fine-tuning value exceeds a semitone (Chrono Cross: Dragon God) (#124)
* Fix incorrect loop points (#125)
* Fix the bug where basic single-zone instruments are not saved when exporting SF2, unlike DLS (#126)
* Fix crash when loading sequences that do not use any custom instruments and drum kits (Chrono Cross: Dream's Creation) (#168)
* Provide more detailed and accurate information on file events

### Core

* Fix memory leaks on Save as Wav
* Set default velocity value to VGMSeq
* Allow VGMInstrSet without instruments to be accepted (a painful way to support the old AKAO)
* Add an experimental option variable to convert linear pan amplitude to GM2 sin/cos scale (This option is not used in the end, as it only changes the volume slightly, and if a song use pan and volume slides at the same time, the volume adjustment output will be indeterminate.)

## Motivation and Context

I wanted to add Final Fantasy 7 AKAO support to VGMTrans to celebrate the FF7 Remake.

When I came up with that plan, I was not familiar with the PS1 AKAO format and felt that the existing implementation of VGMTrans was somewhat inaccurate at some points. As I went through the analysis of Final Fantasy 7, I faced the need to analyze every game that used the same AKAO driver and spent months analyzing different aspects of the driver. This PR is the culmination of that.

For your information, the technical information I found through my research has been fed back to the following documents.

* [サウンドデータ (AKAO Format) - sagafrontier @ うぃき](https://w.atwiki.jp/sagafrontier/pages/43.html) (Japanese)
* [FF7/PSX/PSX Sound - Final Fantasy Inside](https://wiki.ffrtt.ru/index.php?title=FF7/PSX/PSX_Sound)

## How Has This Been Tested?

I have changed some of the shared code, but nothing should change in formats other than AKAO. I picked up and tested some songs from each of these games.

* Final Fantasy 7 (1997)
[...]

PR #532, opened 2024-08-05
Allow instrument sets with no defined instruments

mikelow:
Some Akao sets have no instruments defined, but instead only use "articulation" based program changes. Those instruments are generated on the fly via AkaoColl::preSynthFileCreation and deleted afterward via AkaoColl::postSynthFileCreation.

VGMInstrSet was deleting all instruments sets that have no instruments defined post-load. Prior to the major refactor, we got around this with a flag. I have instead just removed the condition altogether.

I've also updated the hack that loads FF7 psfs, though I'm not sure how to properly integrate the second sample collection for the final boss music. Its inclusion was causing problems, so I commented it out.

Finally, I updated the README to reflect changes  in supported formats.

## Motivation and Context
Fix the rest of the Akao games.

## How Has This Been Tested?
Aside from relevant Akao games, I ran through a test suite of all supported formats. There was no apparent regression in detected files.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,14 +12,15 @@ VGMTrans converts a music files used in console video games into standard midi a
 - Sony's PS2 sequence and instrument formats (.bq, .hd, .bd)
 - Squaresoft's PS2 sequence and instrument formats (.bgm, .wd)
 - Nintendo's Nintendo DS sequence and instrument formats (SDAT)
-- Late versions of Squaresoft's PS1 format known as AKAO - sequences and instruments
+- Squaresoft's PS1 format known as AKAO - sequences and instruments
 - Sony's PS1 sequence and instrument formats (.seq, .vab)
 - Heartbeat's PS1 sequence format used in PS1 Dragon Quest games (.seqq)
 - Tamsoft's PS1 sequence and instrument formats (.tsq, .tvb)
-- Capcom's QSound sequence and instrument formats used in CPS1/CPS2/CPS3 arcade games
+- Capcom's sequence and sampled instrument formats used in CPS1/CPS2/CPS3 arcade games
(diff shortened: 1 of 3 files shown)


sykhro (approved):
thank you for the fix! what is the reason for changing the offsets for FF7? iirc @loveemu mentioned bad FF7 dumps existing and the new akao code had issues with those

mikelow:
> thank you for the fix! what is the reason for changing the offsets for FF7? iirc @loveemu mentioned bad FF7 dumps existing and the new akao code had issues with those

It hadn't occurred to me that there might be different dumps. I just assumed the reason is that the new PSFLoader is ignoring the offset at which to overlay a psflib because it treats the psflib as a separate file and doesn't write it over the exe portion. But that's just an assumption, I didn't even double check that it's ignoring the offset.

sykhro:
> > thank you for the fix! what is the reason for changing the offsets for FF7? iirc @loveemu mentioned bad FF7 dumps existing and the new akao code had issues with those
> 
> It hadn't occurred to me that there might be different dumps. I just assumed the reason is that the new PSFLoader is ignoring the offset at which to overlay a psflib because it treats the psflib as a separate file and doesn't write it over the exe portion. But that's just an assumption, I didn't even check that it's ignoring the offset.

See `Limitations`, looks like the only problem was missing tags? https://github.com/vgmtrans/vgmtrans/pull/196

mikelow:
> See Limitations, looks like the only problem was missing tags? https://github.com/vgmtrans/vgmtrans/pull/196

The missing tags was one of the problems. Another was that we removed the flag for not deleting instrument-less sample collections. The third is the change in address. Thinking more on it, I'm not sure my assumption about psflib offsets makes sense, given how minor the offset change was. I should probably look into it. Hopefully not an issue of different psf sets using different addresses.

PR #537, opened 2024-08-13
Improve Windows binary packaging / include version in filename

mikelow:
- use 'vgmtrans' target
- extract version from CMake scripts
- rename bin directory to reflect version
- remove unused build directories, like 'Debug', 'Release', and 'RelWithDebInfo'

*Update*: 
- updated the macOS job to include the version in the dmg filename.
- updated the version to 1.2 in the root CMakeLists.txt file

This also temporarily removes the arm64 build, as it appears to be broken. Its output has been a sub 1mb zip file, and it is generating an error when we set the target to 'vgmtrans'.

## Description
This updates the build.yml workflow for creating a Windows binary artifact. Right now, the zip file contains a "bin" directory, and the bin directory contains empty directories like "Debug" and "Release". This cleans up the binary zip so that its contents are consistent with the 1.1 preview release.

## Motivation and Context
Clean up the Windows binary before a new release.

## How Has This Been Tested?
Currently testing by downloading the resulting zip artifact.

## Types of changes

## Checklist:


sykhro:
> This also temporarily removes the arm64 build, as it appears to be broken.

how did I not notice D: i tested this locally with the arm64 build tools and it worked, must be a CI issue

mikelow:
I have now added the version to the macOS dmg filename (eg: VGMTrans-v1.2.dmg). I have also updated the version to 1.2 in the root CMakeLists.txt file.

sykhro (commented):
I think this is a fair change, I was thinking that the has in the filename would be useful too (people download from CI a lot), but we do log that in the program...

mikelow:
> I was thinking that the has in the filename would be useful too

Typo? Could you repeat?

sykhro:
> > I was thinking that the has in the filename would be useful too
> 
> Typo? Could you repeat?

SHA*, commit hash

mikelow:
> SHA*, commit hash

Yeah, this makes sense. I think for the actual releases we would exclude the SHA, but we can just manually edit those filenames before uploading.

mikelow:
Ok, I've updated build.yml to append the SHA to the zip file.

Now I'm questioning the value of adding the version to the filename. 🤷‍♂️. I initially thought these artifacts would be selectable when we draft a release, and therefore it would be helpful to add the version to make it release ready. Now I see that's not the case.

What do you think, shall I remove it?

mikelow:
I went ahead and removed the version from the filenames but kept the SHA.

sykhro (approved):

sykhro:
> Ok, I've updated build.yml to append the SHA to the zip file.
> 
> Now I'm questioning the value of adding the version to the filename. 🤷‍♂️. I initially thought these artifacts would be selectable when we draft a release, and therefore it would be helpful to add the version to make it release ready. Now I see that's not the case.
> 
> What do you think, shall I remove it?

CMake already picks the version from the Git tag, so, in an ideal scenario, we have automation that triggers once you create the github release and then uploads the build to it. but I don't particularly like that because the moment you create a release there won't be binaries until CI is done running. we also don't have much in place to automate release notes (unless we start to always squash-merge and rely on PR titles)

PR #540, opened 2024-08-19
Update macOS dmg

mikelow:
This updates the dmg background image and README file. Most importantly, I have included instructions to ctrl-click the script that removes the quarantine attribute and self-signs the app, since macOS may otherwise totally block execution.

## Description
- update the dmg background image. It now has a music notation motif. The tif file bundles both a 1x and 2x image for retina displays.
- update the dmg README file. The instructions are numbered now and hopefully more straight-forward. The copy was slightly tweaked for clarity.
- rename README_MACOS.txt to README.txt. This file is only included in the dmg, so there's no conflict in naming.

I should have said this earlier: let me know if you're uncomfortable with claiming Apple's $99 requirement is monopolistic.

Also, please feel free to propose a different background image or tweaks to this one.

## Screenshot:

Screenshot 2024-08-18 at 10 33 56 PM

Second iteration:

Screenshot 2024-08-19 at 12 28 41 PM


sykhro:
> I should have said this earlier: let me know if you're uncomfortable with claiming Apple's $99 requirement is monopolistic.

No objections :)

> Also, please feel free to propose a different background image or tweaks to this one.

I like this style, some thoughts:
- the background has a green-ish tint to it that we have nowhere else in the program (reminds me of Finale Notepad :p)
- I'm not sure if people will understand the musical iconography correctly, especially that I needs to be done before II

mikelow:
> * the background has a green-ish tint to it that we have nowhere else in the program (reminds me of Finale Notepad :p)

Yeah, I'm just trying to find a pleasant off-white color to provide contrast to the white VGMTrans icon. I can try something more gray or sepia. Alternatively, I'm in favor of changing the app icon, but that would require a lot more deliberation.

> * I'm not sure if people will understand the musical iconography correctly, especially that I needs to be done before II

I can think of two dumb ways to solve this:

1. We write out "Step I" and "Step II" instead of "I" and "II".
2. We expand the height and put instructions at the top.

Both of these hurt the visual elegance, but that's ok.

Also, worth bearing in mind that if the user runs the script before moving the app into Applications, the script will tell them to do that first. Secondly, the README provides immediate clear instructions.

mikelow:
Updated. I changed the roman numerals to "Step I" etc, expanded the width, and set the background to a sepia / paper color. Feel free to provide more feedback.

sykhro (approved):
I find this clear enough now, thank you for the changes. If there's one thing I've learned from our issue tracker is that nothing is ever obvious to users :p

mikelow:
> If there's one thing I've learned from our issue tracker is that nothing is ever obvious to users :p

No doubt!

PR #528, opened 2024-08-02
AkaoMatcher onFinishedScan()

mikelow:
This PR is branched off of `akao-matcher` and rebased on top of `scanner-format`. It's probably best to ignore it until those PRs are addressed.

This follows up on #527 and the discussion in #526 to add an onFinishedScan() callback to AkaoMatcher. This vastly improves the efficacy of AkaoMatcher.

## How Has This Been Tested?
This is working with every PSF set I have tested, and is able to match files within ISO images for games that use a single sample collection per song with ID associations.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.

PR #542, opened 2024-08-19
Refactor FileGroupMatcher

mikelow:
This should vastly improve the reliability of FileGroupMatcher.

- add Matcher::onFinishedScan as virtual method
- Root now calls onFinishedScan after each scan()
- refactor FileGroupMatcher - it now utilizes onFinishedScan() and does not retain state between RawFile loads
- remove obviated code in CLIVGMRoot

## Description
This is an extension of the work I did on AkaoMatcher (#527 and #528), specifically around adding a Format to the Scanner class so it can invoke an `onFinishedScan()` callback on Matcher when a RawFile has finished loading. `onFinishedScan()` is now a virtual method of Matcher, and it is invoked in VGMRoot::setupNewRawFile() after scanning a file.

I've updated the logic of FileGroupMatcher so that it works in the simple way I had originally envisioned - it assumes an association between detected files in a single RawFile. It does not retain state from previous RawFile loads. It is consequently much more reliable and should be less confusing to users.

One thing I removed is the ability for FileGroupMatcher to create sequence-less collections. I think this will confuse users more than it will add value, and it violates how I think of collections as a concept.

## How Has This Been Tested?
I ran a test suite of all the supported formats and saw no regression. I also verified the CLI is working with these changes.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/components/Matcher.cpp
+++ b/src/main/components/Matcher.cpp
@@ -6,6 +6,7 @@
 
 #include "Matcher.h"
 #include "Root.h"
+#include <regex>
 
 using namespace std;
 
@@ -36,31 +37,38 @@ bool Matcher::onCloseFile(std::variant<VGMSeq *, VGMInstrSet *, VGMSampColl *, V
   return false;
 }
 
-bool Matcher::makeCollectionsForFile(VGMFile* /*file*/) {
-  return false;
-}
-
 // ****************
 // FilegroupMatcher
 // ****************
 
 FilegroupMatcher::FilegroupMatcher(Format *format) : Matcher(format) {}
 
+
+void FilegroupMatcher::onFinishedScan(RawFile* rawfile) {
+  // xsflib files are loaded recursively with xsf files. We want to scan the xsflib and the xsf
+  // files together as if they're one file, so ignore this callback and wait for the xsf to finish.
+  if (std::regex_match(rawfile->extension(), std::regex(R"(\w*sf\w?lib$)")))
+    return;
+
+  lookForMatch();
+
+  seqs.clear();
+  instrsets.clear();
+  sampcolls.clear();
+}
+
 bool FilegroupMatcher::onNewSeq(VGMSeq *seq) {
   seqs.push_back(seq);
-  lookForMatch();
   return true;
 }
 
 bool FilegroupMatcher::onNewInstrSet(VGMInstrSet *instrset) {
   instrsets.push_back(instrset);
-  lookForMatch();
   return true;
 }
 
 bool FilegroupMatcher::onNewSampColl(VGMSampColl *sampcoll) {
   sampcolls.push_back(sampcoll);
-  lookForMatch();
   return true;
 }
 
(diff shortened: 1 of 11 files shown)


src/main/components/Matcher.cpp
```diff
@@ -36,31 +37,38 @@ bool Matcher::onCloseFile(std::variant<VGMSeq *, VGMInstrSet *, VGMSampColl *, V
...
 // FilegroupMatcher
 // ****************
 
 FilegroupMatcher::FilegroupMatcher(Format *format) : Matcher(format) {}
 
+
+void FilegroupMatcher::onFinishedScan(RawFile* rawfile) {
+  // xsflib files are loaded recursively with xsf files. We want to scan the xsflib and the xsf
+  // files together as if they're one file, so ignore this callback and wait for the xsf to finish.
+  if (std::regex_match(rawfile->extension(), std::regex(R"(\w*sf\w?lib$)")))
```
sykhro, line 50: `std::regex` and friends are _terribly_ slow.
This probably won't have the biggest impact since it's not called all the time, but if we need to match regular expressions then there's nothing better than [CTRE](https://github.com/hanickadot/compile-time-regular-expressions)

src/main/components/Matcher.h
```diff
@@ -49,9 +50,9 @@ class Matcher {
...
 class SimpleMatcher : public Matcher {
- protected:
-    // The following functions should return with the id variable containing the retrieved id of the
-    // file. The bool return value is a flag for error: true on success and false on fail.
+protected:
+  // The following functions should return with the id variable containing the retrieved id of the
+  // file. The bool return value is a flag for error: true on success and false on fail.
   virtual bool seqId(VGMSeq *seq, IdType &id) = 0;
   virtual bool instrSetId(VGMInstrSet *instrset, IdType &id) = 0;
   virtual bool sampCollId(VGMSampColl *sampcoll, IdType &id) = 0;
```
sykhro, line 58: Random thought unrelated to the PR: it might be nice to change these to return  `std::optional<IdType>`, getting rid of the out param

sykhro (approved):

src/main/components/Matcher.cpp
mikelow: I am actually concerned about the performance ramifications of this regex line. Right now, we're running about 33 scanners, and a large chunk of those are using FilegroupMatcher.

I'm thinking we should add this library. What do you think?

src/main/components/Matcher.h
mikelow: Solid suggestion. I'll make a note of it.

src/main/components/Matcher.cpp
sykhro: > we're running about 33 scanners, and a large chunk of those are using FilegroupMatcher

oh welp, didn't think it was that many.
in that case, yes, let's do it. CTRE is built on templates, the regex is evaluated as a constexpr. compiler optimizations go very aggressive and the end result is basically an automata in machine code

PR #583, opened 2025-04-19
Fix SSEQ file size read

MicaWE:
This change adjusts the size read for the actual file interactions to equal the size in the header. Previously this size was read as a short even though the header contains a word field for size.

Dragon Quest VI - Realms of Reverie contains SND_BGM_M30_LULLABY which has a size of 0x1a228 and was previously not parsable.


sykhro:
Good catch, thank you!!

MicaWE:
Seems like auto-merge is being prevented due to deprecation

PR #588, opened 2025-05-06
Early work on Konami Mystic Warrior format

mikelow:
Early support for the format used in Konami Arcade games with [Mystic Warrior-based hardware](https://www.system16.com/hardware.php?id=573).  This format appears to share a lot of logic with games on earlier hardware (using the YM2151), like TMNT2.

I've been sitting on this for awhile. While the driver is far from complete, it may as well be merged in. A lot of the basics are here.

## How Has This Been Tested?
Every game loads without issue. Haven't run into crashes during load or playback.

## Types of changes

## Checklist:


src/main/formats/KonamiArcade/KonamiArcadeDefinitions.h
```diff
@@ -0,0 +1,30 @@
...
+constexpr int K054539_CLOCK_RATE = 18432000;
+#define NMI_TIMER_HERZ(data, skipCount) (((38 + (data)) * (K054539_CLOCK_RATE / 384.0f / 14400.0f)) / ((skipCount)+1))
+
+// The volume and pan table logic is based on the MAME k054539 emulation source code by
+// Olivier Galibert. That source code is distributed under the BSD-3-Clause license.
+
+constexpr std::array<double, 256> generate_volume_table() {
+  std::array<double, 256> volTable = {};
+  for (int i = 0; i < 256; ++i) {
+    // volTable[i] = constexpr_pow(10.0, (-36.0 * static_cast<double>(i) / 64.0) / 20.0) / 4.0;
```
sykhro, line 15: Dead code? What's the / 4 for?
If we need to remember something, it would be better to have a comment explaining rather than a magic formula

src/main/util/constexprHelpers.h
sykhro, line 1: You might want to make these `consteval`

src/main/util/BytePattern.h
```diff
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include "common.h"
```
sykhro, line 6: How necessary is this? We have quite a lot of transitive headers in the codebase

sykhro (approved):
Left a couple minor comments. Thanks!

src/main/util/constexprHelpers.h
mikelow: I've updated the the functions entirely. They're now using consteval and renamed pow() and sqrt() but within a namespace `ct` (constant time). My thanks to ChattyG. Apparently C++23 makes the built-in functions constexpr, which is dope. Too bad we're on C++20.

src/main/formats/KonamiArcade/KonamiArcadeDefinitions.h
mikelow: Updated with a comment. The MAME [implementation](https://github.com/mamedev/mame/blob/7ba2d22d474506f5ba5998fa3bd5e0d8fadf3f5e/src/devices/sound/k054539.cpp#L521) divides by 4, but I found the volume was far too quiet this way.

src/main/util/BytePattern.h
mikelow: Apparently not necessary at all. Not sure why I added it.

PR #591, opened 2025-05-09
FilegroupMatcher and Vab improvements

mikelow:
This improve FilegroupMatcher and updates Vab to take advantage of the changes.

## Description

FilegroupMatcher works fine for files that contain a single sequence, instrument set, and sample collection. It doesn't do a good job when there are more than one of a given file type.

This attempts to improve how we handle these cases. Part of this is the addition of a new VGMInstrSet::isViableSampCollMatch() virtual function. It returns true by default, but a subclass can override it to help filter out sample collections for matching. In the case of VAB, we now keep track of the list of VAG sizes (inspired PS1 speak for "sample"), which allows us to compare the size of a prospective sample collection with what the instrument set expects.

This doesn't fix matching entirely for PS1 games: when there are multiple instrument set / sample collection pairs, there's still no way to know which belongs to a given sequence. But it does fix a lot of cases.

The new algorithm is summed up in the code comment I added to FilegroupMatcher.h:

> FilegroupMatcher handles formats where the primary method of associating sequences, instrument sets, and sample collections is that they are loaded together within the same RawFile. When loading is complete, FilegroupMatcher processes the VGMFiles in the order they were added and groups them into collections: thus, it assumes association by order. FilegroupMatcher does not retain state between loads, so it will never create a collection that spans multiple RawFiles, with the exception that FilegroupMatcher processes a psf file and all of its psflib dependencies together as a single load.
> 
> When loading is complete, FilegroupMatcher:
> 
> 1. Sorts sequences, instrument sets, and sample collections by their offset, ascending.
> 2. Attempts to pair each instrument set with a sample collection. It calls VGMInstrSet::isViableSampCollMatch() to filter out non-viable sample collections. If the file contains a single instrument set and sample collection, they are always paired. After pairing, a sample collection is removed from the pool for matching, unless it is the last one. 
> 3. Creates a new collection for each sequence by iterating over the instrument set / sample collection pairs. If that list is exhausted before the sequence list, it reuses the last instrument set / sample collection pair.

This PR also changes the logic for loading VAB files with bundled sample collections. The prior code assumed that if a VAB was discovered at offset 0, it would be bundled with a sample collection. This wasn't necessarily true, and there are counter examples in PSF collections. The older code also assumed an older version of the bundled VAB format. In newer revisions (I believe it's tied to the format), the attached sample collection is preceded by a 32 bit size value. Since I found more instances of this bundled format, and since we can validate by comparing the total VAG sizes against the size value, I've removed the old logic and added support for this bundled format. The old format should still be detected via the searchForPSXADPCM() function. Really, the only time this bundle load logic makes a difference is when a sample collection has been partially zeroed out via PSF optimization.

## Motivation and Context
Get PS1 SEQ/VAB loading working more reliably.

## How Has This Been Tested?
A lot of formats use this Matcher, so I am going to to run a test suite of every supported format.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/components/instr/VGMInstrSet.h
+++ b/src/main/components/instr/VGMInstrSet.h
@@ -25,6 +25,7 @@ public:
   virtual bool parseHeader();
   virtual bool parseInstrPointers();
   virtual bool loadInstrs();
+  virtual bool isViableSampCollMatch(VGMSampColl*) { return true; }
 
   VGMInstr *addInstr(uint32_t offset, uint32_t length, uint32_t bank, uint32_t instrNum,
                      const std::string &instrName = "");
--- a/src/main/components/matcher/FilegroupMatcher.cpp
+++ b/src/main/components/matcher/FilegroupMatcher.cpp
@@ -76,8 +76,8 @@ void FilegroupMatcher::lookForMatch() {
 
   // An instrument set plus an optional extra sample collection to add to the VGMColl
   struct InstrAssoc {
-    VGMInstrSet* instr;
-    VGMSampColl* extraSamp;   // nullptr when no extra sample collection is needed
+    VGMInstrSet* instrSet;
+    VGMSampColl* sampColl;
   };
 
   std::vector<InstrAssoc> instrAssocs;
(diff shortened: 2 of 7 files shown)


mikelow:
Thank you for testing and staring at the code! 😀

PR #594, opened 2025-05-11
move LICENSE to root directory

mikelow:
This moves our LICENSE to the root from LICENSE/LICENSE so that github hopefully detects it. I renamed the LICENSE directory to "licenses".


mikelow:
Going to merge without review seeing as this is a straight forward change.

mikelow:
Yeah, I noticed that. If I could go back in time I’d go with the MIT license, which is practically equivalent and more widely known. Oh well.

PR #600, opened 2025-05-14
Remove VGMItem::Icon. Use VGMItem::Type to determine icon and color.

mikelow:
This expands upon #599.

This PR removes VGMItem::Icon completely. Icon is now determined by an item's `Type` using `iconForItemType()`. Color is still determined using `colorForItemType()` per #599.

I combed through every sequence format and fixed every generic sequence event which specified an icon and color. Along the way, I updated a lot of these classifications to use more specific types. For example, we now have types for volume slide, pitch slide, and pan slide.

Despite the expansion of item types, this PR doesn't add any new icons, but it does make it trivial for us to add new icons with immediate effect.

Beyond this PR I would like to:
1) Add programmatic icon tinting so that icon color is consistent with HexView. I'm already working on this.
2) Rework the actual Item colors.
3) Add more icons

## Motivation and Context
Three goals here:
1) Change the paradigm for item visualization so that icon and color are derived from an item's type rather than flexibly assigned.
2) Enforce consistency - each Type will always use the same icon and color
3) Simplify generic event creation, though we should strive to reduce it.

This will also make it easier to implement icon tinting.
## How Has This Been Tested?
I've taken a look at the results among many formats, and things look fine to me. There will be differences in icon assignment and color. The changes to icon assignment are deliberate, though I'm sure it's not perfect.

## Types of changes

- [x] Refactor

## Checklist:


mikelow:
> I really like this direction. The less UI details in vgmtranscore, the better. Can't really find anything to note here, but looking at the code did remind me of all of those suspicious str().c_str(). Wondering if we have any memory unsafety dragons here

Awesome, I figured you'd agree with the change. Removing UI details from vgmtranscore is a better phrasing of my first goal.

Yeah, there be dragons in the old string code. Worth another sweep in the future.

PR #602, opened 2025-05-14
rewrite of PSXSPU::decompVAGBlk(). fixes potential clipping

mikelow:
This rewrites the PSX ADPCM sample decompression algorithm with a lot of help from ChatGPT o3. I found that the previous version was potentially clipping samples (ex: FFIX - Crossing Those Hills sample 18, heard obviously in sample and faintly at end of song). This clipping is inconsistent and depends on the compiler used. 

It looks like the `decompVAGBlk()` function was doing some bit manipulation / casting that caused implementation-defined behavior. Also, probably most critically, it wasn't clamping the value to the signed 16 bit range.

This new implementation is int instead of float based and should be more reliable and faster. I dug into the logic and compared against another [int based implementation](https://github.com/vgmstream/vgmstream/blob/1ef807b594be26c57c4d004bec23f5b296d49992/src/coding/psx_decoder.c#L162). I also asked ChatGPT to compare it against several others implementations. I am confident the logic is sound.

## How Has This Been Tested?
Clipping is gone in the problem sample. Tested other conversions with no issues.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/formats/common/PSXSPU.cpp
+++ b/src/main/formats/common/PSXSPU.cpp
@@ -404,16 +404,15 @@ double PSXSamp::compressionRatio() {
 }
 
 void PSXSamp::convertToStdWave(uint8_t *buf) {
-  int16_t *uncompBuf = reinterpret_cast<int16_t*>(buf);
+  s16 *uncompBuf = reinterpret_cast<s16*>(buf);
   VAGBlk theBlock;
-  f32 prev1 = 0;
-  f32 prev2 = 0;
+  s32  prev[2] = {0, 0};
 
   if (this->bSetLoopOnConversion)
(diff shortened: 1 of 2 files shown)


mikelow:
> This int-based version should be closer to what real hardware does, I have no way of testing it. There's a lot of wiggle room with interpolation results. But at least this should return the same result on all platforms (the sign-extension of the old implementation looks suspicious)

Yeah, the bit shifting with signed values to the point they overflow was a concern. Very interesting that I saw discrepancies between the 1.2 release build and a local clang-19 build on macOS, even when I checked out the exact 1.2 commit (perhaps we should start using tags).

Interpolation isn't part of the conversion, fortunately, since it decompresses to the very standard 16 bit signed PCM.

PR #603, opened 2025-05-15
Icon tinting

mikelow:
This adds a new QIconEngine subclass TintableSvgIconEngine for programmatically tinting icons. Icons in the file tree view and status bar and now tinted consistently with the colors used in the hex view.

Each QIcon statically declared in `iconForItemType()` is now constructed by passing a TintableSvgIconEngine. TintableSvgIconEngine paints a tinted version of the icon on render and caches it using the dimensions as the key.

I've also renamed qt/resources/images directory to "icons"

## Motivation and Context
Icon colors were random and inconsistent with Hex View. This hopefully adds some visual clarity.

## How Has This Been Tested?
Seems to be working 🤷‍♂️. Performance hit seems acceptable, not even sure if its perceptible.

image

## Types of changes

## Checklist:


mikelow:
> I will think about this one for a second, the most obvious implementation for this uses a QStyledItemDelegate. But that can't be reused for the status bar

If you think there's a better approach, I'm open to it. I'm leaning toward not caring if the effect doesn't work on the status bar icon.

mikelow:
Looking into this further, I'm having trouble seeing the benefit of using `QStyledItemDelegate` over utilizing `QIconEngine` to custom render the icons. I'd like to hear your thoughts.

If my understanding is correct: 

- Either way we'll want to cache the icon pixmaps. The item delegate would necessarily use a global cache like QPixmapCache, though we can easily adapt the QIconEngine implementation to use it. Only benefit of global cache is ability to easily clear it for all icons, say for themes.
- No performance difference.
- Item delegate approach seems to require more code, as we need to override the delegate's `paint()` to render our icon (or pull it from cache) then draw the icon, then call the parent delegate's paint() method to draw everything else.
  * Complications: I guess we'll need to temporarily remove the DecorationRole lest the parent class tries to draw an icon? Won't it draw the text in the wrong place without an icon? So then we need to also X translate the QPainter before calling the parent class `paint()` method? 
  * Aren't we going to make assumptions about how the default implementation draws each Row? We're assuming the the position of the icon. Does our code break when a future Qt version changes appearance?
  * We could of course draw the text ourselves, but then we've taken on more responsibility.
- Delegate approach only works with Widgets employing `QStyledItemDelegate`. Will it even be general across all of them or will the paint() function have to differ in terms of where the icon goes, or its size?


The QIconEngine approach seems simpler and more intuitive to me. But I'd love to hear the case why I'm wrong and a big stupid head.

PR #629, opened 2025-07-21
Volume logic refactoring

mikelow:
I noticed that my last PR broke volume in CPS2TrackV2 due to the amplitude scale conversion being applied twice. In the course of fixing this, I've tried to remove repetitive code in SeqTrack which handles volume conversion (for Volume, Expression, and Master Volume events).

I also realized that it's silly that we have no methods for adding volume events that take percent values. Right now, we receive a 7-bit integer value which we then convert to a percent, then (usually) convert the scale, then convert back to 7-bit MIDI values. We should have the option to pass volume as a percent in the first step, so I've added overloaded `addVol()` and `addExpression()`. In time, I'd like to do the same for the `insert` and `NoItem` variant methods, and also note velocity. It's more intuitive except in cases where a driver uses a 7-bit value natively. Relatedly, in a future update, I'd like to make the linear amplitude scale option true by default, since it's the norm, not the exception.

I've centralized the logic for adding volume events into a new method `addLevelNoItem()`. This method takes a percent volume value, the controller to add events for (volume, expression, master volume), and the resolution (7-bit or 14-bit). It can also handle event insertion with an optional absTime param.

### Changes

SeqEvent:
  - update VolSeqEvent and ExpressionSeqEvent to alternatively accept a percent level value, add related constructors

SeqTrack:
  - add applyLevelCorrection() which takes a percent level value and applies pan correction and/or linear amplitude scale correction
  - addLevelNoItem() which handles adding or inserting volume, expression and master volume events of 7 or 14 bit resolution, add/insert Vol/Level/MastVol events to utilize it
  - add alternative addVol and addExpression methods which take percent level values rather than 7 bit ints
  - add insertExpressionNoItem()
  - update vol, expression, and mastVol to be u16 to allow for 14 bit values

ScaleConversion:
  - update convertPercentAmpToStd14BitMidiVal() to use the a full 14bit range as input instead of 127*127

MidiFile:
   - add ExpressionFineEvent class
   - addinsertVolumeFine(), addExpressionFine(), insertExpressionFine()
   - add MasterVolEvent constructor which takes an lsb value, and pass lsb into Sysex data.
   - update addMasterVol and insertMasterVol to accept an lsb byte

CPS2Seq:
  - update to use linear amplitude scale

## How Has This Been Tested?
I've tested that volume slides are still working as intended. I've tested by ear for breakage of various formats.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/components/seq/SeqEvent.cpp
+++ b/src/main/components/seq/SeqEvent.cpp
@@ -87,6 +87,9 @@ SetOctaveSeqEvent::SetOctaveSeqEvent(SeqTrack *pTrack,
 VolSeqEvent::VolSeqEvent(SeqTrack *pTrack, uint8_t volume, uint32_t offset, uint32_t length, const std::string &name)
     : SeqEvent(pTrack, offset, length, name, Type::Volume), vol(volume) { }
 
+VolSeqEvent::VolSeqEvent(SeqTrack *pTrack, double volume, uint32_t offset, uint32_t length, const std::string &name)
+    : SeqEvent(pTrack, offset, length, name, Type::Volume), percentVol(volume) { }
+
 // ***********
 // Volume14BitSeqEvent
 // ***********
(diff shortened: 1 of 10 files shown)


mikelow:
Making this a draft as I just discovered other formats that don't use linear amplitude scaling need fixing.

mikelow:
False alarm. Seems fine.

PR #650, opened 2025-08-26
Fix mp2k sample unsigning

mikelow:
This probably broke as a result of my changes to VGMSamp.

I was unable to test BDPCM conversion because I couldn't find an example that uses it.

## Types of changes

## Checklist:


mikelow:
> iirc we do the same thing for NDS, could you confirm if that works or needs the same fix?

Good call. Totally forgot. Confirmed it was broken and now fixed.

PR #653, opened 2025-08-26
Minimum sequence size option

mikelow:
This adds a menu option to filter out sequences beneath a size threshold. This is useful for Sega Saturn and CPS titles, where sequences are often loaded via tables that contain sound effect sequences. It's also generally useful for *sf formats, which sometimes contain leftover sfx sequences in ram.

I'm still trying to figure out how best to frame these options in the UI. Do we have separate menus for conversion options vs scanning / loading options? The proposed change in this PR is a new "Detection" menu. I guess we can iterate on it as more options are added.

The logic checks the size of sequences after loading completes, but before they are added to the UI via VGMRoot calls.

I would like to add a toast notification indicating how many files were skipped, but this will require a bit more work.

## How Has This Been Tested?
Tested on CPS sets and psf files with extraneous minor sequences.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/Options.h
+++ b/src/main/Options.h
@@ -54,6 +54,9 @@ public:
   bool skipChannel10() const { return m_skip_channel_10; }
   void setSkipChannel10(bool should) { m_skip_channel_10 = should; }
 
+  int minSequenceSize() const { return m_min_sequence_size; }
+  void setMinSequenceSize(int size) { m_min_sequence_size = size; }
+
   void load(OptionStore& store) {
     auto g = store.beginGroup("ConversionOptions");
     const int bs = store.getInt("bankSelectStyle", static_cast<int>(BankSelectStyle::GS));
@@ -61,6 +64,7 @@ public:
                                                                 : BankSelectStyle::GS;
     m_sequence_loops = store.getInt("sequenceLoops", 1);
     m_skip_channel_10 = store.getBool("skipChannel10", true);
+    m_min_sequence_size = store.getInt("minSequenceSize", 0);
   }
 
   void save(OptionStore& store) const {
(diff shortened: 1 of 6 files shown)


mikelow:
> I wonder, does it make sense to actually skip scanning at all?
> It might be easier to add a filter to the UI itself (and have it set to a sensible default)

I hadn't considered that. I'm trying to think how that would best work in th UI. I was thinking an upside of the original filter concept is that it could improving matching, especially if we allow filtering on sample collection size.

mikelow:
> That's true. Maybe we need a "scan options" dialog to show up after a file is dropped on the program?

I agree that a dialog would be appropriate. I was thinking we show one before scanning files above a certain size. It would let users select which formats to scan for.

> But then it's still hard to add something to the UI that shows the search has dropped items of a certain size.

Maybe I'm not following, but just surfacing a toast after the scan would do it, right? "X sequences were excluded because they were below the minimum file size". It would require some rewiring to track that information, of course.

> For my suggestion, I was thinking of a filter section to apply to the collections/vgmfiles lists. Unsure where exactly (nor how)

I can envision a filter UI specific to each docking window. But how do you make an out-of-the-way filter UI that works across the Detected Files and Collections windows that is also intuitive? Maybe you make a filter UI only for Detected Files, and it updates the list of collections?

I think I prefer having a filter be a scan option because:
1) It's intuitive
2) It hardly complicates our UI and doesn't take up screen space.
3) The filter can improve the results of fuzzy Matchers
4) It's way easier to build (is built).
5) Users will see it more as something to set and leave alone, whereas a filter for a list UI is perceived as a temporary search/organizational tool.

The danger: users forgetting the setting and getting confused why files aren't being found. This applies to both approaches. Toasts can solve this.

At any rate, I don't feel super strongly about this future. I am finding it moderately useful as I play with it.

Issue #655, opened 2025-08-26
MAME ROMs can no longer be loaded?


fotisandstuff:
### Environment details

Windows 10 x64

### Describe the bug

Hello! I tested the latest vgmtrans build (5170c6d) and tried to load The Punisher (punisher.zip, from a MAME 0.279 romset). It cannot load it. I tried an earlier vgmtrans build (7cf66dd) and it can load it and extract sounds without issues.

### Steps to reproduce

1. Load the game
2. ...

### Expected behaviour

VGMTrans should be able to load The Punisher ROM.

### Logs

<details><summary>Click to expand log</summary>

```
REPLACE THIS LINE WITH YOUR LOG
```

</details>

fotisandstuff:
> does the log show anything?

No, it doesn't. I tested again with version e2594b6 and The Punisher from the MAME 0.280 romset. Still nothing.

felineki:
I am also encountering this problem on Windows 11. Nothing in the log. Program simply does not display any of the scanned data after attempting to load the ROM file.

LF222:
Same issue here, wanted to test all those new konami listings but none of them will work

mikelow:
I was able to reproduce the issue. It's due to the vendored zlib-ng library not being detected by minizip-ng. A fix is pending.

mikelow:
Fixed with #673 

Thanks to everyone who reported the issue!

PR #671, opened 2025-09-26
Fix zlib-ng detection by minizip-ng

mikelow:
Just putting this here as an optional fix for our issue #655. I'm leaning against this solution and toward your suggestion of reverting to zlib and minizip instead of zlib-ng / minizip-ng.

The problem is that minizip-ng isn't detecting the vendored zlib-ng. I suspect the problem arose when we changed zlib-ng to use ZLIB_COMPAT mode, which causes it to use the same interface as normal zlib. We did that so that libchdr could find the vendored zlib, but just getting that working was hacky.

This seems to work as a solution for keeping zlib-ng / minizip-ng, but I don't like it. @sykhro what are your thoughts?


mikelow:
Sounds good. Well revert to zlib / minizip instead.

PR #673, opened 2025-09-26
Revert back to zlib and minizip from zlib-ng / minizip-ng

mikelow:
This replaces the zlib-ng and minizip-ng git submodules with a zlib submodule tagged at v1.3.1 (the most recent release).

zlib-ng causes some conflict with other dependencies that use zlib - specifically libchdr and libunarr. The resolution of this issue led to minizip-ng not discovering the vendored zlib-ng, though compilation still succeeded. This in turn led to issue #655 where mame roms ceased to load.

While we could [fix](https://github.com/vgmtrans/vgmtrans/pull/671) zlib-ng to work properly with minizip-ng, given that we're not utilizing most of its performance benefits it seems more prudent to revert back to zlib and minizip.

## How Has This Been Tested?
Tested that rom sets and chds are loading successfully macOS Sonoma, Windows 10, and Ubuntu 22.04.

## Types of changes

## Checklist:


mikelow:
👍 What was the reason we couldn't utilize the AVX flags? Maximum CPU compatibility?

mikelow:
Interesting. Had to ask AI why they don't just compile for all AVX/non-AVX implementations and choose the appropriate one at runtime.

mikelow:
Very strange, that option `WITH_RUNTIME_CPU_DETECTION` is ON by default. I don't understand why it was breaking for older CPUs in that case, but sounds like it was.

PR #686, opened 2025-10-24
Add ControlFlowState for loop detection

mikelow:
tldr
_**Problem:**_ some formats aren't looping as specified by the Sequence Loops options.

_**Solution:**_ find the loop point by treating SeqTrack as a state machine: log a state for every event we parse so we can determine when we revisit a previously visited state - that's our loop point.

## In detail
This adds a general solution for detecting the point where a track has played every one of its events and is entering its infinite loop. We need to determine this point so that we can accurately loop the sequence however many times is specified by the Sequence Loops option. Most format code right now often relies on fragile logic, typically assuming that a jump event to an already visited offset is indicative of an infinite loop point. This can easily fail. For example: a jump could move the offset forward and exist within a loop, or it could occur within a call/return that is used multiple times. Any event that moves the curOffset has the potential to be an infinite loop point, but we don't handle all of the possibilities.

We can solve this in a generalized way by treating each track like a finite-state machine. As we execute each event, we enter a new state. When we enter a state we've already visited, we've hit the infinite loop point. State isn't defined as simply the offset of the event, but also includes every variable that affects control flow - for example: information about the call/return stack and every loop we're in.

In practice, the approach is as follows: every time we parse an event, we generate a hash of the state and store it in a table. When we execute an event that moves the current offset (ie a jump event, a loop event, etc) - events that can lead to an infinite loop - we calculate the hash of the state we are to enter. If it has already been visited, we've found the infinite loop point.

To use this detection, I've added a few methods and SeqEvent subclasses for control flow events. Specifically, I've added `SeqTrack::addJump(), addCall, and addReturn()`. I expect to add more control flow methods/events down the road as more formats are integrated. For now, I have integrated this detection into the NDSSeq format as a proof of concept. Detection is enabled via: `VGMSeq::setShouldTrackControlFlowState(bool enable);`

## How Has This Been Tested?
I've tested this against many NDSSeq sets, including those which are especially bug-inducing, like Animal Crossing.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/components/seq/SeqTrack.cpp
+++ b/src/main/components/seq/SeqTrack.cpp
@@ -1,5 +1,5 @@
 /*
- * VGMTrans (c) 2002-2024
+ * VGMTrans (c) 2002-2025
  * Licensed under the zlib license,
  * refer to the included LICENSE.txt file
  */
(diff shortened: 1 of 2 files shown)


Bonboon229:
Edit: DS Formats now loop properly, many PS2 games still don't.

mikelow:
> Edit: DS Formats now loop properly, many PS2 games still don't.

"For now, I have integrated this detection into the NDSSeq format as a proof of concept."

mikelow:
> Oop, missed that part Great work btw

Thanks!

PR #691, opened 2025-11-01
make Mac playback slider behavior universal across platforms

mikelow:
This adds a new subclass of QSlider called `ClickJumpSlider` and uses it for the playback slider in place of a vanilla `QSlider`. Instead of requiring the user to click and drag the handle element to change position (as is required on Windows and Linux), the user can simply click any position on the slider to jump to it. This is the macOS behavior.

Logically, this is pretty basic. We override `QSlider::mousePressEvent()` and handle left button mouse clicks. If a user clicks the handle element, we allow the default behavior to run. Otherwise, we calculate the new value of the slider based on click position.

The calculation is straightforward - we get the click position relative to the width of the bar, but we also factor out the width of the handle. The handle element doesn't extend beyond the widget, so if the user clicks at the beginning of the bar at an x position <= half the handle width, it's the same as clicking position 0. Likewise, if the user clicks the end of the bar up to half the handle width, it's treated as position 1.

## Motivation and Context
I think the macOS slider behavior is much more usable and should be universal.

## How Has This Been Tested?
Tested on macOS, Win10, and Ubuntu 22.04.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/CMakeLists.txt
+++ b/src/ui/qt/CMakeLists.txt
@@ -74,6 +74,8 @@ add_executable(
   util/LambdaEventFilter.h
   util/UIHelpers.cpp
   util/UIHelpers.h
+  widgets/ClickJumpSlider.cpp
+  widgets/ClickJumpSlider.h
   widgets/MarqueeLabel.cpp
   widgets/MarqueeLabel.h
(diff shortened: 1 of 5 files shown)


sykhro:
wasn't aware of this difference, LGTM

Issue #697, opened 2025-11-29
Failed to Build on Linux Due to Forward Declaration of Multiple Classes


Thysbelon:
**Build environment details**
Followed the wiki instructions? yes
Operating system: Ubuntu 24.04.3 LTS
CMake version: 3.28.3-1build7
Qt version: 6.4.2+dfsg-21.1build5
VGMTrans commit being built: 4c049328b75bc44d6775993dc568ba92004e2f14
Full CMake log: [vgmtrans-cmake-log.txt](https://github.com/user-attachments/files/23834992/vgmtrans-cmake-log.txt)

**Failure details**
The same issue occured on both GCC and Clang.
I used `cmake --build build --target vgmtrans --parallel 4` to build.

Thank you for reading.

mikelow:
I wonder if there's a difference with how your Debian (dfsg) version of Qt6 handles meta-object compilation compared to vanilla Qt6. It doesn't appear like it's an issue with vanilla Qt6, though I'm not certain. Our Github Actions script uses 6.8.3, and I've been compiling on various lesser versions of Qt6 without issue.

Could you try adding the following #includes to MenuBar.h and remove the respective forward class declarations?
```
#include "VGMFile.h"
#include "VGMColl.h"
#include "RawFile.h"
```

Does that reduce the errors?

Thysbelon:
I'll give it a try. Thank you.

mikelow:
Awesome. We'll make the change in the codebase to maximize compatibility.

PR #698, opened 2025-11-29
Fix a moc compilation issue afflicting debian qt6 package

mikelow:
A user reported a compilation problem (issue #697) on Ubuntu 24.04 using Qt version 6.4.2+dfsg-21.1build5. The errors occur in the meta object compilation step and pertain to VGMFile, VGMColl, and RawFile being forward declared and not defined.

The build log includes lines like:
```
qt6/QtCore/qmetatype.h:842:23: error: invalid application of 'sizeof' to an incomplete type 'VGMColl'
```

In MenuBar.h - the file causing the errors - we have three QList containers of those classes:
```
QList<VGMFile*> m_selectedVGMFiles;
QList<VGMColl*> m_selectedVGMColls;
QList<RawFile*> m_selectedRawFiles;
```

It appears these lines are causing this version of Qt6 to run sizeof(VGMFile) etc during the moc step, which fails as size cannot be determined from forward declarations. This problem is not occurring on vanilla versions of Qt6.

The fix is simply adding the relevant header files in MenuBar.h instead of forward declaring the classes.

## Motivation and Context
Maximize build compatibility.

## How Has This Been Tested?
The user reports this fixed the compilation entirely.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/Root.h
+++ b/src/main/Root.h
@@ -10,12 +10,12 @@
 
 #include "common.h"
 #include "VGMTag.h"
-#include "VGMFile.h"
-#include "VGMColl.h"
-#include "RawFile.h"
 
 class VGMScanner;
+class VGMColl;
 class VGMItem;
+class VGMFile;
+class RawFile;
 class VGMSeq;
 class VGMInstrSet;
 class VGMSampColl;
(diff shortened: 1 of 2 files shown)


sykhro (approved):
LTGM, let's squash-merge. Puzzling that it's happening in the first place...

PR #704, opened 2025-12-12
Konami TMNT2 format

mikelow:
This is a format used on Konami arcade games circa 1991. I've called it "TMNT2" to correspond to a hardware grouping in [MAME](https://github.com/mamedev/mame/blob/master/src/mame/konami/tmnt2.cpp) and [system16](https://www.system16.com/hardware.php?id=569), although this is a rough classification as it works on games outside that group and not necessarily every one within it. 

The machines pair a YM2151 with a Konami 53260 - a 4 channel 8-bit PCM / 4-bit ADPCM chip with variable pitch and panning. It's a bit more powerful than the CPS1, which uses a YM2151 paired with a OKI MSM6295 - a 4 channel 4-bit ADPCM chip with fixed pitch and no panning.

The most glaring omissions:
- support for a timing parameter that effectively changes tempo (event E0, sometimes)
- support for software vibrato (event E5)
- support for hardware vibrato (event E9)
- support for portamento / slides (event EB)

I've only added 4 games into mame_roms.json for now: tmnt2, ssriders, blswhstl, and vendetta (the last one only has sequence support).

Also worth noting that the K053260 ADPCM logic is very close but not identical to the K054539; it just uses a table with a single differing value.

## Audio demo
Needs to be unmuted. Rendered using the juce3 branch.

https://github.com/user-attachments/assets/23689398-a493-4772-a46a-df75a4ba645f

## Types of changes

## Checklist:

### DIFF
--- a/src/main/formats/KonamiTMNT2/KonamiTMNT2Instr.cpp
+++ b/src/main/formats/KonamiTMNT2/KonamiTMNT2Instr.cpp
@@ -103,7 +103,6 @@ bool KonamiTMNT2SampleInstrSet::parseMelodicInstrs() {
     VGMRgn* rgn = new VGMRgn(instr, offset, sizeof(konami_tmnt2_instr_info));
     rgn->sampOffset = instrInfo.start();
     rgn->sampDataLength = (instrInfo.length_msb << 8) | instrInfo.length_lsb;
-    // rgn->setVolume((instrInfo.volume & 0x7F) / 127.0);
 
     instr->addRgn(rgn);
     aInstrs.push_back(instr);
@@ -188,7 +187,6 @@ bool KonamiTMNT2SampleInstrSet::parseDrums() {
       double relativePitchCents = k053260_pitch_cents((drumInfo.pitch_hi << 8) + drumInfo.pitch_lo);
       rgn->coarseTune = relativePitchCents / 100;
       rgn->fineTune = static_cast<int>(relativePitchCents) % 100;
-      // rgn->setVolume((drumInfo.volume & 0x7F) / 127.0);
 
       drumKit->addRgn(rgn);
       drumNum += 1;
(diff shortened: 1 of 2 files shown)


src/main/formats/KonamiTMNT2/KonamiTMNT2Instr.cpp
```diff
@@ -0,0 +1,291 @@
...
+      offset,
+      sizeof(konami_tmnt2_instr_info),
+      instrNum / 128,
+      instrNum % 128,
+      name
+    );
+    VGMRgn* rgn = new VGMRgn(instr, offset, sizeof(konami_tmnt2_instr_info));
+    rgn->sampOffset = instrInfo.start();
+    rgn->sampDataLength = (instrInfo.length_msb << 8) | instrInfo.length_lsb;
+    // rgn->setVolume((instrInfo.volume & 0x7F) / 127.0);
```
sykhro: If this is not deadcode, it should have an explanation comment

src/main/formats/KonamiTMNT2/KonamiTMNT2Seq.cpp
```diff
@@ -0,0 +1,693 @@
...
+  m_globalTranspose = 0;
+  m_masterAttenYM2151 = 0;
+  m_masterAttenK053260 = 0;
+}
+
+bool KonamiTMNT2Seq::parseTrackPointers() {
+  // if (m_ym2151TrackOffsets.size() == 8) {
+  //   auto *track = new KonamiTMNT2Track(true, this, m_ym2151TrackOffsets[6], 0, "FM Track");
+  //   aTracks.push_back(track);
+  // } else {
```
sykhro: more deadcode?

sykhro (changes requested):
Great! Some minor observations

src/main/formats/KonamiTMNT2/KonamiTMNT2Instr.cpp
mikelow: agreed. it's deleted in the later PRs. I'll delete it here.

src/main/formats/KonamiTMNT2/KonamiTMNT2Seq.cpp
mikelow: Left it as convenient debug code as I work on the format. But I don't mind deleting it.

PR #712, opened 2026-01-02
Fix silently failing saves for paths with unicode on Windows (issue #711)

### LINKED ISSUE #711: Using a path containing Chinese characters will cause the program to crash or malfunction.
### Environment details

* Operating System+version: **Windows Windows11,23H2**
* VGMTrans version: **VGMTrans-v1.2-windows-x64**
* Game ROM/music dump name and sha256: **ea8e238c817ba5a7000b8988da7f402e119f9d5ab510c0eb99ba7642857f5686.nds**, **ea8e238c817ba5a7000b8988da7f402e119f9d5ab510c0eb99ba7642857f5686**

### Describe the bug

It seems that it includes importing files and exporting directories. As long as the directory contains Chinese characters, it will not function properly.

### Steps to reproduce

1. Import an nds game that does not contain Chinese characters in the current directory.
2. Choose any piece of music and select "Export MIDI and SF2".
3. The selected directory includes directories with Chinese characters for export.
4. Program crashed.

### Expected behaviour

The ability to read and export directories that contain encodings other than the one used by the program.

### Additional context

_No response_

### Logs

_No response_

mikelow:
This fixes #711.

- added pathFromUtf8() to convert to a UTF-8 string to a filesystem path
- updated buffer-based save routines to open files through UTF-8-aware filesystem paths, preventing silent failures when saving to directories with non-ASCII characters on Windows

## How Has This Been Tested?
Tested that saving works for unicode-containing paths that previously didn't work, and tested that ASCII-only paths still work. Tested on Windows 10 and macOS (only Windows had exhibited the bug).

## Types of changes

## Checklist:

### DIFF
--- a/src/main/util/helper.h
+++ b/src/main/util/helper.h
@@ -12,7 +12,7 @@
 #include <filesystem>
 
 inline std::filesystem::path pathFromUtf8(const std::string& path) {
-  return { std::u8string(reinterpret_cast<const char8_t*>(path.c_str())) };
+  return std::filesystem::u8path(path);
 }
 
 template <class T>


sykhro (approved):

mikelow:
> I'm a little puzzled because I was assuming all of the strings we handled were already UTF-8

This confused me as well. You are correct - our paths are stored as UTF-8 in std::strings.

The issue is with how Windows is handling std::ofstream. Windows API calls come in two flavors:  "A" (ANSI) functions that take (ie `char*`) and "W" (Wide) functions that take (`wchar_t*` ie UTF-16).

The issue is that on Windows, when we construct a `std::ofstream` using a `std::string`, it calls the A Windows API functions, so the std::string bytes are treated as ANSI and not UTF-8. The fix is to call the `std::ofstream` constructor that takes a `std::filesystem::path`, which calls the W Windows API functions instead.

sykhro:
> I've updated pathFromUtf8() to use the deprecated std::filesystem::u8path constructor instead of a more dangerous reinterpret_cast call.
> 
> What we should probably do is stop using std::string for file paths and switch to using std::filesystem::path. This would be a much larger change, though.

Welp! Just noticed we don't define `_UNICODE`/`UNICODE` anywhere so it's never auto-picking the W family of functions (should be in CMake presets).

PR #720, opened 2026-01-08
add NDEBUG definition for non-debug builds

mikelow:
When building the app with our CMake presets, NDEBUG is never defined for release builds. This causes failing assert() calls to crash the app.

## How Has This Been Tested?
Confirmed that failling asserts no longer crash release builds. Debug builds still throw on failed asserts.

## Types of changes

## Checklist:


sykhro (approved):
I'm somewhat surprised that CMake doesn't define this automatically in Release mode...

PR #725, opened 2026-01-11
DLS: add vibrato and fix region fineTune and attenuation

mikelow:
This is a new feature and a fix for DLS conversion:
1. Adds vibrato delay, frequency, and depth parameters to DLS regions, if applicable (same as we do for SF2)
2. Fixes logic where fineTune and attenuation were pulled from either a region or the sample, but never combined together. Now, the two values are added and applied to the region-level WSMP. This was previously fixed for SF2, but never for DLS.
3. Fixes region/sample attenuation using wrong sign in DLS conversion (should be negative to apply attenuation).

## How Has This Been Tested?
Tested vibrato and attenuation on [SpessaFont](https://spessasus.github.io/SpessaFont/) and [SpessaSynth](https://spessasus.github.io/SpessaSynth/).

## Types of changes

## Checklist:

### DIFF
--- a/src/main/conversion/DLSConversion.cpp
+++ b/src/main/conversion/DLSConversion.cpp
@@ -237,20 +237,13 @@ bool mainDLSCreation(
         if (realUnityKey == -1)
           realUnityKey = 0x3C;
 
-        short realFineTune;
-        if (rgn->fineTune == 0)
-          realFineTune = samp->fineTune;
-        else
-          realFineTune = rgn->fineTune;
-
-        long realAttenuation;
-        if (rgn->attenDb() == 0 && samp->attenDb() == 0)
-          realAttenuation = 0;
-        else if (rgn->attenDb() == 0)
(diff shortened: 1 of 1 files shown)


mikelow:
> I was wondering about our DLS correctness just recently. It would be great to have some tests but I honestly can't picture how they'd work (some sort of integration with some synth?). I'm always afraid of code at the format-level potentially accounting for things that should be done at DLS-conversion level instead

Yeah, DLS has very much been neglected once the app switched to rendering SF2 for preview. I was happy to discover SpessaSynth/Font as it provides an easy way to test DLS files. Hopefully it's accurate. I'm not sure how we'd do a unit or integration test, but that's not to say we shouldn't pursue it (and not just for DLS).

I hope too that we've moved away from a lot of the format-level stuff being encoded outside of format code. I may want to revisit the SynthFile class at some point, as it was originally intended as an abstraction between the VGM classes and SF2 / DLS. It's not functioning that way (it's not even used with DLS), and it seems kind of redundant.

PR #730, opened 2026-01-29
RhiHexView: Refactor VGMItem::getItemAtOffset() to avoid the O(n) lookup. 

mikelow:
## Description

Currently, VGMItem::getItemAtOffset() iterates through all children linearly, which can be slow with large child lists, and it also needs to handle items with overlapping ranges correctly. This PR significantly speeds up lookup by using a binary search tree in conjunction with a bounded backward scan, which allows us to quickly find the list of items that overlap an offset so we can iterate over a small subset of sibling items rather than all siblings.

Codex wrote this and does a better job summarizing the details:
- Approach: keep children sorted by dwOffset, binary‑search for the first child whose start is after the target offset, then scan backward through only the potentially overlapping items.
- Prefix‑max bound: maintain a cached array where prefixMaxEnd[i] is the maximum end offset among children [0..i]. While scanning backward, if prefixMaxEnd[i] <= offset, then no earlier child can overlap the offset, so the scan can stop safely.
- Complexity: O(log n + k), where k is the number of overlapping candidates.
- Implementation notes: cache invalidated on child mutations; rebuilt lazily when a lookup needs it.

## Motivation and Context
This is one part of a much larger change: a rewrite of HexView using Qt's "Rendering Hardware Interface" (RHI) so that we can highlight active events in real-time during playback. The CPU-bound rendering of Qt Widgets, which our current HexView uses, is simply too slow. RHI let's us utilize the platform's 3d API (D3D, Metal, OpenGL). It was necessary to improve the performance of getItemAtOffset() for real-time highlighting.

The branch `event-tracking3` contains the latest work on the RHI HexView with real-time highlighting. It's finished, short of some possible cleanup / commenting.

## How Has This Been Tested?
I've been using this on the new HexView and tested many. I have yet to see any problems.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/components/VGMItem.cpp
+++ b/src/main/components/VGMItem.cpp
@@ -46,9 +46,22 @@ VGMItem* VGMItem::getItemAtOffset(uint32_t offset, bool matchStartOffset) {
     }
   }
 
-  for (const auto child : m_children) {
-    if (VGMItem *foundItem = child->getItemAtOffset(offset, matchStartOffset))
+  ensureChildrenSorted();
+  // Binary search by start offset, then scan backwards only across potentially overlapping items.
+  const auto begin = m_children.begin();
+  const auto end = m_children.end();
+  auto it = std::upper_bound(begin, end, offset,
+      [](uint32_t off, const VGMItem* child) { return off < child->dwOffset; });
+
+  for (auto idx = static_cast<size_t>(it - begin); idx > 0; --idx) {
+    VGMItem* child = m_children[idx - 1];
+    if (VGMItem* foundItem = child->getItemAtOffset(offset, matchStartOffset)) {
       return foundItem;
+    }
+    if (!m_childrenPrefixMaxEnd.empty() &&
+        m_childrenPrefixMaxEnd[idx - 1] <= static_cast<uint64_t>(offset)) {
+      break;
+    }
   }
 
   return nullptr;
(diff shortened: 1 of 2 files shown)


src/main/components/VGMItem.cpp
```diff
@@ -95,40 +108,82 @@ bool VGMItem::isValidOffset(uint32_t offset) const {
...
   m_children.clear();
+  m_childrenSorted = true;
+  m_childrenPrefixMaxEnd.clear();
+}
+
+void VGMItem::ensureChildrenSorted() {
+  // Lazy path: only sort/rebuild when a lookup needs it.
+  if (m_children.size() < 2) {
+    m_childrenSorted = true;
+  }
```
sykhro: It took me a bit to understand this, but I think it would be better to remove this and, in do
```
if (m_children.size() == 1) return m_children[0]->getItemAtOffset(offset, matchStartOffset);
```
directly in `getItemAtOffset()` (after the empty() check). Beyond readability, it also skips the potentially expensive `.reserve()` call down below

sykhro (changes requested):
I'm not going to lie: had to stare at these changes for a bit to figure out how it works. I think it's correct, I have a minor request

mikelow:
Closing this, as codex review found a major problem - we're not invalidating the the prefix max cache when dwOffset or unLength are set directly. I will instead make two new PRs - one which refactors dwOffset and unLength so that they use getter and setter methods (and I'll rename them to m_offset, m_length), a second PR which replicates the change here and adds invalidation to the cache in the setter methods.

@sykhro, I see your feedback and I'll incorporate it. Sorry for the hassle!

sykhro:
> Closing this, as codex review found a major problem - we're not invalidating the the prefix max cache when dwOffset or unLength are set directly. I will instead make two new PRs - one which refactors dwOffset and unLength so that they use getter and setter methods (and I'll rename them to m_offset, m_length), a second PR which replicates the change here and adds invalidation to the cache in the setter methods.
> 
> @sykhro, I see your feedback and I'll incorporate it. Sorry for the hassle!

No worries! I don't remember what manipulates these directly, but it's good to be safe. We have a lot of scary-public-usage around the codebase

PR #735, opened 2026-01-31
vgmtrans-shell: Preliminary implementation of introspection tool

sykhro:
## Description
This PR introduces a new interactive terminal-based program, `vgmtrans-shell`.
Think of it as an actually useful alternative to `vgmtrans-cli`.

## Motivation and Context
I recently came across an old SSD and decided to sift through it before time could claim its contents forever.
[...]
After some rebasing, here is `vgmtrans-shell`.

There were a few reasons to implement this:
- as a terminal-based interface for when Qt isn't available or undesired
- for automated (integration) testing
  - you feed it a text file via stdin and then compare stdout to an existing file. easy!
- to eventually become a comfortable environment/tool for debugging and working on reverse engineering formats. At the start, I was envisioning integrating it with lldb as well as being able to poke/edit live data structures.

It depends on `linenoise` for basic shell interaction (history, tab completions).
Since I need  it here, I have reintroduced `libfmt` as a submodule and pointed spdlog to use that. This also removed the transitive dependency on `libfmt` from `vgmtranscore`.

## How Has This Been Tested?
I ran it and tried all of the commands. Dogfooding myself by using it to implement a new format (in a separate PR).

Not pictured: tab completions, history.

image

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,7 @@ jobs:
 
       - name: "Build project"
         run: cmake --build --preset "macos-${{ matrix.target }}-release" \
-          --target vgmtrans
+          --target vgmtrans vgmtrans-shell
 
       - name: "Package DMG (macOS)"
         working-directory: "build/macos-${{ matrix.target }}/"
@@ -230,7 +230,7 @@ jobs:
           cmake --preset win-${{ matrix.target }}
 
        - name: "Build project"
-         run: cmake --build --preset win-${{ matrix.target }}-release --target vgmtrans
+         run: cmake --build --preset win-${{ matrix.target }}-release --target vgmtrans vgmtrans-shell
 
        - name: "Delete empty directories"
          run: Get-ChildItem -Directory -Recurse | Where-Object { $_.GetFileSystemInfos().Count -eq 0 } | Remove-Item -Force -Recurse


mikelow:
@codex review

mikelow:
> When I'm trying to hackily plumb in new things like a scanner, I find that sometimes the UI can get in the way (not showing items, accidentally triggering some callback infinite loop, ...);

Yeah, that makes sense. I can imagine faster iteration for early-stage format implementation. Also, obviously could be useful for integration testing, as you mentioned.

> Interesting, I'm on macOS too (15.7.3, kitty terminal). I'll try to see if I can reproduce. What's the value of TERM? Maybe try with export TERM=xterm-color before running it?

I'm running Sonoma 14.1. `TERM` was set to `xterm-256color`, and changing it to `xterm-color` didn't help. I installed kitty and ran it in there - worked fine. I'd be curious if your stock Terminal app has the same problem as mine.

sykhro:
> I'd be curious if your stock Terminal app has the same problem as mine.

yep, broken in the same way. It might affect any `libfmt` app, I'll check

btw, codex was right. I tried building on windows and it failed: CI has an explicit `--target vgmtrans` which I completely forgot about. So what I'm thinking is:
- default ENABLE_SHELL to OFF if `termios.h` is not available (or stub linenoise if it's a quick job - it won't have tab completions and arrow keys and history but it's better than nothing)
- update remove that specific target option from CI

mikelow (approved):

sykhro:
> So the shell won't work in Windows unless linenoise is stubbed? That's surprising and unfortunate. I'm not opposed to the easy solution of just disabling on Windows. We can solve it down the road if we choose.

I remembered M$ ported redis to Windows a few years back and.. https://github.com/microsoftarchive/redis/blob/win-3.2.100/deps/linenoise/linenoise.c
they ported linenoise too! Let's see how much of a pain it is to port this over.
The library is not updated super often so it should be a tolerable maintainance burden. Worst case scenario I can try to send a PR upstream :)

I think the reason for building `vgmtrans` explicitly was because of third-party dependencies having extra targets we don't care about (higher build times). I'll check the deployment script keeps working

mikelow:
> I remembered M$ ported redis to Windows a few years back and.. https://github.com/microsoftarchive/redis/blob/win-3.2.100/deps/linenoise/linenoise.c
> they ported linenoise too! Let's see how much of a pain it is to port this over.

Good find. Unfortunate they didn't find a way to integrate this into the official project.

sykhro:
adding Windows support to linenoise was surprisingly easy, I kinda cheated with the `setmode` binary call but it seems to work just as well as mac/linux! I'll wait for CI and then merge

PR #742, opened 2026-02-09
RhiHexView 2: New HexView rendering foundation

mikelow:
This branch introduces the core RHI architecture for HexView and moves rendering responsibilities out of the old paint-based widget flow. The key structural change is that `HexView` is now a `QAbstractScrollArea` with an embedded RHI surface in its viewport, rather than a plain widget doing all rendering in `paintEvent`.

The main entry point for this shift is in `HexView.h` and `HexView.cpp`. `HexView` now focuses on interaction state, layout, selection state, and frame data preparation. Rendering happens via a dedicated host/widget/renderer stack. When HexViewRhiRenderer renders a frame, it first requests an immutable frame snapshot from `HexView` via `HexView::captureRhiFrameData()`. HexView can also signal that a new frame should be rendered and specify targeted invalidation with `HexView::requestRhiUpdate(markBaseDirty, markSelectionDirty)` so the renderer can avoid rebuilding everything on every change.

A new frame contract is defined in `HexViewFrameData.h`. `HexViewFrame::Data` carries all render inputs for one frame: viewport metrics, scroll position, column visibility flags, style table + style-id map, selection ranges, fade ranges, palette-derived colors, and glyph atlas references/versioning. This avoids renderer-side dependence on mutable widget internals and makes data flow explicit.

The new rendering classes split responsibilities like so:
- `HexViewRhiHost.h` + `.cpp`: lightweight container that owns the renderer and surface widget.
- `HexViewRhiWidget.h` + `.cpp`: `QRhiWidget` subclass that selects backend per platform, drives initialize/render/release, and routes input.
- `HexViewRhiEventForwarder.h`: forwards mouse/key/wheel/tooltip events from the RHI surface to HexView’s viewport.
- `HexViewRhiInputCoalescer.h`: coalesces drag/wheel input and normalizes Alt+wheel behavior on Linux/Windows before dispatch.

`HexViewRhiHost` may seem like a redundant layer of abstraction, as within this branch it simply wraps the `HexViewRhiWidget` class. However, it will take on purpose in the next PR with the introduction of a new `HexViewRhiWindow` class. If you run this branch on Windows or macOS, you will find that draw performance is abysmal, while Linux is perfectly fine. It's an issue with QRhiWidget, and if I understand correctly, caused by using `QRhiWidget` within a `QSplitter` or `QMdiArea` (we do both). This will be discussed further in the next PR.

Likewise, the `HexViewRhiEventForwarder` encapsulates event handling logic and will be shared by both `HexViewRhiWidget` and `HexViewRhiWindow` when it is introduced.

The `HexViewRhiInputCoalescer` improves performance by limiting the number of `UpdateRequest` events that are sent to `HexViewRhiWidget` (and `HexViewRhiWindow`). High-polling input devices (like a "gaming" mouse) can cause Qt to hammer the class with Update events faster than screen refresh rate, which can hog the CPU and throttle rendering performance.

The renderer in `HexViewRhiRenderer.h` + `.cpp` introduces a 3-pass pipeline:
1. Base content pass into `contentTex` (background spans + glyphs).
2. Selection mask pass into `maskTex` (selection rectangles only).
3. Full-screen composite pass to the output target (samples `contentTex` + `maskTex` and applies column-aware dim/restore logic).

At a lower level, the renderer builds instanced geometry (rect + glyph streams), maintains a cache window of file lines around the viewport, uploads a glyph atlas texture on version change, and updates dynamic instance buffers only when marked dirty. This is where the branch establishes most of the performance and architecture groundwork.

Shader additions in `src/ui/qt/workarea/hexview/shaders/` correspond directly to that pipeline:
- `hexquad.vert` / `hexquad.frag`: instanced colored rectangles (background + mask quads).
- `hexglyph.vert` / `hexglyph.frag`: instanced glyph quads sampling glyph atlas alpha.
- `hexscreen.vert`: full-screen quad with backend-dependent UV flip handling.
- `hexcomposite.frag`: combines `contentTex` and `maskTex`, dims selected columns via overlay opacity, and restores selected regions.

I won't pretend to have a strong understanding of all the renderer and shader logic.

Using shaders and QRhi requires we update dependencies in `src/ui/qt/CMakeLists.txt`. We add the `Qt6::ShaderTools` so that we can call `qt_add_shaders` to compile/embed shaders. For later Qt6 versions, it is necessary to include the `Qt6::GuiPrivate` package.

`VGMFileView.cpp` also removes the extra `QScrollArea` wrapper and hosts `HexView` directly, since HexView now manages its own scrollable viewport behavior.

## How Has This Been Tested?
This has been tested on macOS. I have only thoroughly the final PR which completes implementation.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/workarea/hexview/HexViewRhiWidget.cpp
+++ b/src/ui/qt/workarea/hexview/HexViewRhiWidget.cpp
@@ -66,7 +66,8 @@ void HexViewRhiWidget::render(QRhiCommandBuffer* cb) {
   info.renderPassDesc = rt->renderPassDescriptor();
   info.pixelSize = rt->pixelSize();
   info.sampleCount = rt->sampleCount();
-  info.dpr = rt->devicePixelRatio();
+  // Use the widget's DPR as reported by the windowing system. This is important on some hi-dpi systems (notably Retina screens)
+  info.dpr = devicePixelRatio();
   m_renderer->renderFrame(cb, info);
 }
 


sykhro:
(I will get to actually reviewing the PR soon)

src/ui/qt/workarea/hexview/HexViewRhiWidget.cpp
```diff
@@ -0,0 +1,89 @@
...
+  if (!rt) {
+    return;
+  }
+
+  HexViewRhiRenderer::RenderTargetInfo info;
+  info.renderTarget = rt;
+  info.renderPassDesc = rt->renderPassDescriptor();
+  info.pixelSize = rt->pixelSize();
+  info.sampleCount = rt->sampleCount();
+  info.dpr = rt->devicePixelRatio();
```
sykhro: Fixes blurriness on macOS:
```suggestion
  // Use the widget's DPR as reported by the windowing system. This is important on some hi-dpi systems (notably Retina screens)
  info.dpr = devicePixelRatio();
```

sykhro (changes requested):
Great work overall.
I've left a fix for a DPI issue on macOS (+ comment about vendored Qt)

mikelow:
Another subject I didn't discuss yet is why I'm opting for OpenGL over Vulkan on Linux. I'll talk more about platform specific complications in the next PR. Suffice to say there's a good reason.

sykhro:
> > just so you know, this breaks our vendored Qt: https://github.com/vgmtrans/vgmtrans-ext-win/blob/10629a63e499b7306841c51715e6f68e21557946/vgmtrans-amd64.yml#L46 two possible choices:
> > 
> > * I try to recompile it with the shader things and update the binaries
> > * we get rid of the vendored Qt
> > 
> > I honestly can't see a reason for users building VGMTrans from source, now that CI builds both debug and release mode for every commit. if they're that kind of power user, they can certainly install Qt and set `Qt_DIR` to point to it. maybe we can go for option 2 and update build instructions?
> 
> Ah, I think I discovered this early on and totally forgot. I looked into the tool we use for downloading a subset of Qt modules - it seems like a major headache to maintain as new version of Qt roll out. Given that we only support this on Windows anyway, I'm in favor of dropping it. We probably need to update the wiki on build instructions anyway - perhaps we can point to the best install method for Qt on Windows.

yes, it's not ideal. I kept it around as convenience, but I would no longer rely on it. good riddance

PR #746, opened 2026-02-09
RhiHexView 3: HexViewRhiWindow

mikelow:
## Description
This PR adds the new HexViewRhiWindow class, which is used by Windows and macOS to fix the poor performance of HexViewRhiWidget. The cause of this platform-specific issue is unclear, but may be related to the fact we embed a QRhiWidget within a QSplitter or MDIArea.

`HexViewRhiHost` owns the RHI surface (either the Window or Widget class) and interfaces to it in a platform-agnostic way, allowing `HexView` to be happily ignorant of these details. When `HexViewRhiHost` instantiates a `HexViewRhiWindow`, it contains it within a Widget using `QWidget::createWindowContainer()`. 

A cleaner solution would be to get rid of `HexViewRhiWidget` and use `HexViewRhiWindow` on all platforms, but Qt laughs in our face and insists we be made to suffer. No, the Wayland architecture, which has widely replaced X - or more specifically, Qt's implementation on top of Wayland - does not like something about embedding a HexViewRhiWindow within the Widget of another window. This is probably because Wayland has no concept of "child windows". When we try to use HexViewRhiWindow, we get a permanently stuck, unresizable window on top of the rest of our app, and drawing breaks.

Thus, we need `HexViewRhiWidget`, `HexViewRhiWindow`, and their loving parent `HexViewRhiHost`.

`HexViewRhiWindow` differs from `HexViewRhiWidget` in that it has to own and handle the `RHI/swapchain` lifecycle, which includes handling window resizes. Both classes handle incoming QEvents (and share logic using `HexViewRhiEventForwarder`), but `HexViewRhiWindow`, being its own QWindow, also has to handle drag and drop related events. When such events are received, it emits signals which are routed to `MainWindow` via `HexViewRhiHost` so that `MainWindow` can own drag/drop behavior.

The MainWindow drag and drop behavior is thus organized into methods for easier integration as signal callbacks.

HexViewRhiWindow also has logic to limit the rate at which scroll bar events can trigger render calls. High-polling input devices can cause a deluge of update events, killing frame renders.

On Linux systems, we opt to use OpenGL over Vulkan. For whatever reason, using a Vulkan window surface causes app window decorations (title bar and its minimize, maximize, close buttons) to disappear.

This PR also sets the Qt::AA_DontCreateNativeWidgetSiblings attribute at startup. Without this attribute, the QSplitters that separate the MDIArea from our docks cannot be clicked and dragged consistently due to a complication with the Rhi Window. Setting this attribute has the adverse effect of making HexView's vertical scrollbar never appear on macOS, which is "transient" by default, meaning it appears and disappears only when used. Presumably it's no longer being drawn _on top of_ the surface. We fix this by making it non-transient in HexView's constructor, so it never fades in/out.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/CMakeLists.txt
+++ b/src/ui/qt/CMakeLists.txt
@@ -99,6 +99,8 @@ add_executable(
   workarea/hexview/HexViewRhiRenderer.h
   workarea/hexview/HexViewRhiWidget.cpp
   workarea/hexview/HexViewRhiWidget.h
+  workarea/hexview/HexViewRhiWindow.cpp
+  workarea/hexview/HexViewRhiWindow.h
   workarea/MdiArea.cpp
   workarea/MdiArea.h
   workarea/RawFileListView.cpp
--- a/src/ui/qt/MainWindow.cpp
+++ b/src/ui/qt/MainWindow.cpp
@@ -175,43 +175,51 @@ void MainWindow::routeSignals() {
 
 void MainWindow::dragEnterEvent(QDragEnterEvent *event) {
   event->acceptProposedAction();
-  if (m_dragOverlay) {
-    updateDragOverlayGeometry();
-    m_dragOverlay->show();
-    m_dragOverlay->raise();
-  }
+  showDragOverlay();
 }
 
 void MainWindow::dragMoveEvent(QDragMoveEvent *event) {
   event->acceptProposedAction();
-  if (m_dragOverlay && !m_dragOverlay->isVisible()) {
-    m_dragOverlay->show();
-    m_dragOverlay->raise();
-  }
+  showDragOverlay();
 }
 
 void MainWindow::dragLeaveEvent(QDragLeaveEvent *event) {
   event->accept();
+  hideDragOverlay();
+}
+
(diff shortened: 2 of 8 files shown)


sykhro (approved):
LGTM!

sykhro:
> A cleaner solution would be to get rid of `HexViewRhiWidget` and use `HexViewRhiWindow` on all platforms, but Qt laughs in our face and insists we be made to suffer. No, the Wayland architecture, which has widely replaced X - or more specifically, Qt's implementation on top of Wayland - does not like something about embedding a HexViewRhiWindow within the Widget of another window. This is probably because Wayland has no concept of "child windows". When we try to use HexViewRhiWindow, we get a permanently stuck, unresizable window on top of the rest of our app, and drawing breaks.
> [...]
> On Linux systems, we opt to use OpenGL over Vulkan. For whatever reason, using a Vulkan window surface causes app window decorations (title bar and its minimize, maximize, close buttons) to disappear.

Both these things are bizarre. I'm far removed from the Linux desktop experience, so I wouldn't know where to look for a fix. I think these are both acceptable compromises!

PR #752, opened 2026-02-12
RhiHexView 6: Active event highlighting

mikelow:
## Description
This branch adds a highlight for active events. It doesn't include a fade or glow effect yet.

`VGMFileView` now forwards playback state into HexView, instead of only updating VGMFileTree. When playback context is valid, HexView is toggled into playback-active mode and receives the current active playback items; when playback becomes invalid/stops, HexView playback state is cleared. This keeps HexView and tree playback state synchronized.

On the HexView side, playback selections are now tracked separately from manual selection, included in frame data, and used by highlight-state logic. Playback-active mode also participates in the dim/highlight state so playback highlights remain visible even when there is no manual selection. (There are couple lines in HexView.cpp that set selection shadow params. Their inclusion here is a harmless mistake, as shadow isn't rendered yet).

In the renderer, playback ranges are added to the existing mask pass. For this PR, playback uses the same visible mask channel as selection, so visuals stay intentionally simple: active playback bytes are highlighted through the current composite path. Future PRs will add a new glow pass and fade animation.

## Motivation and Context
Showing active events is useful for debugging. Also, it's fun.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/workarea/VGMFileView.cpp
+++ b/src/ui/qt/workarea/VGMFileView.cpp
@@ -305,9 +305,8 @@ void VGMFileView::onPlaybackPositionChanged(int current, int max, PositionChange
     m_treeview->setPlaybackItems(m_playbackItems);
   }
 
-  const bool playbackActive = shouldHighlight && !m_playbackItems.empty();
   if (hexVisible) {
-    m_hexview->setPlaybackActive(playbackActive);
+    m_hexview->setPlaybackActive(shouldHighlight);
   }
 
   if (m_playbackItems == m_lastPlaybackItems) {


mikelow:
> Looks great! What happens if a file is closed in the middle of rendering?

Haven't experienced any issues yet. I asked codex to analyze for a race condition:

Prompt: During playback with the HexView open, what happens when the VGMSeq file is closed? Do we ensure there's no race condition that might allow the HexView to attempt to read from the closed file?
> Short answer: mostly yes in practice, but not as a strict lifetime guarantee.
>
> What happens on close during playback:
> 1. File close emits `UI_removeVGMFile` in `Root.cpp:186`, and the actual `delete targFile` happens later at `Root.cpp:204`.
> 2. UI closes the view in `MdiArea.cpp:367` via `close()` (`:375`), and the view is `WA_DeleteOnClose` (`VGMFileView.cpp:32`).
> 3. Associated collection removal triggers `SequencePlayer::stop()` (`SequencePlayer.cpp:100`), which emits playback/status signals.
> 4. Closed `VGMFileView` playback handler returns immediately when not visible (`VGMFileView.cpp:201`), before touching `m_vgmfile`.
> 5. RHI window render path also bails when not exposed (`HexViewRhiWindow.cpp:315`).
> 
> So there’s no obvious active race where HexView keeps reading a closed file after close.  
> But there is still a theoretical gap because `m_vgmfile` is a raw pointer and file deletion happens before deferred widget destruction. If you want this airtight, I’d harden by disconnecting playback signals and nulling/invalidation on close.
>

PR #753, opened 2026-02-12
RhiHexView 7: Add active event fade animation

mikelow:
## Description
This branch adds fade behavior on top of the previous PR’s active playback highlighting. HexView now tracks both active playback selections and a second set of fading playback selections. When playback items advance, removed items are moved into a fade list with a start timestamp; a timer updates each fade alpha over time and prunes finished entries. That keeps fade progression in HexView state, not in renderer-side ad hoc logic.

Frame data now carries fading playback ranges and per-range alpha to the renderer. In the mask pass, channels are explicitly encoded as: manual selection in R, active playback in G, fading playback membership in B, and fading playback alpha in A. The renderer emits mask rects accordingly, including per-fading-range alpha writes.

The composite shader was updated to consume those mask channels directly. Highlight restoration is now computed as max(selection, activePlayback, fadingPlayback * fadeAlpha), so fading playback decays visually without introducing glow/edge effects yet. That will come in the next PR.

## Motivation and Context
Most sequence events are some manner of control event with 0 duration. These events are treated as having a nominal duration for the sake of highlighting, but they still show up as a blip if they get rendered at all. Adding a fade to highlighted events makes it clearer when 0 duration events get activated. It also helps visualize which events are activating as the seek bar is dragged.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/workarea/hexview/HexViewRhiRenderer.cpp
+++ b/src/ui/qt/workarea/hexview/HexViewRhiRenderer.cpp
@@ -1199,17 +1199,7 @@ void HexViewRhiRenderer::buildSelectionInstances(int startLine, int endLine,
   }
 
   if (!frame.fadePlaybackSelections.empty()) {
-    // Fading playback uses blue for membership and alpha for fade amount.
-    std::vector<HexViewFrame::SelectionRange> fadeRanges;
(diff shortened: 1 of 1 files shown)


sykhro:
very useful and very clever shader design. will test it out/review soon

PR #759, opened 2026-02-20
RhiHexView 10: add Linux QRhiWidget primer in MainWindow

mikelow:
## Description
This is the final planned PR for the RHI HexView rewrite 🎉. I should have put this one earlier, but didn't want to rearrange these iterative branches.

This PR fixes a problem on Linux, which uses HexViewRhiWidget instead of the HexViewRhiWindow class. When HexViewRhiWidget is first displayed, Qt (or the OS?) swaps the app window, presumably with one that can utilize the GPU more directly. This swap isn't hidden, it's visible and jarring - the app window closes and a new window pops up at a different screen position.

To fix this, we add a "primer" QRhiWidget in MainWindow, which lets us initialize the app in an already GPU-capable window so there's never any swap. During `createElements()`, the app creates this widget, pins it to the OpenGL API (`setApi(QRhiWidget::Api::OpenGL)`), marks it non-interactive (`WA_TransparentForMouseEvents`, `NoFocus`), and keeps it hidden. Thus, the primer widget only affects initialization and doesn't interfere in any way with the UI.

## Testing
This PR marks the code state where I did extensive testing across Windows, macOS, and Linux and tested for issues with just about every format we support.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/main_ui.cpp
+++ b/src/ui/qt/main_ui.cpp
@@ -52,18 +52,19 @@ int main(int argc, char *argv[]) {
   QFontDatabase::addApplicationFont(":/fonts/Roboto_Mono/RobotoMono-VariableFont_wght.ttf");
 
   MainWindow window;
+
 #if defined(Q_OS_LINUX) && QT_CONFIG(opengl)
+  // Prime QRhiWidget once at startup to avoid first-use window re-creation. Not necessary for other platforms
+  // where we use a QWindow instead of QRhiWidget.
   auto* rhiPrimer = new QRhiWidget(&window);
   rhiPrimer->setApi(QRhiWidget::Api::OpenGL);
-  rhiPrimer->setAttribute(Qt::WA_TransparentForMouseEvents);
-  rhiPrimer->setFocusPolicy(Qt::NoFocus);
   rhiPrimer->hide();
 #endif
+
   window.show();
+
 #if defined(Q_OS_LINUX) && QT_CONFIG(opengl)
-  QTimer::singleShot(0, &window, [rhiPrimer]() {
-    rhiPrimer->deleteLater();
-  });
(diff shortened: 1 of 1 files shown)


sykhro (commented):
Thanks for all the great work, happy to see this feature-complete.

About this PR, have you already tried setting `QSurfaceFormat::setDefaultFormat()` before `MainWindow` is created? I believe that's the way to force all windows to be in a OGL context

mikelow:
> Thanks for all the great work, happy to see this feature-complete.

Thanks!

> About this PR, have you already tried setting `QSurfaceFormat::setDefaultFormat()` before `MainWindow` is created? I believe that's the way to force all windows to be in a OGL context

I just tried this approach, but unfortunately it doesn't work. I tried calling setDefaultFormat() before instantiating VGMTransApplication and immediately after. Either way, it still recreates the top-level window when HexView is first shown.

But I did find that I can create short-lived QRhiWidget primer in main_ui and delete it after showing the app window. That seems cleaner than keeping a permanent hidden member on MainWindow.

sykhro:
> > Thanks for all the great work, happy to see this feature-complete.
> 
> Thanks!
> 
> > About this PR, have you already tried setting `QSurfaceFormat::setDefaultFormat()` before `MainWindow` is created? I believe that's the way to force all windows to be in a OGL context
> 
> I just tried this approach, but unfortunately it doesn't work. I tried calling setDefaultFormat() before instantiating VGMTransApplication and immediately after. Either way, it still recreates the top-level window when HexView is first shown.
> 
> But I did find that I can create short-lived QRhiWidget primer in main_ui and delete it after showing the app window. That seems cleaner than keeping a permanent hidden member on MainWindow.

That sounds cleaner indeed!

PR #767, opened 2026-02-27
update qt cmake script to use the freetype font engine on Windows

mikelow:
This adds an install-time patch to qt.conf on Windows to ensure we add

```
[Platforms]
WindowsArguments = fontengine=freetype
```

On my Windows 10 machine, using the default fontengine (directwrite) results in obvious aliasing within HexView when the DPR is fractional (specifically when display scale is set to 150% or 175%). This occurs regardless of the fact we round the "high dpi scale factor" via `setHighDpiScaleFactorRoundingPolicy()`, though I intend to remove this line in a separate PR anyway.

@sykhro when you're available (no rush), I'd like to get info from you about your experience with freetype rendering on Windows, since what you previously wrote about this conflicts with what I'm seeing.

My experience is that rendering at fractional DPRs is working well on every platform except for Windows. On the Linux side, I'm testing with Ubuntu, which uses Wayland.

## How Has This Been Tested?
I built and installed locally with CMake and verified the qt.conf file was amended. It reads as follows:

```
[Paths]
Prefix = .
Libraries = .
Binaries = .

[Platforms]
WindowsArguments = fontengine=freetype
```

## Screenshots
Here's an example of aliasing at 150% scaling when directwrite is used.
image

Here's how it renders at 150% scaling with freetype:
image

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/CMakeLists.txt
+++ b/src/ui/qt/CMakeLists.txt
@@ -278,6 +278,26 @@ if (NOT FLATPAK)
       DEPLOY_TOOL_OPTIONS ${DEPLOY_EXTRA_OPTIONS}
   )
   install(SCRIPT ${deploy_script})
+
+  # On Windows, ensure qt.conf sets WindowsArguments = fontengine=freetype after deployment.
+  if(WIN32)
+    file(GENERATE
+            OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/patch-qtconf-$<CONFIG>.cmake"
+            CONTENT [==[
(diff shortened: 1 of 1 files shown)


sykhro:
> Given all the issues you mentioned with FreeType, I did more experimentation trying to render with DirectWrite in HexView sans aliasing. I had some success. The results don't look quite as good as they did with FreeType, but it's much improved. I'll close this PR and make a new one for the fix.

Thanks for looking into that! I was happening to be testing it on a real hidpi Win11 machine yesterday, there's a couple commits from https://github.com/vgmtrans/vgmtrans/commits/hidpi-fixes/ that you might want to cherry-pick

PR #777, opened 2026-03-07
Bundle the current changelog in release artifacts

sykhro:
## Description
- Add `CHANGELOG.md` based on https://keepachangelog.com/
- Add a small python script to extract a given changelog section (version) as either trimmed markdown or xml for appstream
- Add `INSTALL_CHANGELOG` as an opt-in flag in CMake, enable it on CI
  - bundles `CHANGELOG.md` in the release archive
  - fills `<release>` in `metainfo.xml`

## Motivation and Context
I was a bit bothered by Flatpak not including release notes (me? OCD? nahh) and figured this would also be useful in a future CI pipeline to automate releases (the idea being finalize notes -> tag vX.Y -> push tags -> CI does the rest).
This would also encourage us to add items to release notes as part of PRs and not forget about new features/fixes when it's release time.

## How Has This Been Tested?
- `python3 build-aux/build-changelog.py --input CHANGELOG.md --format markdown`
- `python3 build-aux/build-changelog.py --input CHANGELOG.md --format appstream`

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,7 +94,7 @@ jobs:
 
       - name: Generate artifact attestation
         uses: actions/attest@v4
-        if: ${{ github.event_name != 'pull_request' }}
+        if: ${{ github.event_name == 'push' }}
         with:
           subject-name: VGMTrans-${{ github.sha }}-${{ env.BRANCH }}-${{ matrix.target }}-${{ runner.os }}
           subject-path: "build/macos-${{ matrix.target }}/VGMTrans-${{ github.sha }}-${{ env.BRANCH }}-${{ matrix.target }}-${{ runner.os }}.dmg"
@@ -154,7 +154,7 @@ jobs:
 
       - name: Generate artifact attestation
         uses: actions/attest@v4
-        if: ${{ github.event_name != 'pull_request' }}
+        if: ${{ github.event_name == 'push' }}
         with:
           subject-name: VGMTrans-${{ github.sha }}-${{ env.BRANCH }}-x86_64-${{ runner.os }}
           subject-path: "VGMTrans-${{ github.sha }}-${{ env.BRANCH }}-x86_64-${{ runner.os }}.flatpak"
@@ -212,6 +212,7 @@ jobs:
           # Force to have a flat install package layout.
           # Windows users get frightened at the sight of GNU directories.
           cmake --preset win-${{ matrix.target }} `
+            -DINSTALL_CHANGELOG=ON `
             -DQT_DEBUG_FIND_PACKAGE=${{runner.debug == '1' && 'ON' || 'OFF' }} `
             -DCMAKE_VERBOSE_MAKEFILE=${{runner.debug == '1' && 'ON' || 'OFF' }} `
             -DCMAKE_FIND_DEBUG_MODE=${{runner.debug == '1' && 'ON' || 'OFF' }} `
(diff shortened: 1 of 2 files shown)


build-aux/build-changelog.py
```diff
@@ -0,0 +1,260 @@
...
+    while section_lines and not section_lines[-1]:
+        section_lines.pop()
+
+    output_lines.extend(section_lines)
+
+    body = "\n".join(output_lines).rstrip()
+    reference_lines = collect_references(body, references)
+    if reference_lines:
+        body = f"{body}\n\n" + "\n".join(reference_lines)
+    return f"{body}\n"
```
mikelow:
```suggestion
    
    full_changelog_section = (
        "## Full changelog\n"
        "- See the full version history "
        "[here](https://github.com/vgmtrans/vgmtrans/blob/master/CHANGELOG.md)."
    )
    return f"{body}\n\n{full_changelog_section}\n"
```

CHANGELOG.md
```diff
@@ -0,0 +1,25 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
+
```
mikelow:
```suggestion
```

mikelow (changes requested):
I have some conflicting thoughts on this. First, I think it's a good idea for us to maintain a changelog. While making it easily accessible to users is important, I think packaging it in the macOS dmg isn't a great idea. I'm also ambivalent about the net benefit of its inclusion on other platforms. We're complicating our deployment process a little bit. The  python script that may need to be maintained and we're adding a mandatory step (adding an entry to CHANGELOG.md matching the git tag to be).

I think we should make the full changelog accessible in the app as a menu action. It can be as simple as opening up the URL to our github changelog in a browser. This need not be mutually exclusive with packaging the CHANGELOG.md file, but maybe it should be?

Going back to macOS - our dmg layout is already complicated, and we have an unconventional installation process that requires users to follow instructions which are provided via the layout and README.txt. I think anything extraneous here is a bad idea, as it can make the dmg overwhelming and increase the chance users ignore our instructions.

If we do keep the script and package the changelog with other platforms, I suggest we add a section at the bottom that links to the full changelog page. I've added a code suggestion for implementing that.

Second, this may be a personal quirk, but I really don't like the preamble in the CHANGELOG file - the part that reads:

```
All notable changes to this project will be documented in this file.
The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
```

I know this is a template, and that's part of why I don't like it. The first sentence is obvious. The second is a plug. Keep a Changelog strikes me as a self-important "standard" for something not in need of one.

That said, I'm fine with keeping track of our pending changes via an "Unreleased" section at the top.

sykhro:
> If you mean automatically generating release notes, I think we could copy over the changelog info for sure, but I'd still want to add a summary.

Right, I didn't think of that.
So I guess the flow would be:
- finalize release notes for vX.Y, committing a summary to changelog.md or something
- tag vX.Y, push
- CI compiles the whole thing (tags are workflow triggers), creates a new release for vX.Y copying over the relevant changelog part via this script and attaches the compiled artifacts

mikelow:
> So I guess the flow would be:
> - finalize release notes for vX.Y, committing a summary to changelog.md or something

I may be confused. My understanding was that we simply rename the [Unreleased] section to whatever tag we are about to push (eg. v1.4). The summary would be just part of the github release. Do I misunderstand?

mikelow (approved):

sykhro:
> > You are right, if we're adding a summary to the file before tagging it might as well rename [Unreleased] to the tag right then and there. whoops
> 
> I meant that the summary isn't even part of the changelog. Maybe it should be though? I guess I thought of them as separate because the summary is less formal - more of a "this is what we've got going on in this release." But I don't feel strongly either way.

In that case the CI job can create the release as a draft, so we can manually insert a preamble before it goes live

PR #781, opened 2026-03-14
CPS2/CPS3: fix vibrato and tremelo

mikelow:
At some point before V1.3, I pushed a change that always skips the vibrato and tremelo effect pipeline in the CPS2/3 driver. It's an unnecessary version check, but also compares an enum type against the value of a different enum type - so it's nonsensical as written.

## How Has This Been Tested?
Verified vibrato is working again.

## Types of changes

## Checklist:


sykhro (approved):
This is why I like strong types/enum class

PR #788, opened 2026-03-15
Replace ClickJumpSlider with SeekBar widget

mikelow:
On macOS, we're seeing a major performance hit from using QSlider as a playback seek bar. On an M1, it's using about 30% of a performance core during playback - over 6 times what BASS typically uses for processing audio. This has a noticeable effect on draw performance in the upcoming PianoRollView. The performance hit is directly proportional to the rate at which QSlider's value is updated.

To solve this, I've replaced our QSlider subclass (ClickJumpSlider) with a simple QWidget subclass called SeekBar. This new class acts like a QSlider but with better draw performance. When its value is changed, it checks whether it needs to redraw, and if so, invalidates a minimal rect. The result is a major reduction in CPU usage. The extent of that reduction depends on the length of the bar and how fast the thumb moves, but in worst cases I'm seeing roughly 1/3 usage compared to before, and in typical cases around 5%-10% relative to before.

The new class caches both the track bar and circular thumb into pixmaps for better draw performance. The thumb is drawn with a drop shadow effect, so caching is especially important for it.

Additionally, I've updated IconBar to only update the SeekBar's value every other position change signal from SequencePlayer when the origin of that signal is `Playback`. This means we update at about 30hz. This reduces CPU usage and still feels quite smooth.

It seems neither Windows or Linux exhibit the performance issue. QWidget draw performance on both platforms seems to be significantly faster than on MacOS, which was evident with pre-RHI HexView. Regardless, I think it makes sense to use SeekBar across all platforms, both for consistency, and because I think the Windows and Linux QSliders are ugly. Sure, it looks more MacOS than native, but I think that's a good thing in this case (and most others).

## How Has This Been Tested?
Tried and failed to break it.

## Screenshots:

Windows light:

Capture-new-win-light

Windows dark:

Capture-new-win-dark

MacOS light:

Screenshot 2026-03-15 at 2 30 28 PM

MacOS dark:

Screenshot 2026-03-15 at 2 30 51 PM

Ubuntu light:

ubuntu-light

Ubuntu dark:

ubuntu-dark

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/widgets/SeekBar.cpp
+++ b/src/ui/qt/widgets/SeekBar.cpp
@@ -165,44 +165,53 @@ void SeekBar::showEvent(QShowEvent* event) {
 }
 
 void SeekBar::invalidatePixmaps() {
-  m_pixmapsDirty = true;
+  m_trackPixmapDirty = true;
+  m_thumbPixmapDirty = true;
 }
 
 void SeekBar::ensurePixmaps() {
   const qreal dpr = devicePixelRatioF();
-  const QSize logicalSize = size();
-  if (!m_pixmapsDirty && m_cachedPixmapSize == logicalSize && qFuzzyCompare(m_cachedPixmapDpr, dpr)) {
+  const QRectF track = trackRect();
+  const QSize logicalTrackSize = track.size().toSize();
+  const QSize trackPixelSize = QSizeF(track.width() * dpr, track.height() * dpr).toSize();
+
+  // Redraw the track pixmap if necessary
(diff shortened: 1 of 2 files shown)


src/ui/qt/widgets/SeekBar.cpp
```diff
@@ -0,0 +1,327 @@
...
+    return track.left();
+  }
+
+  const qreal ratio = static_cast<qreal>(value - m_minimum) /
+                      static_cast<qreal>(m_maximum - m_minimum);
+  return track.left() + ratio * track.width();
+}
+
+int SeekBar::displayedThumbStep(int value) const {
+  // Quantize the rendered thumb position to half-device-pixel steps.
```
sykhro: is the comment stale? I'm not sure I 100% understand what's happening here

PR #802, opened 2026-03-30
New Layout - Part 4: Improved dock behavior

mikelow:
This PR makes the main dock layout persistent and gives `VGMCollView` its own dock instead of grouping it with `VGMCollListView`.

The docks can now be shown, hidden, floated, rearranged, and restored across launches. The `View` menu also adds `Reset Dock Layout` so the user can get back to the default arrangement.

**Details**

- `Collection Contents` is now a separate dock, independent from `Collections`.
- The bottom dock area is now a horizontal dock row instead of a tabified area.
- The `View` menu lists the dock visibility actions directly and adds `Reset Dock Layout`.
- The window bar now includes a toggle button for `Collection Contents`.

The app now saves and restores:
- main window geometry
- Qt dock layout state
- floating dock geometry

The dock layout code also remembers the last intended width of the left dock column and the last intended height of the bottom dock row. If the window is shrunk and Qt temporarily compresses those areas, enlarging the window restores the remembered sizes instead of treating the compressed sizes as the new preferred layout.

`Collection Contents` has additional handling because it can move between the left stack and the bottom row. Its height is remembered while it sits at the bottom of the left stack. When the bottom dock row is reintroduced, that remembered height is applied to the bottom area so the transition back into the row is seamless instead of snapping to an unrelated height.

This PR also adds explicit placement rules for `Collection Contents`:
- If it is the only dock left in the bottom area, it is moved back into the left stack.
- If `Collections` or `Logs` are visible in the bottom area, it is moved into that row with them.
- When it spans the left column and the bottom row, its width is locked to the left-column width so the edges stay aligned.

The dock-specific behavior is handled by `MainWindowDockLayout`, which owns dock restoration, remembered sizes, placement rules, and deferred reconciliation after Qt finishes rearranging the layout.

The PR also removes the old `IconBar` and its unused playback SVG assets, adds settings support for saved dock state, removes the awkward `Preview` menu path for collection playback (play / pause buttons suffice), and includes a small macOS `openURL` ownership fix.

## Types of changes

## Checklist:


sykhro:
I will try it again tomorrow. also, friendly changelog entry reminder!

mikelow:
> is the shadow on menus expected to look like that? also, I can't screenshot it easily, but hovering on items results in a rectangle clipping through the rounded corners

The default QMenu style for Windows 11 has a number of issues (bad colors, not enough padding, ugly drop shadow), and setting a stylesheet to fix them led to even more complications. I've now fixed these issues by using a QProxyStyle specifically for Windows 11. I'll test how it works on Windows 10.

There's another bug with the dock I discovered that I also need to fix - detaching the collections dock by dragging its title bar leads the Content Collection dock in a bizarre state.


> friendly changelog entry reminder!

Ah yes, I'll address that too.

mikelow:
Ok, the code has been updated. In addition to fixing menu appearance in Windows 11, I've fixed a number of issues with docking behavior, specifically:
- the Collection Contents dock getting into a broken state when transferring to the left anchor area by closure of the bottom anchor area if the bottom dock that was removed became floating via a title bar drag
- the app window and docks sometimes resizing when floating docks were restored to a docking area

The MainWindowDockLayout class is basically a big workaround for the shortcomings of Qt's dock implementation. It's big and messy, but I think unavoidably so.

I will test whether to enable the menu changes for Windows 10, but I'm not planning further code changes in this branch otherwise.

sykhro:
I noticed the qmsetup thing is back, is that intentional?

mikelow:
> I noticed the qmsetup thing is back, is that intentional?

Nope! Must have reverted it via some force push. Not good.

sykhro:
> The MainWindowDockLayout class is basically a big workaround for the shortcomings of Qt's dock implementation. It's big and messy, but I think unavoidably so.

Agreed. It seems like qwidgets are starting to be broken in many places, see for example the "selected item" with the win11 theme

image

that highlight definitely shouldn't be per-cell... and it's very questionable to begin with, in a table.

In the future, I might take a stab at making a better "modern windows" theme (kinda like phantomstyle was a better fork of Fusion)

sykhro:
> > The MainWindowDockLayout class is basically a big workaround for the shortcomings of Qt's dock implementation. It's big and messy, but I think unavoidably so.
> 
> Agreed. It seems like qwidgets are starting to be broken in many places, see for example the "selected item" with the win11 theme
> image
> 
> that highlight definitely shouldn't be per-cell... and it's very questionable to begin with, in a table.
> 
> In the future, I might take a stab at making a better "modern windows" theme (kinda like phantomstyle was a better fork of Fusion)

it's also funny how the left side of the selected item has rounded corners but the right one is perfectly square

mikelow:
> Agreed. It seems like qwidgets are starting to be broken in many places, see for example the "selected item" with the win11 theme

Yeah, it's pretty sloppy. I think the overall look is much better than fusion, though.

sykhro:
> > Agreed. It seems like qwidgets are starting to be broken in many places, see for example the "selected item" with the win11 theme
> 
> Yeah, it's pretty sloppy. I think the overall look is much better than fusion, though.

I agree, fusion looks kinda dated at this point

mikelow:
@sykhro let me know if there are any outstanding issues I've missed. I've addressed the highlighting issue in my [newer PR](https://github.com/vgmtrans/vgmtrans/pull/805).

sykhro (approved):
I think we're good to go. I would still like to improve icons, but that can come later

Issue #178, opened 2020-09-04
SuzukiSnes doesn't map percussion instruments


ElSaico:
**Describe the bug**
When playing audio files from Super Mario RPG, which uses the Square/Suzuki driver, several of them sound strange.

After some investigation, it turns out this happens when one or more tracks has a Percussion On instruction ($EE) instead of Program Change ($DE xx); presumably, they're not mapped out to the proper instrument(s).

**To Reproduce**
1. Load a .spc file with at least one track with a Percussion On instruction
2. Play

**Expected behavior**
All tracks should play correctly.

**Desktop (please complete the following information):**
 - OS: macOS Mojave 10.14.6
 - VGMTrans version: c3e122
 - Game ROM or music dump SHA-256

This happens with at least the following files:

5d43f46757cd9e53fce97a21e16dfba55f8ebea2dc2d512f34c69719253a8223  smr-107.spc (Super Pipe House)
b0d6e3b1528f40912641d0e163ff8c94dd7e930533cb69b10381dfced15470a1  smr-110.spc (Fight Against Monsters)
edd24489d49207225076f44beb721abad7df461ac8c98759d9bd275a23e06bc1  smr-119.spc (Fight Against an Armed Boss)
d9519d73410c3f24ff3f612934e96dd024c977f2732afd1b7b75b21a7b04f494  smr-207.spc (Fight Against Culex)

loveemu:
You're right. It simply hasn't been implemented yet.

In fact, there is no SoundFont conversion implementation for drum instruments in almost all SNES formats. Many formats currently output percussion notes to the MIDI 10 channel (which I don't like very much, as it creates complications). These notes are played silently because there is no Bank 127 instrument in SoundFont.

However, there are a few exceptions. Hudson and Square/Hidenori Suzuki (SMRPG) output the percussion notes to their own channels. In this case, the instruments in bank 0 are used, so the percussion notes will not be silent, but will turn into something noisy.

ElSaico:
*A four-year-old-update!*

I found out [a SMRPG editor with full audio support](https://github.com/Yakibomb/LAZYSHELL-UPDATED), which means it has to know how to parse the percussion data.

My findings: it seems to work similarly to the Chrono Trigger driver (the only Akao variant with a drum kit), and is none other than that list of 5-byte "unknowns" at the beginning of the header.
They mean, respectively, the note index, sample number, key, volume and pan.

If everything goes well, I plan to cook up a PR in the next few days.

PR #807, opened 2026-04-04
SuzukiSnes: Percussion support

ElSaico:
Reads percussion data on SPC files which use the Hidenori Suzuki driver (late Square titles, e.g. Super Mario RPG).

## Description
There is already a Square driver with dedicated percussion support (the Chrono Trigger iteration of AkaoV4), so it was assumed they work similarly. As [the Lazy Shell tool](https://github.com/Yakibomb/LAZYSHELL-UPDATED) features a sound editor, the data format could be deduced from it and a solution devised with ~a blatant copy/paste job over~ AkaoSnes as a basis.

## Motivation and Context
The lack of support makes any attempted conversions not only incomplete but outright _grating_ to listen to, which made me open #178 a few years ago.

## How Has This Been Tested?
With the Super Mario RPG and Bahamut Lagoon soundtracks from snesmusic.org (Trials of Mana was also used as a sanity check to ensure things don't break, as it lacks a percussion program).

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## Checklist:
- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.

### DIFF
--- a/src/main/formats/SuzukiSnes/SuzukiSnesInstr.cpp
+++ b/src/main/formats/SuzukiSnes/SuzukiSnesInstr.cpp
@@ -19,13 +19,15 @@ SuzukiSnesInstrSet::SuzukiSnesInstrSet(RawFile *file,
                                        uint16_t addrVolumeTable,
                                        uint16_t addrADSRTable,
                                        uint16_t addrTuningTable,
+                                       uint16_t addrDrumKitTable,
                                        const std::string &name) :
     VGMInstrSet(SuzukiSnesFormat::name, file, addrSRCNTable, 0, name), version(ver),
     spcDirAddr(spcDirAddr),
     addrSRCNTable(addrSRCNTable),
     addrVolumeTable(addrVolumeTable),
     addrADSRTable(addrADSRTable),
-    addrTuningTable(addrTuningTable) {
+    addrTuningTable(addrTuningTable),
+    addrDrumKitTable(addrDrumKitTable) {
 }
 
 SuzukiSnesInstrSet::~SuzukiSnesInstrSet() {}
@@ -90,6 +92,11 @@ bool SuzukiSnesInstrSet::parseInstrPointers() {
     return false;
   }
 
+  if (addrDrumKitTable && readShort(addrDrumKitTable) >= 0x80) {
+    SuzukiSnesDrumKit *newDrumKitInstr = new SuzukiSnesDrumKit(this, version, DRUMKIT_PROGRAM, spcDirAddr, addrTuningTable, addrADSRTable, addrDrumKitTable, "Drum Kit");
+    aInstrs.push_back(newDrumKitInstr);
+  }
+
   std::sort(usedSRCNs.begin(), usedSRCNs.end());
   SNESSampColl *newSampColl = new SNESSampColl(SuzukiSnesFormat::name, this->rawFile(), spcDirAddr, usedSRCNs);
   if (!newSampColl->loadVGMFile()) {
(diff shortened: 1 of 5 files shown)


ElSaico:
After wasting too many hours by misunderstanding the data (the byte I thought to mean "source" was actually "base instrument"), it's functional.

There are *definitely* things that warrant improvement here - the cymbals going overboard in Fight Against Culex, for one - but it's a joy to the ears already!

mikelow:
Great job and thank you. This definitely fixes percussion. Could you please provide more technical explanation, both of how the percussion data works and anything notable about your implementation? It doesn't need to be long, but context would make this easier to review.

I haven't looked closely yet, but I think we'll want to rename those header entries from "Percussion" to something more descriptive (I assume they're instrument definitions). It would be good to give them child items so we have a break down of what data they provide.

ElSaico:
> Great job and thank you. This definitely fixes percussion. Could you please provide more technical explanation, both of how the percussion data works and anything notable about your implementation? It doesn't need to be long, but context would make this easier to review.

So, the Lazy Shell editor knows how to parse this data, but I could not find anything on that source WRT actually *playing* it. I then worked from the hypothesis of an evolution from Akao's implementation in Chrono Trigger, explicitly modeling my solution after `AkaoSnesDrumKit` / `AkaoSnesDrumKitRgn`, and that gamble paid off in the end.

* `AKAOSNES_V4_CT`
  * starts in `$F120`
  * fixed size, indexed by note
  * (instrument, key, pan)
* `SUZUKISNES_BL` / `SUZUKISNES_SMR` (was scarcely used in BL and not at all in SD3)
  * starts in `$2000`, prepending the track pointer list
  * variable size, index thus not relevant
  * (note, instrument, key, volume, pan)

> I haven't looked closely yet, but I think we'll want to rename those header entries from "Percussion" to something more descriptive (I assume they're instrument definitions). It would be good to give them child items so we have a break down of what data they provide.

They are properly defined in `SuzukiSnesInstrSet`, instead, so it might be a matter of a) remove those entries from the header definition or b) assign everything to a single "Percussion Data" blob.

ElSaico:
As an aside: while playing CT songs to see how they work, it caught my attention that "Burn! Bobonga!" is sounding a bit... *off*.

Further investigation pointed to a track which uses instrument 0, so I know where to start looking.

mikelow:
I've looked into this a little further. I think there’s an ownership/coupling issue here.

As I understand the format, the percussion table at the start of the sequence header is sequence-local data: it defines a drum note map for that particular song, using the instrument set as a building block. In other words, it looks more like “for this sequence, note N uses base instrument I with key/volume/pan overrides” than like a drum kit that is inherent to the instrument set itself.

With the current implementation, we pass `addrSeqHeader` into `SuzukiSnesInstrSet`, synthesize a `Drum Kit` inside the instr set from that header data, and then have the sequence switch to and from that synthetic instrument on `PERC_ON`/`PERC_OFF` events. That means the instrument set is now coupled and partly defined by one particular sequence’s header. 

I think that's the wrong boundary. We should not pass `addrSeqHeader` as a parameter to the instrument set.

If two sequences shared the same Suzuki instrument tables but had different percussion header tables, this design would imply a single canonical drum kit on the instrument set when the mapping is actually sequence-specific. This may never happen in practice, but the format is making this distinction.

So let's keep percussion mapping owned by the sequence and leave the instrument set untouched. It should just contain base instruments, no drumkit. That means when the percussion flag is set, the sequence will have to look to the drumkit table and make program, volume and pan changes on the fly - not exactly elegant, I know, but we have other formats that do similar.

There is another solution I would like to eventually implement, but it will require changes in the core code - allowing the VGMInstrSet to inspect the VGMSeq, but only at conversion time. We already can do this for sequences: the VGMSeq class has a `useColl()` method that gets called at conversion, which allows the VGMSeq to read data from any file in its collection to affect the conversion output while not changing the originally parsed data and child items. I think we could more elegantly solve this drumkit issue by adding a `useColl()` method to VGMInstrSet so it could create a drumkit instrument during conversion using data from the VGMSeq its associated with. 

However, adding `useColl()` to VGMInstrSet won't be so simple, because instrument set conversion is more complicated than sequence conversion.

For now, let's just stick to keeping drumkit data confined to SuzukiSnesSeq.

src/main/formats/SuzukiSnes/SuzukiSnesSeq.cpp
```diff
@@ -53,12 +54,12 @@ bool SuzukiSnesSeq::parseHeader() {
...
       uint8_t firstByte = readByte(curOffset);
       if (firstByte >= 0x80) {
-        header->addChild(curOffset, 1, "Unknown Items End");
+        header->addChild(curOffset, 1, "Percussion End");
         curOffset++;
         break;
       }
       else {
-        header->addUnknownChild(curOffset, 5);
+        header->addChild(curOffset, 5, "Percussion");
```
mikelow: We should add more child items here. This is what I meant by my earlier comment:

```
        auto drum = header->addChild(curOffset, 5, "Drum Instr");
        drum->addChild(curOffset, 1, "Key");
        drum->addChild(curOffset+1, 1, "Sample Index");
        drum->addChild(curOffset+2, 1, "Unity Key");
        drum->addChild(curOffset+3, 1, "Volume");
        drum->addChild(curOffset+4, 1, "Pan");
```

I may have gotten the ordering and meaning wrong there.

ElSaico:
> So let's keep percussion mapping owned by the sequence and leave the instrument set untouched. It should just contain base instruments, no drumkit. That means when the percussion flag is set, the sequence will have to look to the drumkit table and make program, volume and pan changes on the fly - not exactly elegant, I know, but we have other formats that do similar.

As much as I like this idea for the simplification factor, consider:

Captura de tela 2026-04-06 173938
(yes this is on stable, I'm not on my desktop and opened it just for display)

This is a percussion track (see the `$EE` at the very beginning) yet every other note is preceded by a Volume instruction (`$E2 xx`); roughly 2/3 of its bytecode would be pointless in case of a naked override.

mikelow:
> As much as I like this idea for the simplification factor, consider:
> This is a percussion track (see the `$EE` at the very beginning) yet every other note is preceded by a Volume instruction (`$E2 xx`); roughly 2/3 of its bytecode would be pointless in case of a naked override.

The idea isn't to override volume events. We could, for example, use an expression event to modify volume without touching the volume controller.

However, while an approach like this could work, it would be more complicated than I initially thought. Since the drum data is presumably overriding the base instrument data, we would need to emulate the volume and pan override using sequence events, which would either have to account for the original instrument values, or we could have the instrument set export duplicate instruments with default volume and pan values.

I'd rather not do that. It would make more sense to just fix this the proper way, and I wouldn't expect you to jump through those hoops anyhow. So here's what I propose:
- implement the feedback I suggested around adding child items to the drum items. Clarify if you understand each drum to be pointing to a base instrument index or a just a sample index.
- I'll merge the PR
- I'll implement a means for VGMInstrSet to read VGMSeq data on conversion and create temporary instruments for conversion
- I'll update this driver to use that feature instead of passing `addrSeqHeader` as a parameter to the instr set.

mikelow:
@codex review

src/main/formats/SuzukiSnes/SuzukiSnesInstr.cpp
```diff
@@ -138,55 +146,165 @@ bool SuzukiSnesInstr::loadInstr() {
...
+  VGMInstr(instrSet, addrDrumKitTable, 0, programNum >> 7, programNum & 0x7F, name), version(ver),
+  spcDirAddr(spcDirAddr),
+  addrSRCNTable(addrSRCNTable),
+  addrTuningTable(addrTuningTable),
+  addrADSRTable(addrADSRTable),
+  addrDrumKitTable(addrDrumKitTable) {
+}
+
+SuzukiSnesDrumKit::~SuzukiSnesDrumKit() {
+}
```
sykhro, line 182: Prefer `~SuzukiSnesDrumKit() = default;` in the header

src/main/formats/SuzukiSnes/SuzukiSnesScanner.cpp
```diff
@@ -209,17 +209,19 @@ void SuzukiSnesScanner::searchForSuzukiSnesFromARAM(RawFile *file) {
   uint16_t addrVolumeTable;
   uint16_t addrTuningTable;
   uint16_t addrADSRTable;
+  uint16_t addrDrumKitTable;
   if (file->searchBytePattern(ptnLoadInstr, ofsLoadInstr)) {
     addrSRCNTable = file->readShort(ofsLoadInstr + 5);
     addrVolumeTable = file->readShort(ofsLoadInstr + 10);
     addrADSRTable = file->readShort(ofsLoadInstr + 18);
     addrTuningTable = file->readShort(ofsLoadInstr + 30);
+    addrDrumKitTable = version != SUZUKISNES_SD3 ? addrSeqHeader : 0;
```
mikelow, line 218: SD3 also has an optional drumkit definition in its sequence header. The difference is that it comes after the track pointers instead of before.

Changing this line to:

```suggestion
    addrDrumKitTable = addrSeqHeader + (version == SUZUKISNES_SD3 ? 0x10 : 0);
```

Fixes drums in those tracks. However, the logic for adding UI Items needs to be updated as well, as the DrumKit isn't added into the tree view.

mikelow (changes requested):

mikelow:
I'm just going to merge this and address the remaining issues in a follow-up. It will save me some inconvenience for other work that would need to be rebased.

Issue #626, opened 2025-07-15
VGM Trans won't open spc files exported from snes9x


PhantomAvenger:
### Environment details

* Operating System+version: Windows 11
* VGMTrans version: both 1.2 and Latest Nightly
* Game ROM/music dump name and sha256: Super Metroid spc dump

### Describe the bug

I'm exporting spc data from super metroid using the snes9x emulator, but whenever i try to open it with vgmtrans it just doesn't do anything. I've already checked with other spc players, and they are working as intended, so the issue might be in the program itself.

### Steps to reproduce

1. export the spc files following the proper instructions
2. try to open them in vgmtrans

### Expected behaviour

the vgmtrans program should be able to read them and reproduce them in the player as normal

### Logs

[log.log](https://github.com/user-attachments/files/21237575/log.log)

KungFuFurby:
This issue should have been closed with pull request #808... but the text was malformed and GitHub accidentally forgot to close it. Requesting a manual closing of the issue.

mikelow:
Thanks for the heads up

PR #808, opened 2026-04-04
SPCFile.cpp: fix detection of ID666 tags

### LINKED ISSUE #610: SFX Discovery
### Environment details

* Operating System+version: **Windows 10 - Unsure of Version**
* VGMTrans version: **Latest Nightly**
* Game ROM/music dump name and sha256: N/A - Self Dump

### Describe the bug

I noticed why some snes songs are not loading, especially for custom SPCs/SFX SPCs
Checking logs, I noticed this error in the log
"invalid stoi argument"

Forgive me for this, my first time actually making a bug report here

### Steps to reproduce

1. Load any self Dumped SPC, or any spc from SMWCentral
2. SPC will fail to load, error is logged

### Expected behaviour

SPC should load

### Additional context

_No response_

### Logs

[error.log](https://github.com/user-attachments/files/20562643/error.log)

<details><summary>Click to expand log</summary>

```
invalid stoi argument
```

</details>

CBongo:
The check for ID666 tags was checking for nonzero value instead of 0x1a; a value of 0x1b indicates no tags, which resulted in std::stoi throwing an exception when reading the  secondsToPlay and fadeLength.  Fixed the check for tags, and wrapped the std::stoi calls with try/catch to protect against the unexpected and allow slightly malformed files to load anyway.

Closes #610 and #626

### DIFF
--- a/src/main/components/SPCFile.cpp
+++ b/src/main/components/SPCFile.cpp
@@ -28,7 +28,8 @@ SPCFile::SPCFile(const RawFile& file) {
     throw std::runtime_error("Invalid SPC signature");
   }
 
-  m_hasID666Tag = file.readByte(0x23);
+  m_hasID666Tag = file.readByte(0x23) == 0x1a;
+  L_DEBUG("SPC file has ID666 tag? {}", m_hasID666Tag);
   m_versionMinor = file.readByte(0x24);
   file.readBytes(0x100, m_ram.size(), m_ram.data());
   file.readBytes(0x10100, m_dspRegisters.size(), m_dspRegisters.data());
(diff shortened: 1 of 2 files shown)


src/main/components/SPCFile.cpp
```diff
@@ -28,7 +28,8 @@ SPCFile::SPCFile(const RawFile& file) {
     throw std::runtime_error("Invalid SPC signature");
   }
 
-  m_hasID666Tag = file.readByte(0x23);
+  m_hasID666Tag = file.readByte(0x23) == 0x1a;
+  L_DEBUG("SPC file has ID666 tag? {}", m_hasID666Tag);
```
sykhro: I would argue there is not much value in keeping this debug line

sykhro (changes requested):
Thank you for your contribution! I have one minor comment

Bonboon229:
A longterm issue solved at last!

src/main/components/SPCFile.cpp
CBongo: Agreed; removed.  Do you think there would be value in logging something at DEBUG or INFO only in the case where there is no tag?  If so this would be a good time to add it; if not then we should be good to go.  After writing it down I'm leaning towards "no" myself...

sykhro (approved):

mikelow:
Thanks for the fix!

src/main/components/SPCFile.cpp
sykhro: I would say no. Tags are not something that sensitive so I wouldn't bother with logging details of parsing!

PR #812, opened 2026-04-08
CI: add CLI and shell Windows builds

Inklay:
Added cli and shell builds on CI

## Description
The only easily accessible windows build is the GUI one, I updated the CI to also build the shell and CLI versions to make them easier to get online

## Motivation and Context

I needed a Windows CLI build and couldn't find one without having to build it myself

## How Has This Been Tested?

I ran the pipeline in my fork and it the build stage worked and I got the 3 .exe in the artifact's zip

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -250,6 +250,8 @@ jobs:
           # Windows users get frightened at the sight of GNU directories.
           cmake --preset win-${{ matrix.target }} `
             -DINSTALL_CHANGELOG=ON `
+            -DINSTALL_CLI=ON `
+            -DINSTALL_SHELL=ON `
             -DQT_DEBUG_FIND_PACKAGE=${{runner.debug == '1' && 'ON' || 'OFF' }} `
             -DCMAKE_VERBOSE_MAKEFILE=${{runner.debug == '1' && 'ON' || 'OFF' }} `
             -DCMAKE_FIND_DEBUG_MODE=${{runner.debug == '1' && 'ON' || 'OFF' }} `
@@ -262,7 +264,7 @@ jobs:
          run: |
            cmake --build `
             --preset win-${{ matrix.target }}-release `
-            --target vgmtrans vgmtrans-shell
+            --target vgmtrans vgmtrans-cli vgmtrans-shell
 
        - name: "Package project"
          run: |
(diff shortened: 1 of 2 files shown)


sykhro:
vgmtrans-cli is due for removal.
use vgmtrans-shell instead

.github/workflows/build.yml
```diff
@@ -250,6 +250,8 @@ jobs:
           # Windows users get frightened at the sight of GNU directories.
           cmake --preset win-${{ matrix.target }} `
             -DINSTALL_CHANGELOG=ON `
+            -DINSTALL_CLI=ON `
```
sykhro, line 253:
```suggestion
```

.github/workflows/build.yml
```diff
@@ -262,7 +264,7 @@ jobs:
          run: |
            cmake --build `
             --preset win-${{ matrix.target }}-release `
-            --target vgmtrans vgmtrans-shell
+            --target vgmtrans vgmtrans-cli vgmtrans-shell
```
sykhro, line 267:
```suggestion
            --target vgmtrans vgmtrans-shell
```

src/ui/cli/CMakeLists.txt
```diff
@@ -6,6 +6,11 @@ target_sources(vgmtrans-cli
...
 target_include_directories(vgmtrans-cli PUBLIC "${PROJECT_BINARY_DIR}/src")
-
 target_link_libraries(vgmtrans-cli PRIVATE vgmtranscore)
 target_compile_features(vgmtrans-cli PRIVATE cxx_std_20)
+
+if(INSTALL_CLI)
+  install(TARGETS vgmtrans-cli
+          RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
+  )
+endif()
```
sykhro, line 16:
```suggestion
```

sykhro (changes requested):
we never really supported `vgmtrans-cli` and are planning to remove it soon. 
as for `vgmtrans-shell`, I'm not sure I see the value in end users having it. it's mostly a dev tool... if we do end up distributing it, it definitely needs adding to changelog. and then we have to double check packaging, since we don't want the flatpak to end up with both.

so, I have to ask, what makes you need _prebuilt_ `vgmtrans-shell`?

mikelow:
> as for vgmtrans-shell, I'm not sure I see the value in end users having it. it's mostly a dev tool... if we do end up distributing it, it definitely needs adding to changelog. and then we have to double check packaging, since we don't want the flatpak to end up with both.

I agree. I don't think we should add the shell as a CI target. There's doubtfully much demand, it would add complexity when we have plenty already, and it will probably hurt our Github Actions limits.

sykhro:
That settles it then. Thank you for your contribution regardless!

PR #821, opened 2026-04-15
HexView Refactor 3: optimize selection mask generation

mikelow:
## Description

This PR rewrites HexView’s selection and playback mask builder inside `HexViewRhiRenderer`. The older logic first collected per-line intervals for each range, then sorted and merged those intervals, then emitted mask rectangles from the merged intervals, and finally tracked matching intervals across lines to coalesce edge geometry. This branch replaces that with a single bitmask-based builder in `appendMaskForSelections()`. Each visible line is represented as a compact `uint16_t` byte mask, mask rectangles are emitted directly from those masks, and edge runs are tracked with fixed-size arrays keyed by `(startCol, endCol)` instead of dynamic maps and intermediate interval vectors.

The same change also simplifies a couple of related hot paths. Fade playback alpha overlays no longer go through the full edge-building path when they only need plain mask rectangles, and the now-unused interval and edge-run helper declarations are removed from `HexViewRhiRenderer.h`. The intent here is to reduce per-frame allocation, sorting, and bookkeeping in the selection/playback builder while keeping the existing renderer structure intact.

Future PRs will change the rendering pipeline more fundamentally while using this new mask paradigm.

## Types of changes

- [x] Refactor

## Checklist:

### DIFF
--- a/src/ui/qt/util/NonTransientScrollBarStyle.h
+++ b/src/ui/qt/util/NonTransientScrollBarStyle.h
@@ -1,0 +1,60 @@
+/*
+ * VGMTrans (c) 2002-2026
+ * Licensed under the zlib license,
+ * refer to the included LICENSE.txt file
+ */
+
+#pragma once
+
+#include <QProxyStyle>
+#include <QScrollBar>
+#include <QStyleFactory>
+
+namespace QtUi {
+
+#ifdef Q_OS_MAC
+
+class NonTransientScrollBarStyle final : public QProxyStyle {
+public:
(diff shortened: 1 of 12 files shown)


sykhro (approved):
this is cool! I'm curious to see what the compiler emits for arm64 after the other renderer changed are carried over.

mikelow:
Glad you approve! Can you explain further the curiosity around arm64? I sense I'm missing something.

(oops, edited your messages instead of replying 🤦)

PR #823, opened 2026-04-18
Add support for NinSnes drum kits

mikelow:
## High level

This updates both NinSnesInstrSet and NinSnesSeq so that the converted instrument set contains a drum kit used by percussion note events. Previously, percussion note events were just generating noise by playing arbitrary notes without a program change.

Similar to Suzuki and PS1 Akao, crucial data for the drum kit is stored in the sequence itself, so we utilize `VGMInstrSet::useColl()` and `VGMInstrSet::addTempInstr()` to allow the instrument set to access sequence data and create the drum kit at conversion time. 

## Low level 

In NinSnes sequence data, the `EVENT_PERCUSSION_PATCH_BASE` event indicates the base instrument index to use for drums. The opcode of a drum note event indicates which instrument index offset to add to this base. The driver has a lot of versions, but for most, percussion note events start with opcode 0xCA. A percussion note with opcode 0xCB would therefore use the base instrument index + 1. We assign notes for these events at base offset 0x24 (arbitrary) plus the relative instrument index. So in this example, 0xCB would use note value 0x25.

We track the assignment of drum instruments to notes in NinSnesSeq::m_percussionInstrNoteMap. We also save the global transposition value when the note was first encountered, as this affects its pitch (but track transpose does not). Using pitch bend would be messy, so we instead subtract the global transposition from the drum kit region's unity key.

With that data stored, at conversion time the NinSnesInstrSet creates a new temporary drum kit instrument and iterates over all of the instr indexes in the map to create a new region for each using the given note index, which also factors into the unity key alongside global transposition. The rest of the data for that instrument (sample number, ADSR, Gain, etc) was already parsed on load, so we just copy it over.

## Other things

I found a case of an instrument using an insane pitch scale value, which was not handled properly by VGMTrans and resulted in a very wrong unity key. Specifically, in SimCity, one of the drum instruments had a pitch scale value of 0x8000 (we call these bytes "unity key" and "fine tune", but it's a slight misnomer). This value was causing the DSP 14-bit pitch scale register to wrap. I added some logic to handle values like this. Since calculating accurate pitch in this case requires knowing the note ahead of time, the code assumes it's a percussion note (which uses the fixed pitch for note index 0x24).

Another thing: for some reason, @loveemu reassigned what would normally be the EVENT_PERCUSSION_PATCH_BASE opcode to NOP for the HAL version of the driver. I tested many HAL games and couldn't determine why he did this (it breaks drum kit conversion) so I took it out.

## Motivation and Context
This fixes percussion on a good chunk of classic (and not-so-classic) SNES games.

## How Has This Been Tested?
Tested a large number of NinSnes games, including the Quintet versions of the driver which use special logic in the EVENT_PERCUSSION_NOTE handler.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/formats/NinSnes/NinSnesInstr.cpp
+++ b/src/main/formats/NinSnes/NinSnesInstr.cpp
@@ -302,8 +302,8 @@ NinSnesRgn::NinSnesRgn(NinSnesInstr *instr,
   double coarse_tuning;
 
   // Very large NinSnes pitch multipliers can wrap the SNES DSP's 14-bit pitch instead of acting
-  // like a normal transposition. We have only observed this on percussion instruments (ex. F-Zero
-  // Big Blue), which the driver plays at fixed note 0x24 (base pitch 0x0217). When that happens,
+  // like a normal transposition. We have only observed this on percussion instruments (ex. SimCity
+  // Metropolis), which the driver plays at fixed note 0x24 (base pitch 0x0217). When that happens,
   // assume percussion usage and replace the raw multiplier with the wrapped pitch actually heard at
   // 0x24, expressed again as an equivalent scale factor for SF2/DLS unity key / fine-tune calculation.
   bool wrapsPercussion = (((uint32_t)0x0217 * pitch_scale) >> 8) > 0x3FFF;
--- a/src/main/formats/NinSnes/NinSnesInstr.cpp
+++ b/src/main/formats/NinSnes/NinSnesInstr.cpp
@@ -4,8 +4,35 @@
  * refer to the included LICENSE.txt file
  */
 #include "NinSnesInstr.h"
+#include "NinSnesSeq.h"
 #include "SNESDSP.h"
+#include "VGMColl.h"
 #include <spdlog/fmt/fmt.h>
+
+namespace {
+
+VGMRgn* cloneRgnForDrumKit(VGMInstr* instr, VGMRgn* sourceRgn, uint8_t noteIndex, int8_t globalTranspose) {
+  auto* newRgn = new VGMRgn(instr,
+                            sourceRgn->offset(),
+                            sourceRgn->length(),
+                            noteIndex,
+                            noteIndex,
+                            sourceRgn->velLow,
+                            sourceRgn->velHigh,
+                            sourceRgn->sampNum);
+
+  newRgn->sampOffset = sourceRgn->sampOffset;
+  newRgn->unityKey = sourceRgn->unityKey + (noteIndex - 0x3C) - globalTranspose;
+  newRgn->fineTune = sourceRgn->fineTune;
+  newRgn->attack_time = sourceRgn->attack_time;
+  newRgn->decay_time = sourceRgn->decay_time;
+  newRgn->sustain_level = sourceRgn->sustain_level;
+  newRgn->sustain_time = sourceRgn->sustain_time;
+  newRgn->release_time = sourceRgn->release_time;
+  return newRgn;
+}
+
+}  // namespace
 
 // ****************
 // NinSnesInstrSet
@@ -117,6 +144,49 @@ bool NinSnesInstrSet::parseInstrPointers() {
   return true;
 }
 
+void NinSnesInstrSet::useColl(const VGMColl* coll) {
+  if (coll == nullptr || coll->seq() == nullptr) {
+    return;
+  }
+
+  const auto* seq = dynamic_cast<const NinSnesSeq*>(coll->seq());
+  if (seq == nullptr || seq->rawFile() != rawFile() || seq->version != version) {
+    return;
+  }
+
+  const auto& percussionInstrNoteMap = seq->percussionInstrNoteMap();
+  if (percussionInstrNoteMap.empty()) {
+    return;
+  }
+
+  // Create the drumkit instrument for percussion note events
+  auto* drumKit = new VGMInstr(this, 0, 0, 127, 0, "Drum Kit");
+  for (const auto& [instrIndex, percussionDef] : percussionInstrNoteMap) {
+    // Get the referenced instrument by checking its instrument number
+    VGMInstr* sourceInstr = nullptr;
+    for (auto* instr : aInstrs) {
+      if (instr->instrNum == instrIndex) {
+        sourceInstr = instr;
+        break;
+      }
+    }
+    if (sourceInstr == nullptr) {
+      continue;
+    }
+
+    for (auto* sourceRgn : sourceInstr->regions()) {
+      drumKit->addRgn(cloneRgnForDrumKit(drumKit, sourceRgn, percussionDef.noteIndex, percussionDef.globalTranspose));
+    }
+  }
+
+  if (drumKit->regions().empty()) {
+    delete drumKit;
+    return;
+  }
+
+  addTempInstr(drumKit);
+}
+
 // *************
 // NinSnesInstr
 // *************
@@ -219,7 +289,7 @@ NinSnesRgn::NinSnesRgn(NinSnesInstr *instr,
   uint8_t adsr1 = readByte(offset + 1);
   uint8_t adsr2 = readByte(offset + 2);
   uint8_t gain = readByte(offset + 3);
-  int16_t pitch_scale;
+  uint16_t pitch_scale;
   if (version == NINSNES_EARLIER) {
     pitch_scale = (int8_t) readByte(offset + 4) * 256;
   }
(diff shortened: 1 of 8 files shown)


src/main/formats/NinSnes/NinSnesSeq.cpp
```diff
@@ -904,27 +951,43 @@ bool NinSnesTrack::readEvent(void) {
...
+
       uint8_t duration = (shared->spcNoteDuration * shared->spcNoteDurRate) >> 8;
       duration = std::min(std::max(duration, (uint8_t) 1), (uint8_t) (shared->spcNoteDuration - 2));
 
       // Note: Konami engine can have volume=0
+      switchToPercussionProgramIfNeeded();
       addPercNoteByDur(beginOffset,
                        curOffset - beginOffset,
-                       noteNumber,
+                       static_cast<int8_t>(noteIndex - cKeyCorrection - parentSeq->globalTranspose),
```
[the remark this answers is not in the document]
mikelow: This is incorrect.

src/main/formats/NinSnes/NinSnesSeq.h
```diff
@@ -160,6 +166,13 @@ class NinSnesSeq:
   // Falcom:
   uint16_t falcomBaseOffset;
 
+  void addPercussionInstrNoteMapping(uint8_t instrIndex, uint8_t noteIndex, int8_t globalTranspose) {
+    m_percussionInstrNoteMap[instrIndex] = {noteIndex, globalTranspose};
```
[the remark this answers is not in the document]
mikelow: I considered this but decided it's not worth the added complexity. We would need to create duplicated drum regions for each unique global transpose and it would get messy fast - either that or use pitch bend, which would be worse. In my testing, I didn't encounter a sequence with multiple global transpose events, so this uncommon if not rare. This is an edge case not worth the hassle.

src/main/formats/NinSnes/NinSnesInstr.cpp
```diff
@@ -117,6 +144,49 @@ bool NinSnesInstrSet::parseInstrPointers() {
   return true;
 }
 
+void NinSnesInstrSet::useColl(const VGMColl* coll) {
+  if (coll == nullptr || coll->seq() == nullptr) {
+    return;
+  }
```
sykhro: is it worth logging something at the debug level here?

src/main/formats/NinSnes/NinSnesSeq.cpp
```diff
@@ -719,7 +720,9 @@ NinSnesTrack::NinSnesTrack(NinSnesSection *parentSection, uint32_t offset, uint3
     : SeqTrack(parentSection->parentSeq, offset, length, theName),
       parentSection(parentSection),
       shared(NULL),
-      available(true) {
+      available(true),
+      m_lastNoteWasPercussion(false),
+      nonPercussionProgram(0) {
```
sykhro: I think these could be moved to in-class initializers

sykhro (commented):
couple nitpicks + changelog reminder!

src/main/formats/NinSnes/NinSnesInstr.cpp
mikelow: Looking at this closer, the `coll == nullptr` is redundant since we perform that check in VGMInstrSet immediately before calling this method. I'll remove it. Then this simplifies to just checking for the presence of a sequence in the collection. If there's no sequence, I wouldn't consider that a bug or debug-log worthy. If we want to maintain that a collection must always have a sequence (I'm not opposed to that), we should perform the check further up in VGMInstrSet before the call, but we haven't strictly defined that constraint.

**EDIT: I just added a debug log in VGMInstrSet if a sequence is missing from the collection**

mikelow:
Thanks for the feedback. I've updated the code:
- removed the redundant `coll == nullptr` check
- moved NinSnesSeq member initialization to be set in-class
- added this CHANGELOG entry under "Added": "Added Drum Kit support for the more standard versions of the N-SPC driver (Nintendo's SNES SDK driver)."

mikelow:
I also just added an L_DEBUG log during export in VGMInstrSet if the collection is missing a seq. Doesn't hurt, and it's probably indicative of a bug.

mikelow:
I've tweaked the CHANGELOG entries once again. My sense is that the CHANGELOG organization could be improved by separating format-related entries from everything else. I need to think on it more.

KungFuFurby:
> Another thing: for some reason, @loveemu reassigned what would normally be the EVENT_PERCUSSION_PATCH_BASE opcode to NOP for the HAL version of the driver. I tested many HAL games and couldn't determine why he did this (it breaks drum kit conversion) so I took it out.

The test case, Kirby Super Star/Kirby's Dream Land 3, dummies out the command and uses a hardcoded value instead. Same thing can be said for Kirby no Kirakira Kids. However, all older builds for the HAL variant do support this command.

mikelow:
> The test case, Kirby Super Star/Kirby's Dream Land 3, dummies out the command and uses a hardcoded value instead. Same thing can be said for Kirby no Kirakira Kids. However, all older builds for the HAL variant do support this command.

I see the opcode present in sequences of both titles (producing correct drums), so if they dummied it out and hard-coded the value I don't know why.

KungFuFurby:
Here's the disassembly of Kirby Super Star with a few notes:
- The dummied out VCMD FA call: https://github.com/loveemu/vgm-disasm/blob/49d763bb94b71fcbb60b65604e486e3f40144526/snes/NSPC/HAL%20Laboratory/Kirby%20Super%20Star.s#L530
- The hardcoded base percussion patch ID: https://github.com/loveemu/vgm-disasm/blob/49d763bb94b71fcbb60b65604e486e3f40144526/snes/NSPC/HAL%20Laboratory/Kirby%20Super%20Star.s#L120

You are also correct in that they still execute VCMD FA in the song data, even if it now does nothing, so you're not wrong for trying to use that off of the song data anyways.

mikelow:
> You are also correct in that they still execute VCMD FA in the song data, even if it now does nothing, so you're not wrong for trying to use that off of the song data anyways.

I see. That is oddly redundant. I may need to check that all sequences using drums also use VCMD FA to ensure it's safe to ignore the hard-code.

KungFuFurby:
I'll give you a list of other cases I've run into (note that I'm limiting this to those that are either not programmatically too different from standard Kankichi-kun or are known to have been explicitly covered by vgmtrans)...
- There are a few cases of the offset being ignored entirely, meaning these cases are zero offset cases. Assuming a standard variable ARAM map (of which both of these are), they can be detected via `D5 11 02 FD 10 03 80 A8 CA`. The ones I've noted are...
  - Las Vegas Dream/Vegas Stakes (yes, it's one of the HAL ones)
  - F-1 Grand Prix (this is Yoshikazu Yao's variant, predating a programmatic overhaul)
- We also have hardcoded offset cases to note, meaning they forcibly use this offset.
  - These cases are made in the instrument setup code (detection is via `D5 11 02 FD 10 06 80 A8 CA 60 88 xx` assuming a standard variable ARAM map, which the below case is confirmed)...
    - Ys III (this is Advance Communication's variant, but it predates any custom VCMDs not yet detected by vgmtrans) uses a constant of $19.
  - These adjustments are made outside the instrument setup code via using a subtraction operation (these can be detected via `68 CA 90 07 A8 xx 3F xx xx 8D A4`... coincidentally they're also from HAL)...
    - Kirby Super Star starts from $2C ($CA-$9E=$2C since we're using a SBC opcode)
    - Kirby's Dream Land 3 starts from $23 ($CA-$A7=$23 since we're using a SBC opcode)
    - Kirby no Kirakira Kids starts from $19 ($CA-$B1=$19 since we're using a SBC opcode)
  - These adjustments are made in the VCMD $FA code itself by overwriting the parameter with a constant (two kinds of detection assuming a standard variable ARAM map, which I confirmed at least for the target memory location for the four cases below: `E8 xx C4 5F 6F`, which is used by Gambler Jiko Chuushinha, and `8F xx 5F 6F`, which is used by the other three below)...
    - Gambler Jiko Chuushinha (this is a variant used by T's Music) uses a constant of $11.
    - Super Gachapon World - SD Gundam X (this is a variant used by T's Music) uses a constant of $1B.
    - Gegege no Kitarou - Fukkatsu! Tenma Daiou (this is a variant used by T's Music) uses a constant of $1A.
    - Super Bikkuriman (this is a variant used by T's Music) uses a constant of $07.

mikelow:
> I'll give you a list of other cases I've run into (note that I'm limiting this to those that are either not programmatically too different from standard Kankichi-kun or are known to have been explicitly covered by vgmtrans)...

Thank you for this in-depth breakdown. It's very helpful as I work on this.

PR #826, opened 2026-04-19
Collections search utility

sykhro:
## Description
- Added a search field in the **Collections** title bar.
- Implemented filtering in `VGMCollListView`
- Added reusable empty-state infrastructure and extracted CTA from MdiArea
  - `EmptyStateWidget` switches to painting instructions when "empty state" is tripped

## Motivation and Context
The Collections view needed in-place search to quickly find items.
Certain games can have thousands of collections. Filtering by `BGM_` for example makes for a much nicer experience in browsing certain Pokemon games!

This also centralizes empty-state rendering so other widgets can adopt consistent “no data / no results” states with less duplication.

I have another PR coming up for a "stitch mode" export and having the search bar is really useful for that.

## How Has This Been Tested?
Searching for things, playback

image
image
image

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/workarea/VGMCollListView.cpp
+++ b/src/ui/qt/workarea/VGMCollListView.cpp
@@ -7,16 +7,33 @@
 #include "VGMCollListView.h"
 
 #include <algorithm>
+#include <QStringView>
 #include <QLineEdit>
+#include <QResizeEvent>
 #include <VGMColl.h>
 #include <VGMExport.h>
 #include <VGMSeq.h>
 #include "SequencePlayer.h"
+#include "widgets/EmptyStateWidget.h"
 #include "widgets/ItemViewDensity.h"
 #include "workarea/MdiArea.h"
 #include "QtVGMRoot.h"
 #include "services/MenuManager.h"
 #include "services/NotificationCenter.h"
+#include "util/UIHelpers.h"
+
+static constexpr int kSearchEmptyCompactHeightThreshold = 170;
+
+static InstructionHint SearchEmptyStateHeadingHint() {
+  InstructionHint hint;
+  hint.iconPath = QStringLiteral(":/icons/magnify.svg");
+  hint.text = QStringLiteral("No matching collections");
+  hint.fontScale = 1.30;
+  hint.iconScale = 2.0;
+  hint.fontWeight = QFont::DemiBold;
+  hint.minPointSize = 12;
+  return hint;
+}
 
 static const QIcon &VGMCollIcon() {
   static QIcon icon(":/icons/collection.svg");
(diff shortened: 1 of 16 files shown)


mikelow:
Awesome! This is a good feature. I'll delve into the code in a moment. Some initial feedback:
- This has introduced a new minimum height for VGMCollListView - you can't lower the splitter as far as you once could. I'm guessing this is due to the new empty state widget.
- I was initially torn, but now I think the "No Collections Loaded" widget isn't a good addition. Here's why:
  * It distracts from the CTA, which communicates important instructions and creates slightly more cognitive load for the first-time user.  I'm talking about the fraction of a second when the user first sees the app layout. We want to minimize unpleasantness with unambiguous clarity.
  * It doesn't provide much clarity of its own. The ideas "This is where collections are listed" and "This is empty" are already pretty clear without the text.
  * It feels a little unbalanced having a CTA for one dock but not the others. But you can imagine the visual distraction of adding similar widgets for the other docks

Apologies if I sound like a stickler. I'm influenced by an excellent design book I recently read by Steve Krug titled "Don't Make Me Think". It's a great short read. The primary idea is to design UX that requires minimal cognitive work on the part of the user, especially on their first use when they don't know what is what and may have little patience. This is what motivated me to add the CTAs. Our app is complicated with all of its dock windows and is probably overwhelming for first time users. The CTA is intended to alleviate that with clear unmistakable instructions.

mikelow:
Just a quick idea: it may make sense to have this new empty state visual appear for when the user enters a search and no collections are found.

sykhro:
> * This has introduced a new minimum height for VGMCollListView - you can't lower the splitter as far as you once could. I'm guessing this is due to the new empty state widget.

I missed that, I'll see what I can do!

> * I was initially torn, but now I think the "No Collections Loaded" widget isn't a good addition.

I totally get that. I will just leave the widget for "no collections found" then!

> I'm influenced by an excellent design book I recently read by Steve Krug titled "Don't Make Me Think". It's a great short read

I'll go get that! I'm always on the look for resources on good UI design

sykhro:
I will say, I need `EmptyStateWidget` mostly for #827. I can also drop it from this one entirely

mikelow:
> I will say, I need EmptyStateWidget mostly for https://github.com/vgmtrans/vgmtrans/pull/827. I can also drop it from this one entirely

By all means, keep it.

mikelow:
@codex review

src/ui/qt/workarea/VGMCollListView.cpp
```diff
@@ -135,22 +144,232 @@ VGMCollListView::VGMCollListView(QWidget *parent) : QListView(parent) {
...
+
+void VGMCollListView::clearFilter() {
+  setFilterText(QString());
+}
+
+void VGMCollListView::applyFilter() {
+  const auto& allColls = qtVGMRoot.vgmColls();
+  int visibleCount = 0;
+  const bool hasFilter = !m_filterText.isEmpty();
+  if (!selectionModel()) {
```
mikelow: I'm a little confused by this block of code (`if (!selectionModel())`). When would selectionModel be null?

src/ui/qt/workarea/VGMCollListView.cpp
```diff
@@ -135,22 +144,232 @@ VGMCollListView::VGMCollListView(QWidget *parent) : QListView(parent) {
...
+      if (current == index) {
+        currentHidden = true;
+      }
+    }
+  }
+
+  if (currentHidden) {
+    if (firstVisibleIndex.isValid()) {
+      selectionModel()->setCurrentIndex(firstVisibleIndex, QItemSelectionModel::NoUpdate);
+    } else {
```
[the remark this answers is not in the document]
sykhro: We now grab `->selectedRows()` first

src/ui/qt/workarea/VGMCollListView.cpp
sykhro: Realistically never actually? I couldn't find API guarantees for that though

sykhro: the check is definitely redundant but I am wary of teardown/bringup. there are already unguarded usages in the same codepath so I'll remove it

mikelow:
More feedback:
- Instead of "No collections match query. Try a different search query", I think we could reduce this to: "No matching collections"

Nice to haves:
- Would be nice if we had a little search clear button like the one on the right:  image Just a nice to have, not needed for this PR.
- Not very important, but It would be nice if we retained selection in VGMCollListView such that if I select a set of collections, then type in a search that filters most of them out, then delete my search, the full selection is still there (and also VGMCollView wouldn't deselect on search input). I recognize this may be a major pain in the ass to implement, so I don't expect you to add this.

mikelow:
LGTM aside from that one point of feedback on copy. Hopefully getting rid of that third line of copy in the empty search state will fix this:
image

^ that's what the Collections dock looks like when fully shrunk height-wise.

mikelow (approved):

sykhro:
> Instead of "No collections match query. Try a different search query", I think we could reduce this to: "No matching collections"

Makes sense

> * Would be nice if we had a little search clear button like the one on the right:

I explicitly removed that because the button only appears when there's text and the field changing in length felt janky. I'll try to see if it's less janky with fixed width

> Not very important, but It would be nice if we retained selection in VGMCollListView such that if I select a set of collections, then type in a search that filters most of them out, then delete my search, the full selection is still there (and also VGMCollView wouldn't deselect on search input). I recognize this may be a major pain in the ass to implement, so I don't expect you to add this.

I'm not sure how to best handle the concept of filtering selection indexes like that, it's a fair thing to ask though! Might put in a second PR down the line for that

sykhro:
image
seems ok!

mikelow:
> I explicitly removed that because the button only appears when there's text and the field changing in length felt janky. I'll try to see if it's less janky with fixed width

> I'm not sure how to best handle the concept of filtering selection indexes like that, it's a fair thing to ask though! Might put in a second PR down the line for that

These are both just nice-to-haves. Not critical.

sykhro:
image
image

I am slightly bothered by the vertical spacing but for now this will do

mikelow:
> I am slightly bothered by the vertical spacing but for now this will do

Haha, I confess it bothers me too. I can try to fix it later. I'm impressed you went the extra mile and made it wrap the text.

sykhro:
> > I am slightly bothered by the vertical spacing but for now this will do
> 
> Haha, I confess it bothers me too. I can try to fix it later. I'm impressed you went the extra mile and made it wrap the text.

Too lazy to go the extra centimeter and make it resize the icon as well 🙈 I will try to see if there's a wireframe view of sorts to see the layout bounds, that would help debug the alignment. Doing this kinda thing in qt is always pain and leaving css-like hacks in c++ feels morally wrong........

PR #827, opened 2026-04-19
Export stitched collections

sykhro:
Merge after #826 

## Description
An utility to easily export songs split across multiple collections.
This is implemented as a popup of sorts where users drag the collections to merge.
It started out as a standalone dialog but I didn't like the code duplication and added complexity. I think this is also more fun!

- Added a merge utility to stitch together MIDI and create one big bank-remapped SF2
- Added `sequence stitch` shell command
- Added a baloon/popup in the Qt UI to stitch collections together
  - Users can drag (a selection of) collections into this baloon or on the merge icon
- Added stitch-position indicators in the collections UI to make queue order visible
- Added stitch-related resources/icons and updated user notifications
- Added changelog entry for stitch export

## Motivation and Context
This is required to support practical export of songs that are split into multiple sequence chunks, where users need one continuous result instead of many separate files.

A direct user request asked for MIDI merging support for chunked songs (example discussed: WarioWare DIY, Body Rock). The current batch export path emits separate files, which is inconvenient in DAWs that do not provide a smooth multi-file sequential import/stitch workflow. This feature solves that by building a single ordered export from multiple collections.

## How Has This Been Tested?
Exporting a bunch of stitched collections

image
image

## Types of changes

## Checklist:
- [x] I have read the CONTRIBUTING document.


mikelow:
This is super cool. I'm very impressed by the evolution of the UI into a drag and drop interface (yeah, I peaked at it before). Changing the icon of collections during stitch is very clever. Wish I'd thought to change the icon of the playing collection.

One early point of feedback: is it possible to change the VGMCollListView selection mode (or whatever it's called) to only do drag and drop when the stitcher dialog is visible? This is otherwise the only view that behaves this way, and there's currently no point in dragging collections without the stitcher view. Also, I like being able to click and drag to multi-select without holding shift.

Second thing: I can't seem to move the stitcher dialog. This is on macOS.

mikelow:
@codex review

sykhro:
> This is super cool. I'm very impressed by the evolution of the UI into a drag and drop interface (yeah, I peaked at it before). Changing the icon of collections during stitch is very clever. Wish I'd thought to change the icon of the playing collection.

Thanks! It started out as a clone of Finale's new document setup (the instrument seleciton bit). Then, as I reimplemented a lot of things from the main view into this auxiliary dialog (preview for auditing, search), it clicked that it would've been better to make it an interactive popup

> there's currently no point in dragging collections without the stitcher view.

Drag them to the stitch icon!...       ...yeah it's kinda useless. I will get rid of that

> Also, I like being able to click and drag to multi-select without holding shift.

valid, will go fix this!

> Second thing: I can't seem to move the stitcher dialog. This is on macOS.

This was intentional since it started out as a balloon. Can easily change it though

mikelow:
More non-code feedback:

- I would make the spacing between the +, stich, and search fields consistent. I think a spacing greater than the current stitch / search margin, but less than the + / stitch button margin would probably look good. image

-  The stitch dialog captures keyboard input (cmd-a works) but pressing delete/backspace causes a delete to happen on VGMCollListView. It would be good to have delete/backspace map to the "Remove" button, and also users might be unpleasantly surprised that pressing delete in the dialog is deleting their collections.

- Relatedly, if we delete collections from VGMCollListView, it would be good to delete them from the Stitch dialog. Fortunately, it ignores those deleted collections on export.

- Stitch export consistently fails when a NinSnes collection is included. Does the stitcher logic not handle VGMMultiSectionSeq correctly?

mikelow:
> Drag them to the stitch icon!... ...yeah it's kinda useless. I will get rid of that

Ha, I never would have discovered that.

> This was intentional since it started out as a balloon. Can easily change it though

Oh, I see. That could work with a pointer on the dialog chrome and a fixed position relative to the stitch button.
image
I think it also would work fine as a floating dialog.

~~Given the complexity of the dialog, I think I would opt for floating, but always on top, if that is possible.~~
I think it would work well either way. Just having it always on top is the key thing.

src/main/conversion/MidiMerge.h
```diff
@@ -0,0 +1,54 @@
...
+class MidiFile;
+
+namespace conversion {
+
+enum class MidiMergeMode : uint8_t {
+  Sequential
+};
+
+struct MidiMergeEntry {
+  VGMSeq* sequence = nullptr;
```
mikelow: Isn't including a sequence field redundant? A collection can only have one sequence. I suppose a collection can technically have no sequence, but then we should just ignore the entry.

Actually, I'm fine with assuming that a collection _must_ have a sequence or else it's an invalid collection. It makes sense conceptually, and right now, I don't think there's any way to make a collection without one.

src/main/conversion/MidiMerge.cpp
```diff
@@ -0,0 +1,471 @@
...
+#include "SynthFile.h"
+#include "VGMColl.h"
+#include "VGMInstrSet.h"
+#include "VGMSeq.h"
+
+namespace conversion {
+
+namespace {
+
+const VGMColl* resolveCollection(const MidiMergeEntry& entry) {
```
mikelow: This method is kind of weird. It implies that a MidiMergeEntry can have a VGMSeq _instead of_ a collection, and it will just use the sequence's collection in that case (if it has one).

This is inconsistent with the UI which only let's you add collections, not sequences (and I think that's all good).

If we remove the VGMSeq field from MidiMergeEntry, we could get rid of this method.

sykhro: you're right. to be honest, this did start as merging of VGMSeq: that's why the shell command is `sequence stitch [...]`. will update

sykhro:
> * The stitch dialog captures keyboard input (cmd-a works) but pressing delete/backspace causes a delete to happen on VGMCollListView. It would be good to have delete/backspace map to the "Remove" button, and also users might be unpleasantly surprised that pressing delete in the dialog is deleting their collections.

Backspace is currently mapped to "Remove" in VGMCollListView. So, Qt will complain that it doesn't know which one to trigger. imo there isn't much of a point in having Remove in collections, but it should be doable to make qt prefer the dialog's if it's open

sykhro:
> * 
> 
> * Stitch export consistently fails when a NinSnes collection is included. Does the stitcher logic not handle VGMMultiSectionSeq correctly?

I think it might be because NinSnes uses bank 127 and that messes with the reserving routine

mikelow:
> I think it might be because NinSnes uses bank 127 and that messes with the reserving routine

I'm going to look into this further, but I think the format conversion code should actually use bank 128, as this is SF2 convention. This always confused me - I assume it means major bank 1 minor bank 0 using whichever bank selection style allows 14-bit bank selection. I'll look into it.

Relatedly, I ran a Codex review of the branch locally and it brought up two issues. One is that `normalizeSf2Bank()` reduces the bank number down to 127 (>> 7 then & 0x7F), so bank 128 would become 0.

The second issue concerns the bank select style:

> P1 - Preserve MMA bank numbering when remapping stitched tracks
> When Bank Select Style is set to MMA, SeqTrack::addBankSelectNoItem() emits bank n as CC0 = n >> 7 and CC32 = n & 0x7f. This code adds bankOffset to the existing CC0 byte and then forces CC32 to 0, while saveMergedSoundfont() places the preset on normalizeSf2Bank(bank) + bankOffset. In a simple case like source bank 5 with chunk offset 10, the merged SF2 ends up on bank 15 but the merged MIDI is rewritten to bank 10:0, so stitched exports select the wrong preset bank whenever MMA mode is enabled.

mikelow:
Regarding the MMA bank issue, I've pushed a branch with a possible fix using codex. Feel free to use it or ignore it. It's `feature/stitch-export-mma-fix`.

src/main/conversion/MidiMerge.cpp
```diff
@@ -0,0 +1,480 @@
...
+    return tick;
+  }
+
+  const uint64_t scaled =
+      (static_cast<uint64_t>(tick) * static_cast<uint64_t>(dstPPQN) + (srcPPQN / 2u)) / srcPPQN;
+  return (scaled > std::numeric_limits<uint32_t>::max()) ? std::numeric_limits<uint32_t>::max()
+                                                          : static_cast<uint32_t>(scaled);
+}
+
+uint32_t shiftTick(uint32_t tick, uint32_t startTick) {
```
mikelow: So this is to ensure we don't overflow 32bits? But then if we hit that point, we'll always return std::numeric_limits<uint32_t>::max? I wonder if there's a better approach.

For example, would it be overly complex to calculate the converted tick length of each MIDI before the stitch to see if the sum exceeds the 32 bit bound (and fail the stitch if it does)? Then we could just use the + operator without worry, instead of calling this.

src/main/conversion/MidiMerge.cpp
```diff
@@ -0,0 +1,480 @@
...
+  if (!options.startTimes.empty()) {
+    startTicks = options.startTimes;
+  } else {
+    uint32_t cursor = 0;
+    for (size_t i = 0; i < parts.size(); ++i) {
+      startTicks[i] = cursor;
+      const uint32_t durationInTarget = rescaleTick(parts[i].durationTicks, parts[i].ppqn, targetPPQN);
+      cursor = shiftTick(cursor, durationInTarget);
+      cursor = shiftTick(cursor, options.sequentialGapTicks);
+    }
```
mikelow, line 392: As a follow up to my last comment, I'm wondering if it makes sense to make cursor uint64_t, then simply add `durationInTarget` and `sequentialGapTicks` here. Then when finished summing, we compare the tick against the max 32 bit value and fail if it exceeds it.

Sorry if I'm not making sense. I'm doing my best to understand.

sykhro: makes perfect sense actually. it's easy to end up in overengineering land :) will patch that

src/main/conversion/MidiMerge.cpp
```diff
@@ -0,0 +1,480 @@
...
+      return nullptr;
+    }
+
+    uint16_t ppqn = static_cast<uint16_t>(midi->ppqn());
+    if (ppqn == 0) {
+      ppqn = 48;
+      midi->setPPQN(ppqn);
+    }
+
+    targetPPQN = std::max(targetPPQN, ppqn);
```
mikelow: I wonder if we should use the least common multiple, unless it exceeds a threshold (in which case, fall back to max). We could be conservative about the threshold. Just a thought.

mikelow:
I notice that StitchCommands.cpp houses the definitions of all the UI classes. Do you not think we should put those in a separate file in within qt/widgets/ ?

src/ui/qt/services/commands/StitchCommands.cpp
```diff
@@ -0,0 +1,751 @@
...
+    setObjectName(QStringLiteral("stitchExportBalloon"));
+    setFrameShape(QFrame::NoFrame);
+    setFrameShadow(QFrame::Plain);
+    setMinimumSize(kBalloonMinWidth, kBalloonMinHeight);
+    setSizePolicy(QSizePolicy::Preferred, QSizePolicy::Preferred);
+    setAttribute(Qt::WA_DeleteOnClose, false);
+
+    auto* rootLayout = new QVBoxLayout(this);
+    rootLayout->setContentsMargins(10, 10, 10, 10);
+    rootLayout->setSpacing(8);
```
mikelow: The grab area is quite limited. We can fix it with a few simple changes. Here's the first:

```suggestion
    rootLayout->setContentsMargins(10, 0, 10, 10);
    rootLayout->setSpacing(0);
```

src/ui/qt/services/commands/StitchCommands.cpp
```diff
@@ -0,0 +1,751 @@
...
+
+    m_headingDragHandle = new StitchDragHandleWidget(this);
+
+    auto* headingRow = new QHBoxLayout(m_headingDragHandle);
+    headingRow->setContentsMargins(0, 0, 0, 0);
```
mikelow:
```suggestion
    headingRow->setContentsMargins(0, 10, 0, 8);
```

src/ui/qt/services/commands/StitchCommands.cpp
```diff
@@ -0,0 +1,751 @@
...
+    m_emptyState = new EmptyStateWidget(
+        defaultEmptyStateHeadingHint(QStringLiteral(":/icons/stitch.svg"),
+                                     QStringLiteral("Drop collections to stitch")),
+        m_queueList, this);
+    m_emptyState->setBodyText(
+        QStringLiteral("Drag from Collections, reorder, then export MIDI + SF2."));
+    rootLayout->addWidget(m_emptyState, 1);
+
+    auto* actionRow = new QHBoxLayout();
+    actionRow->setContentsMargins(0, 0, 0, 0);
```
mikelow: This is the last change.

```suggestion
    actionRow->setContentsMargins(0, 8, 0, 0);
```

src/ui/qt/workarea/VGMCollListView.cpp
```diff
@@ -40,6 +42,59 @@ static const QIcon &VGMCollIcon() {
   return icon;
 }
 
+static const QIcon &VGMPlayingCollIcon() {
+  static QIcon icon = stencilSvgIcon(QStringLiteral(":/icons/play.svg"), QColor(0x2f, 0xbf, 0x71));
```
mikelow: I know I haven't been doing this, but perhaps we should start defining these color values as constants in Colors.h, I suppose within the UIColors namespace.

sykhro: good idea, I keep chasing them

src/ui/qt/workarea/VGMCollListView.cpp
```diff
@@ -40,6 +42,59 @@ static const QIcon &VGMCollIcon() {
...
+  return icon;
+}
+
+QIcon StitchPositionIcon(int oneBasedPosition) {
+  constexpr int iconSide = 16;
+  const QPalette palette = qApp ? qApp->palette() : QPalette();
+  QColor bubbleColor = palette.color(QPalette::Highlight);
+  QColor textColor = palette.color(QPalette::HighlightedText);
+  if (!bubbleColor.isValid()) {
+    bubbleColor = QColor(70, 150, 255);
```
mikelow: Likewise, could go in Colors.h

sykhro:
> I notice that StitchCommands.cpp houses the definitions of all the UI classes. Do you not think we should put those in a separate file in within qt/widgets/ ?

right, it was a small inline dialog and it became a 800LoC monster ^^ I will explode it in a couple files

src/ui/qt/services/commands/StitchCommands.cpp
```diff
@@ -0,0 +1,751 @@
...
+
+    adjustSize();
+    if (width() < kBalloonMinWidth) {
+      resize(kBalloonMinWidth, height());
+    }
+    if (height() < kBalloonMinHeight) {
+      resize(width(), kBalloonMinHeight);
+    }
+
+    positionRelativeTo(anchor);
```
mikelow: I propose removing positionRelativeTo() and having the Stitch dialog appear in the lower left corner of MdiArea:


```suggestion
    const QRect mdiRect(MdiArea::the()->mapToGlobal(QPoint(0, 0)), MdiArea::the()->size());
    move(mdiRect.left(), mdiRect.bottom() - height() + 1);
    move(mdiRect.left() + 20, mdiRect.bottom() - height() - 20);
```

Down the road, I think it would be good to allow dropping sequences into the dialog from  "VGMFileListView" . We would check if they have an associated collection on drop and show an error toast if they don't have one.

sykhro: sounds like a good plan

src/main/conversion/MidiMerge.cpp
sykhro: I've added it (I hope correctly), current PPQN is not very conservative atm (considering almost no daw goes beyond 960). any suggestions?

src/ui/qt/services/commands/StitchCommands.cpp
```diff
@@ -0,0 +1,753 @@
...
+#include "util/UIHelpers.h"
+#include "widgets/EmptyStateWidget.h"
+#include "workarea/VGMCollListView.h"
+
+namespace {
+
+constexpr int RoleCollectionPointer = Qt::UserRole;
+constexpr QSize kQueueIconSize(16, 16);
+constexpr QSize kActionIconSize(16, 16);
+constexpr int kBalloonMinWidth = 420;
```
mikelow: I think we can improve the dialog drag reordering experience by expanding the height of each row.

We just need to call setSizeHint() with this constant within makeCollectionItem(). I'll add that below.

```suggestion
constexpr int kQueueRowHeight = 28;
constexpr int kBalloonMinWidth = 420;
```

src/ui/qt/services/commands/StitchCommands.cpp
```diff
@@ -0,0 +1,753 @@
...
+  if (coll && coll->seq()) {
+    label += QStringLiteral(" - ");
+    label += QString::fromStdString(coll->seq()->name());
+  }
+  return label;
+}
+
+QListWidgetItem *makeCollectionItem(VGMColl *coll) {
+  auto *item = new QListWidgetItem(collectionLabel(coll));
+  item->setData(RoleCollectionPointer,
```
mikelow:
```suggestion
  item->setSizeHint(QSize(0, kQueueRowHeight));
  item->setData(RoleCollectionPointer,
```

src/ui/qt/workarea/VGMCollListView.cpp
```diff
@@ -136,11 +242,17 @@ VGMCollListView::VGMCollListView(QWidget *parent) : QListView(parent) {
 
   setContextMenuPolicy(Qt::CustomContextMenu);
   setSelectionMode(QAbstractItemView::ExtendedSelection);
+  setSelectionRectVisible(true);
```
mikelow: I prefer the old style of multi-selection. Do you have any objection to removing this line?

sykhro: nop! can remove, I don't particularly care since I always shift+click for selection

src/ui/qt/services/commands/StitchCommands.cpp
sykhro: done, it's better. I'm thinking of ways of making the draggable nature of these more obvious

mikelow:
> Backspace is currently mapped to "Remove" in VGMCollListView. So, Qt will complain that it doesn't know which one to trigger. imo there isn't much of a point in having Remove in collections, but it should be doable to make qt prefer the dialog's if it's open

We could remove collection deletion. I added it because it seemed incongruous to allow users to add collections but not delete them (short of removing the files).

sykhro:
> > Backspace is currently mapped to "Remove" in VGMCollListView. So, Qt will complain that it doesn't know which one to trigger. imo there isn't much of a point in having Remove in collections, but it should be doable to make qt prefer the dialog's if it's open
> 
> We could remove collection deletion. I added it because it seemed incongruous to allow users to add collections but not delete them (short of removing the files).

my case for removing the option of deleting collections is that there is no easy way to add them back (manual collection is useful but you have to know what to piece together)

mikelow:
> my case for removing the option of deleting collections is that there is no easy way to add them back (manual collection is useful but you have to know what to piece together)

Yeah, I think that's a good point. And the Add Manual Collection feature is probably so seldomly used, I expect collection deletion is being invoked by mistake the vast majority of the time. I'll remove it in a separate PR.

mikelow:
Ok, this LGTM. I may propose minor tweaks in a future PR. Debating if we should rename the "Stitch Queue" to something like "Stitch Collections" or "Stitch Collection Queue" or similar. I also might prefix the rows with the order numbers.

Also, NinSnes sequences still don't work, but 🤷‍♂️.

Nice job! This was unexpected.

sykhro:
> Ok, this LGTM. I may propose minor tweaks in a future PR. Debating if we should rename the "Stitch Queue" to something like "Stitch Collections" or "Stitch Collection Queue" or similar. I also might prefix the rows with the order numbers.

How about "Stitch Order"? On the topic of order numbers, I noticed there's a bit of a performance hit when adding many tracks. It's most likely the pixmap generation for the numbers in the icon panel. Haven't profiled yet, but I think it can be addressed down the line.

> Also, NinSnes sequences still don't work, but 🤷‍♂️.

No clue what's going on there :( will try to debug that eventually

> Nice job! This was unexpected.

Thanks! It was a feature I half-promised someone a couple weeks ago. Was making it into a standalone tool but it ended up being way easier to do it directly here :D

mikelow:
Also, I should note, beyond observing that the MIDI was appended and timing was correct, I haven't tested this in a DAW to ensure that SF2 is stitched and consistent with the MIDI events.

sykhro:
> Also, I should note, beyond observing that the MIDI was appended and timing was correct, I haven't tested this in a DAW to ensure that SF2 is stitched and consistent with the MIDI events.

I've listened to a bunch of things in fluidsynth, all okay so far.. I bet there's problem cases, hoping in bug reports!

mikelow:
> How about "Stitch Order"? 

Well, I think we should give it a title indicative of the action it performs. I'm leaning toward. "Collection Stitcher". I think "Collection" belongs in the title, because that's kind of central to understanding it. And if the user opens the dialog with an item selected, they won't see the explanatory empty state copy.

I also forgot that empty state copy. I may propose a revision to it. But it's fine to merge.

> On the topic of order numbers, I noticed there's a bit of a performance hit when adding many tracks. It's most likely the pixmap generation for the numbers in the icon panel. Haven't profiled yet, but I think it can be addressed down the line.

How many tracks are you talking about?

sykhro:
> How many tracks are you talking about?

~30, there is a bit of a delay after dropping. I guess it would be cool to start computing those pixmaps async soon as drag is initiated so that by the time they're dropped they've already been generated (easier said than done)

mikelow (approved):

PR #830, opened 2026-04-20
Remove CloseVGMCollCommand

mikelow:
This removes the "Remove" action for collections, as this is likely causing more accidental grief than utility.


sykhro (approved):
can now simplify the shortcut code in the stitcher dialog

PR #832, opened 2026-04-20
our toast got roast; fix the host!

mikelow:
This PR fixes toast notification behavior when RHI-backed views are present, while keeping the original simpler logic for Linux where the workaround is not needed.

### The Problem

Toasts are drawn behind Rhi windows. This affects macOS and Windows, but not Linux where we use the `QRhiWidget` class.

Also, toasts were being drawn on top of WindowBar. We want them anchored to the the MdiArea.

### The Fix

We create two modes for ToastHost: ChildWidget and ToolWindow. We use ToolWindow mode on Mac and Windows os that  toast notifications are made Qt::Tool Windows instead of child widgets. On Mac, we go a step further and make them native NSWindows child to the main app window, which fixes a visual latency issue. Linux uses ChildWidget mode, and retains the original logic.

Also, we assign MdiArea as the "anchor" widget. Both modes position toasts to this anchor.

The only apparent downside to the ToolWindow workaround is that the toast stack is no longer bounded by the app window.

### Changes

- Anchor toast placement to `MdiArea` instead of the main window so notifications no longer overlap the `WindowBar`.
- Add explicit toast host modes:
  - `ChildWidget` for the original embedded-widget behavior
  - `ToolWindow` for platforms where toasts must render above native/RHI-backed content
- Make `ToastHost::defaultMode()` choose the platform default:
  - Linux: `ChildWidget`
  - other platforms: `ToolWindow`
- Refactor toast positioning so the host owns anchoring/reflow and `Toast` only handles mode-specific placement.
- In `ToolWindow` mode on macOS, attach toast windows as native child `NSWindow`s of the main app window so they move with the app during live drags. Without this fix, dragging the app window results in visible delay while the toast windows catch up.

## How Has This Been Tested?
Tested on all three platforms.

## Types of changes

## Checklist:

### DIFF
--- a/src/ui/qt/MainWindow.cpp
+++ b/src/ui/qt/MainWindow.cpp
@@ -367,7 +367,7 @@ void MainWindow::createElements() {
       {m_logger->toggleViewAction(), QStringLiteral(":/icons/book-open-variant-outline.svg")},
   });
   createStatusBar();
-  m_toastHost = new ToastHost(this, MdiArea::the());
+  m_toastHost = new ToastHost(this, MdiArea::the(), ToastHost::Mode::ToolWindow);
 }
 
 void MainWindow::configureWindowAgent() {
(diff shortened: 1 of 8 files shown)


sykhro (approved):
LGTM!

PR #840, opened 2026-04-25
HexView Refactor 8 - colored playback range plumbing

mikelow:
This PR lays the data and rendering-plumbing groundwork for colored playback ranges in `HexView`. Playback ranges are now represented by a dedicated `PlaybackSelection` type that carries `offset`, `length`, and an optional per-range `glowColor`, and `HexView::setPlaybackSelectionsForItems(...)` now accepts a parallel list of colors while falling back to the existing default playback glow when none is supplied. Fade-out playback state is updated to preserve that richer playback selection metadata as ranges transition out of the active set.

It also splits playback updates onto their own invalidation path through `HexView`, `HexViewRhiHost`, and `HexViewRhiRenderer`, so playback-only changes can rebuild the overlay data without unnecessarily treating them as general selection or base-content changes. The frame snapshot and renderer are updated to consume the new playback selection structures while keeping the current active and fading playback highlight behavior intact.

I should note that this does not yet actually update the glow colors yet, just the plumbing between HexView and HexViewRhiRenderer. The actual rendering logic will be updated for this in a future PR.

## Types of changes

## Checklist:


src/ui/qt/workarea/hexview/HexViewRhiRenderer.cpp
```diff
@@ -1608,10 +1617,15 @@ void HexViewRhiRenderer::buildSelectionInstances(int startLine, int endLine,
   }
 
   if (!frame.playbackSelections.empty()) {
+    std::vector<HexViewFrame::SelectionRange> playbackRanges;
+    playbackRanges.reserve(frame.playbackSelections.size());
+    for (const auto& selection : frame.playbackSelections) {
+      playbackRanges.push_back({selection.offset, selection.length});
+    }
```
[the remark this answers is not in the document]
mikelow: This is understood. This is just plumbing. Actual color glow will come later

src/ui/qt/workarea/hexview/HexView.cpp
```diff
@@ -61,6 +61,7 @@ constexpr float PLAYBACK_GLOW_RADIUS = 1.8f;
 constexpr float PLAYBACK_GLOW_EDGE_CURVE = 0.85f;
 const QColor PLAYBACK_GLOW_LOW(40, 40, 40);
 const QColor PLAYBACK_GLOW_HIGH(230, 230, 230);
+const QColor PLAYBACK_GLOW_FALLBACK(230, 230, 230);
```
sykhro, line 64: reminder for a later pr: move to uicolors

src/ui/qt/workarea/hexview/HexView.cpp
```diff
@@ -738,15 +751,19 @@ void HexView::setSelectedItems(const std::vector<const VGMItem*>& items,
...
+  for (size_t i = 0; i < items.size(); ++i) {
+    const auto* item = items[i];
     if (!item) {
       continue;
     }
     const uint32_t length = item->length() > 0 ? item->length() : 1u;
-    next.push_back({item->offset(), length});
+    const QColor glowColor =
+        (i < glowColors.size() && glowColors[i].isValid()) ? glowColors[i] : PLAYBACK_GLOW_FALLBACK;
+    next.push_back({item->offset(), length, glowColor});
```
sykhro:
```suggestion
    next.emplace_back(item->offset(), length, glowColor);
```

sykhro (approved):
not a whole lot to note here, left 2 tiny comments

src/ui/qt/workarea/hexview/HexView.cpp
mikelow: Actually, I find that using emplace_back here results in a compiler error on my machine, with the error going deep into internal SDK:

```
.../MacOSX14.5.sdk/usr/include/c++/v1/__memory/allocator_traits.h:304:9: error: no matching function for call to 'construct_at'
```

I'm confused that it doesn't look like there was a problem on the CI cloud servers.

Sans a fix, I'm going to revert this back to push_back in the next PR.

mikelow: I've confirmed the issue is that I'm using AppleClang 15. Given that the performance difference is microscopic, I'll revert to the push_back line for the sake of compatibility.

PR #843, opened 2026-04-27
KonamiSnes improvements

mikelow:
This PR expands support for the KonamiSnes driver:
- adds pitch slide, volume fade, pan fade, and tempo fade, including support for both version variants of these events
- fixes note velocity calculation for V1
- fixes percussion instrument potentially not including all regions
- fixes max duration percussion notes treated as tied

Notable features missing:
- accurate V2-V6 volume handling (final volume uses a table that is indexed after applying channel volume. Very anti-MIDI).
- proper handling of instrument volume and pan settings
- ADSR/GAIN events
- pitch envelope events
- V1's weird software release envelope
- portamento
- vibrato

This work relied heavily on @loveemu's disassembly and annotation of the SPC700 code. Many thanks to him!

## Audio examples (unmute)

https://github.com/user-attachments/assets/b9a6d69a-2083-490f-aed5-463c6c472f79

https://github.com/user-attachments/assets/035d1fbe-44f0-49d1-be8f-45769553f2b3

## How Has This Been Tested?
Tested with many Konami titles

## Types of changes

## Checklist:


src/main/formats/KonamiSnes/KonamiSnesSeq.cpp
```diff
@@ -420,36 +947,40 @@ bool KonamiSnesTrack::readEvent(void) {
...
-      uint8_t dur = len;
-      if (noteDurationRate != parentSeq->NOTE_DUR_RATE_MAX) {
-        if (parentSeq->version == KONAMISNES_V1) {
-          dur = (len * noteDurationRate) / 100;
-        }
-        else {
-          dur = (len * (noteDurationRate << 1)) >> 8;
-        }
+      vel = static_cast<uint8_t>(std::clamp<int>(vel + getLoopVolumeDelta(), 1, 127));
+      vel = convertPercentAmpToStdMidiVal(vel / 127.0);
```
sykhro, line 970: the PR says the velocity is fixed only for V1, so, is this supposed to be running for V2+?

src/main/formats/KonamiSnes/KonamiSnesSeq.cpp
```diff
@@ -319,8 +370,516 @@ uint8_t KonamiSnesTrack::convertGAINAmountToGAIN(uint8_t gainAmount) {
...
+  auto nextEvent = parentSeq.EventMap.find(statusByte);
+  if (nextEvent == parentSeq.EventMap.end()) {
+    return std::nullopt;
+  }
+
+  switch (nextEvent->second) {
+    case EVENT_PITCH_SLIDE_V1:
+    case EVENT_PITCH_SLIDE_V2:
+    case EVENT_PITCH_SLIDE_V3:
+      return readPitchSlide(nextEvent->second, curOffset++);
```
sykhro, line 459: should we be doing isValidOffset per each `EVENT_PITCH_SLIDE_V`? the initial call only guards the statusByte here

sykhro (commented):
great stuff! the next release will be huge

src/main/formats/KonamiSnes/KonamiSnesSeq.cpp
mikelow: I figured this was less wrong than the prior implementation, but I'm not even sure of that.

I'll just add the V2-V6 volume table and index `vel` into that. That's closer to the actual behavior.

The V2-V6 driver logic is essentially:
```
finalVolume = volumeTable[ (velocity + loop vol deltas + instrument vol) * volume ]
```
Where `(finalVolume / 255)` is the percent of max amplitude. I'm ignoring pan for simplicity.

Once I add the table our logic is essentially:
```
finalVolume = volumeTable[ (velocity + loop vol deltas) ] * volume * instrumentVol
```

Accounting for instrumentVol is relatively easy with useColl() - we can remove volume from converted instruments and subtract their volume val from velocity instead (they are attenuation only, I believe). But accounting for the channel volume is tricker. Unlike velocity, volume applies to actively playing notes. So we either drop velocity altogether (always set velocity to max and use volume as the single source of volume control), or we come up with a system where velocity and volume can both be set but must account for the final volume each time they change.

src/main/formats/KonamiSnes/KonamiSnesSeq.cpp
mikelow: I assume you mean the `isValidOffset(...)` above this highlighted code at the start of `consumePitchSlide(...)`?

It's kind of arbitrary, and yes it's not checking if the bytes that may be read beyond the statusByte are valid.

But we have a lot of places in the code where we're reading bytes without checking if they are within bounds: `readVolumeFade()`, `readPanFade()`, `readTempoFade()`, `EVENT_TIE`, `EVENT_VIBRATO`, `EVENT_ADSR_GAIN`.

Adding a check on each seems like bad design. We could add a magic number onto the guard at the top of readEvent() (`if (curOffset >= 0x10000)`) though that doesn't seem clean either.

src/main/formats/KonamiSnes/KonamiSnesInstr.cpp
```diff
@@ -52,9 +62,16 @@ std::vector<PercussionHeader> collectPercussionHeaders(RawFile *file,
   for (uint8_t percussionNote = 0; percussionNote < kKonamiSnesPercussionNoteCount; percussionNote++) {
     const uint32_t addrInstrHeader = tableOffset + (instrItemSize * percussionNote);
     if (!isValidPercussionHeader(file, version, addrInstrHeader, spcDirAddr)) {
-      // Some games leave unused slots at the front, but once real drum entries start,
-      // the first invalid header means we have walked past the table.
-      if (!headers.empty()) {
+      const bool legacyLayout = usesLegacyInstrumentLayout(version);
+      const uint8_t pan = file->readByte(addrInstrHeader + (legacyLayout ? 6 : 5));
```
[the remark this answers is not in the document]
mikelow: I've added a guard here.

src/main/formats/KonamiSnes/KonamiSnesSeq.cpp
sykhro: oh I see, that sounds tricky. tbh i'm never a fan of losing velocity info because it makes for more work when editing the midi

src/main/formats/KonamiSnes/KonamiSnesSeq.cpp
sykhro: that's fair, tbh with our kinda data files are either fine or completely broken. it's probably better to have guards in place at the lowest file-reading levels just to avoid crashes and trust everything else is fine

mikelow:
@loveemu would you be comfortable if I credited you using "Co-authored-by:" on the squashed commit using your gmail address? Totally fine if you’d rather I just credit you in the PR/commit body instead. Also curious what your preference would for future PRs that rely on your reverse-engineering work.

src/main/formats/KonamiSnes/KonamiSnesSeq.cpp
mikelow: > that's fair, tbh with our kinda data files are either fine or completely broken. it's probably better to have guards in place at the lowest file-reading levels just to avoid crashes and trust everything else is fine

I agree. I just want to avoid crashes without inserting a ton of guards, if possible. 

There's a lot of built-in protection against scanner false positives with some of these formats - the SNES formats especially. We're strictly only scanning on .spc files, we're verifying file length is exactly 64kb, and the BytePattern heuristics are highly-specific. Still, parsing bugs could lead to bad address reads.

loveemu:
I don’t have any strong preferences regarding the format of the credit.
I’m happy for you to include my Gmail address using the "Co-authored-by:" format.
Feel free to include links to the disassembly files that served as references.

Thank you for creating such a wonderful PR. I’m really glad to hear that my work has been helpful.

PR #851, opened 2026-05-04
Add global modulator and generator support for instruments

mikelow:
This adds a light abstraction for adding modulators and generators to an instrument at a global level (in DLS parlance they're called "connection blocks"). I know you're already familiar with this, but as a reminder: 
- **_Generators_** allow us to assign fixed values for instrument parameters, like fine tune, vibrato settings, pan, etc. Our code already adds generators at the region level to define all of a region's properties. 
- **_Modulators_** allow us to assign a source input to modify a parameter - for example, assigning a specific controller to change the vibrato rate up to a specified amount. They support other forms of input too, like channel pressure events.

Incidentally, I've also added support for ChannelPressure events.

As we've discussed on the ADSR PR, the DLS2 spec for connection blocks is more constraining than SF2 modulators. It can only utilize a small subset of controllers as sources, and we can't assign a source to control certain properties like vibrato delay. We will have to choose how we handle these limitations going forward - I'm inclined toward a system that allows users to map modulators as they like, even if that means breaking DLS support.

For the time being, I've set up the system to conform to the constraints of DLS. This will change quickly when we need vibrato delay modulator support.

Hopefully this doesn't step too much on the toes of your relevant draft PRs.

## Changes

- Introduces shared instrument modulation types (`ModSource`, `ModDest`, `InstrumentModulator`, `InstrumentGenerator`, `ParamAmount`) so format code can describe modulation in musical units instead of raw SF2/DLS scalars.
- Extends `VGMInstr` and `SynthInstr` to store instrument-global generators and modulators, with helper APIs for pitch, frequency, delay, attenuation, plus standardized vibrato/tremolo helper methods.
- Updates SF2 export to create true instrument global zones and write instrument-level `imod` and `igen` entries, including the required bag/index bookkeeping when a global zone is present.
- Updates DLS export to emit instrument-level articulation connection blocks from the same data, adds `InitialAttenuation` support, and switches vibrato-related routing to the DLS2-specific vibrato source/destination constants.
- Also updates vibrato handling to use the Vibrato LFO instead of the Modulation LFO. This is semantically more correct and will allow independent modulation LFO changes, but breaks support for DLS1.
- Expands supported modulation sources beyond Mod Wheel to include Channel Pressure, Poly Pressure, Pitch Wheel, Volume, Pan, Expression, Reverb Send, and Chorus Send.
- Adds Channel Pressure as a first-class sequence/MIDI event from `SeqEvent`/`SeqTrack` through `MidiTrack`, including absolute-time insertion and MIDI serialization as channel pressure (`0xD0`) events.
- Adds the corresponding event/item enum plumbing so Channel Pressure can be represented in the event model.

One more thing: The `VGMInstr::addStandardVibratoHandling(...)` and `VGMInstr::addStandardTremoloHandling(...)` helper methods are intended to provide easy setup of what I anticipate to be a common modulator / generator pattern for vibrato and tremolo support. However, I've only added support for one format (CapcomSnes - PR forthcoming). As support for vibrato is added to more formats, I expect these methods may change or others may be added. The objective, as always, is to make implementing formats as simple and intuitive as possible.

## Motivation and Context
The goal is create a system for adding vibrato and tremolo support using the intended MIDI/SF2 abstractions rather than many pitch bend and volume/expression controller events. We may want to add an option down the road to alternatively use pitch bend and volume/expression for maximum compatibility (the CPS driver already does this), but I think this is the better default approach, and it's less complicated.

## How Has This Been Tested?
The only regression I could anticipate would be the DLS change to use Vibrato LFO. I don't have a good way to test that, and I'm fairly confident the change is safe.

As for testing the generator and modulator functionality: I've been doing this while implementing vibrato and tremolo support in the CapcomSnes driver.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/components/instr/InstrumentModulation.cpp
+++ b/src/main/components/instr/InstrumentModulation.cpp
@@ -1,0 +1,75 @@
+/*
+ * VGMTrans (c) 2002-2024
+ * Licensed under the zlib license,
+ * refer to the included LICENSE.txt file
+ */
+
+#include "InstrumentModulation.h"
+#include <cmath>
+#include "ScaleConversion.h"
+
+namespace {
+
+constexpr double kSf2LfoReferenceHz = 8.176;
+
+bool isValidFiniteNumber(double value) {
+  return std::isfinite(value);
+}
+
+int32_t hertzToInstrumentCents(double hertz) {
+  return static_cast<int32_t>(std::lround(1200.0 * std::log2(hertz / kSf2LfoReferenceHz)));
+}
+
+}  // namespace
+
+ParamAmount ParamAmount::raw(int32_t amount) {
+  return ParamAmount(amount, true);
+}
+
+ParamAmount ParamAmount::cents(double cents) {
+  if (!isValidFiniteNumber(cents)) {
+    return ParamAmount(0, false);
+  }
+
+  return ParamAmount(static_cast<int32_t>(std::lround(cents)), true);
+}
+
+ParamAmount ParamAmount::hertz(double hertz) {
+  if (hertz <= 0.0 || !isValidFiniteNumber(hertz)) {
+    return ParamAmount(0, false);
+  }
+
+  return ParamAmount(hertzToInstrumentCents(hertz), true);
+}
+
+ParamAmount ParamAmount::hertzRange(double minHertz, double maxHertz) {
+  if (minHertz <= 0.0 || maxHertz <= 0.0 ||
+      !isValidFiniteNumber(minHertz) || !isValidFiniteNumber(maxHertz)) {
+    return ParamAmount(0, false);
+  }
+
+  const double minCents = static_cast<double>(hertzToInstrumentCents(minHertz));
+  const double maxCents = static_cast<double>(hertzToInstrumentCents(maxHertz));
+  const double fullScaleRange = (maxCents - minCents) * 128.0 / 127.0;
+  return ParamAmount(static_cast<int32_t>(std::lround(fullScaleRange)), true);
+}
+
+ParamAmount ParamAmount::seconds(double seconds) {
+  return ParamAmount(secondsToSf2Timecents(seconds), true);
+}
+
+ParamAmount ParamAmount::centibels(double centibels) {
+  if (!isValidFiniteNumber(centibels)) {
+    return ParamAmount(0, false);
+  }
+
+  return ParamAmount(static_cast<int32_t>(std::lround(centibels)), true);
+}
+
+ParamAmount ParamAmount::attenuationDb(double decibels) {
+  if (!isValidFiniteNumber(decibels)) {
+    return ParamAmount(0, false);
+  }
+
+  return ParamAmount(static_cast<int32_t>(std::lround(decibels * 10.0)), true);
+}
--- a/src/main/components/instr/InstrumentModulation.cpp
+++ b/src/main/components/instr/InstrumentModulation.cpp
@@ -1,5 +1,5 @@
 /*
- * VGMTrans (c) 2002-2024
+ * VGMTrans (c) 2002-2026
  * Licensed under the zlib license,
  * refer to the included LICENSE.txt file
  */
@@ -12,10 +12,6 @@ namespace {
 
 constexpr double kSf2LfoReferenceHz = 8.176;
 
-bool isValidFiniteNumber(double value) {
-  return std::isfinite(value);
-}
-
 int32_t hertzToInstrumentCents(double hertz) {
   return static_cast<int32_t>(std::lround(1200.0 * std::log2(hertz / kSf2LfoReferenceHz)));
 }
@@ -27,7 +23,7 @@ ParamAmount ParamAmount::raw(int32_t amount) {
 }
 
 ParamAmount ParamAmount::cents(double cents) {
-  if (!isValidFiniteNumber(cents)) {
+  if (!std::isfinite(cents)) {
     return ParamAmount(0, false);
   }
 
@@ -35,7 +31,7 @@ ParamAmount ParamAmount::cents(double cents) {
 }
 
 ParamAmount ParamAmount::hertz(double hertz) {
-  if (hertz <= 0.0 || !isValidFiniteNumber(hertz)) {
+  if (hertz <= 0.0 || !std::isfinite(hertz)) {
     return ParamAmount(0, false);
   }
 
@@ -44,7 +40,7 @@ ParamAmount ParamAmount::hertz(double hertz) {
 
 ParamAmount ParamAmount::hertzRange(double minHertz, double maxHertz) {
   if (minHertz <= 0.0 || maxHertz <= 0.0 ||
-      !isValidFiniteNumber(minHertz) || !isValidFiniteNumber(maxHertz)) {
+      !std::isfinite(minHertz) || !std::isfinite(maxHertz)) {
     return ParamAmount(0, false);
   }
 
@@ -59,7 +55,7 @@ ParamAmount ParamAmount::seconds(double seconds) {
 }
 
 ParamAmount ParamAmount::centibels(double centibels) {
-  if (!isValidFiniteNumber(centibels)) {
+  if (!std::isfinite(centibels)) {
     return ParamAmount(0, false);
   }
 
@@ -67,7 +63,7 @@ ParamAmount ParamAmount::centibels(double centibels) {
 }
 
 ParamAmount ParamAmount::attenuationDb(double decibels) {
-  if (!isValidFiniteNumber(decibels)) {
+  if (!std::isfinite(decibels)) {
     return ParamAmount(0, false);
   }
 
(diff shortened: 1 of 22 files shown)


sykhro:
> Hopefully this doesn't step too much on the toes of your relevant draft PRs.

I was eager to redo those with a more generalised approach so thanks! I will migrate the NDS one to this

src/main/components/instr/Modulation.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
```
sykhro, line 7: license header is missing

src/main/components/instr/InstrumentModulation.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include "InstrumentModulation.h"
+#include <cmath>
+#include "ScaleConversion.h"
+
+namespace {
+
+constexpr double kSf2LfoReferenceHz = 8.176;
+
+bool isValidFiniteNumber(double value) {
```
sykhro: why not call `std::isfinite()` directly?

src/main/components/instr/VGMInstrSet.cpp
```diff
@@ -168,3 +168,93 @@ VGMRgn *VGMInstr::addRgn(uint32_t offset, uint32_t length, int sampNum, uint8_t
...
+  // nullify default channel pressure to vib lfo pitch modulator
+  addPitchModulator(ModSource::ChannelPressure, ModDest::VibLfoToPitch, 0);
+  addFrequencyGenerator(ModDest::VibLfoFreq, minHertz);
+  addFrequencyRangeModulator(ModSource::ChannelPressure, ModDest::VibLfoFreq, minHertz, maxHertz);
+}
+
+void VGMInstr::addStandardTremoloHandling(double maxDepthDb,
+                                         double minHertz,
+                                         double maxHertz,
+                                         bool attenOnly) {
```
sykhro: I wonder if using an enum like `TremoloMode::LfoOnly / AttenuationOnly / Both` would be easier to read than `attenOnly`

src/main/components/instr/VGMInstrSet.cpp
```diff
@@ -168,3 +168,93 @@ VGMRgn *VGMInstr::addRgn(uint32_t offset, uint32_t length, int sampNum, uint8_t
...
+  addFrequencyGenerator(ModDest::ModLfoFreq, minHertz);
+  addFrequencyRangeModulator(ModSource::ChannelPressure, ModDest::ModLfoFreq, minHertz, maxHertz);
+  addAttenuationModulator(ModSource::ChorusSend, ModDest::ModLfoToVol, maxDepthDb);
+  if (attenOnly) {
```
sykhro: wrt to the above, i'm finding this a bit confusing

sykhro (changes requested):

sykhro:
at any rate, good stuff. we need more of this kinda thing

src/main/components/instr/InstrumentModulation.cpp
mikelow: This is hilarious and embarrassing.

src/main/components/instr/VGMInstrSet.cpp
mikelow: I agree. And if more options are needed, this could become a bitfield.

I'll have to think about the names of the two options - "LfoOnly" doesn't seem right. In case it's unclear, the problem I'm solving here is that the CapcomSnes tremolo only ever attenuates the volume - it does not boost it. SF2 and DLS modToVolume centers on +/- 0 db, so the LFO boosts volume at phase peak and attenuates at trough.

The solution is to add an initialAttenuation modulator in tandem with the modToVolume modulator.

src/main/components/instr/VGMInstrSet.cpp
mikelow: I'll add a comment in the code.

I'll break it down line by line here:

**`addFrequencyGenerator(ModDest::ModLfoFreq, minHertz);`** 

This sets the Mod LFO frequency to the lowest possible frequency in the CapcomSnes driver (1000/16384 Hz or 0.061 Hz)

**`addFrequencyRangeModulator(ModSource::ChannelPressure, ModDest::ModLfoFreq, minHertz, maxHertz);`**

This creates the ChannelPressure -> Mod LFO frequency modulator. The greater the channel pressure, the faster the rate of tremolo. We pass the helper the minimum and maximum frequencies possible in the driver. (0.061 and 0.061 * 255). The helper converts this into a (confusingly-named) "cents" value - a unit of relative frequency. We're saying: "Given that the base frequency is (minHertz), create a modulator that makes the LFO hit (maxHertz) at full blast."

**addAttenuationModulator(ModSource::ChorusSend, ModDest::ModLfoToVol, maxDepthDb);**

I can immediately see the confusion here. You might think this is setting initialAttenuation. It's not. "Attenuation" here refers to the unit. We should rename this to something like `addDecibelModulator()`. These helpers don't dictate _which_ ModDest gets used, they just let us add modulators using more natural units. 

This line is adding the CC93 -> Tremolo Depth modulator. When CC93 is 127 (max), the tremolo will jump between +/- maxDepthDb.

```
 if (attenOnly) {
   addAttenuationModulator(ModSource::ChorusSend, ModDest::InitialAtten, maxDepthDb);
 }
 ```
 
This time, the method actually is creating a modulator for initialAttenutation. The reasoning was explained in the prior comment - CapcomSnes tremolo never boosts volume. This compensates for SF2 / DLS boosting volume.

File: src/main/components/instr/VGMInstrSet.cpp (hunk 3 above)
Comment (mikelow) [reply src/main/components/instr/VGMInstrSet.cpp]: I've updated the code with:

```
// BipolarAroundNominal:
//   SF2 modLfoToVolume is centered around nominal gain. Tremolo can boost above the note's normal volume.
// NoBoost:
//   Adds a matching initialAttenuation offset. Loudest tremolo point is nominal gain; all other points attenuate.

enum class TremoloGainMode {
  BipolarAroundNominal,
  NoBoost,
};
```

```
void addStandardTremoloHandling(double maxDepthDb,
                                double minHertz,
                                double maxHertz,
                                TremoloGainMode gainMode);
```

**Comment by mikelow** [comment]
I've simplified the interface and cleaned up naming:

- ParamAmount -> ModAmount.
- The ParamAmount helpers were renamed cents() -> fromCents(), hertz() -> fromHertz(), etc.
- I removed the VGMInstr helpers:
```
  void addGlobalGeneratorCents(ModDest destination, double cents);
  void addGlobalGeneratorHertz(ModDest destination, double hertz);
  void addGlobalGeneratorSeconds(ModDest destination, double seconds);
  void addGlobalGeneratorDecibels(ModDest destination, double decibels);
```
and
```
  void addPitchModulator(ModSource source, ModDest destination, double cents);
  void addFrequencyRangeModulator(ModSource source, ModDest destination, double minHertz, double maxHertz);
  void addDelayModulator(ModSource source, ModDest destination, double seconds);
  void addAttenuationModulator(ModSource source, ModDest destination, double decibels);
```
Now we just call addGlobalGenerator() or addModulator() and use the unit conversion helpers inline.

**Comment by mikelow** [comment]
More renaming:

- InstrumentModulation.h/.cpp -> Modulation.h/.cpp
- VGMInstr::addGlobalGenerator -> addGenerator()
- VGMInstr::m_globalGenerators -> m_generators
- InstrumentGenerator -> SynthGenerator
- InstrumentModulator -> SynthModulator

File: src/main/components/instr/VGMInstrSet.cpp (hunk 4 above)
Comment (mikelow) [reply src/main/components/instr/VGMInstrSet.cpp]: As I mentioned further below, I did a bit of refactoring and renaming. This method is now:

```
void VGMInstr::addStandardTremoloHandling(double maxDepthDb,
                                         double minHertz,
                                         double maxHertz,
                                         TremoloGainMode gainMode) {
  addGenerator(ModDest::ModLfoFreq, ModAmount::fromHertz(minHertz));
  addModulator(ModSource::ChannelPressure, ModDest::ModLfoFreq, ModAmount::fromHertzRange(minHertz, maxHertz));
  addModulator(ModSource::ChorusSend, ModDest::ModLfoToVol, ModAmount::fromDecibels(maxDepthDb));
  if (gainMode == TremoloGainMode::NoBoost) {
    addModulator(ModSource::ChorusSend, ModDest::InitialAtten, ModAmount::fromDecibels(maxDepthDb));
  }
}
```

The explanation I gave above still applies, but hopefully the removal of the addXGenerator() and addXModulator() helper methods makes this clearer.

src/main/components/instr/VGMInstrSet.cpp
sykhro: > the CapcomSnes tremolo only ever attenuates the volume - it does not boost it. SF2 and DLS modToVolume centers on +/- 0 db, so the LFO boosts volume at phase peak and attenuates at trough

ah! I get it now

sykhro (approved):
much better!

PR #857, opened 2026-05-06
Fixes for issues flagged by UBSan

sykhro:
## Description
- Prevent a misaligned write in RiffFile (undefined behaviour, problematic on arm)
  - I'm aware of the endianess issue but I think the codebase is already riddled with "machine is little endian" assumptions
- Fixed a bug where we were clearing the wrong buffer in RiffFile
- Fix a crash in closing a RawFile while HexView is open
- Fixed a bad cast in the item tree view

## Motivation and Context
I've started using UBSan and ASAN at all times while in debug mode. I find these can be especially useful when AI tools are employed, as machine-generated C++ can have subtle memory bugs. These came up while testing other things. 

## How Has This Been Tested?

Tried a bunch of conversions. No longer triggers ubsan.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/conversion/RiffFile.cpp
+++ b/src/main/conversion/RiffFile.cpp
@@ -31,8 +31,12 @@ void Chunk::setData(const void *src, uint32_t datasize) {
 
 void Chunk::write(uint8_t *buffer) {
   uint32_t padsize = paddedSize(m_size) - m_size;
+
   memcpy(buffer, id, 4);
-  *reinterpret_cast<uint32_t*>(buffer + 4) = m_size + padsize; // Microsoft says the chunkSize doesn't contain padding size, but many software cannot handle the alignment.
+
+  uint32_t value = m_size + padsize;
+  memcpy(buffer + 4, &value, sizeof(value));
+
   memcpy(buffer + 8, data, paddedSize(m_size));
 }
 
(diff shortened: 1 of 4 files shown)


mikelow:
🎉  Amazing. Thank you for tracking down these bugs!

>   - I'm aware of the endianess issue but I think the codebase is already riddled with "machine is little endian" assumptions

Yeah, probably. 😬

> I've started using UBSan and ASAN at all times while in debug mode. I find these can be especially useful when AI tools are employed, as machine-generated C++ can have subtle memory bugs. These came up while testing other things.

Thanks for bringing these tools to my attention. I'll give them a try. Also happy to see an app that makes the name "VGMTrans" seem slightly more normal.

mikelow (approved):

mikelow:
@codex review

sykhro:
> Thanks for bringing these tools to my attention. I'll give them a try.

I have a local cmake preset with `-fsanitize=address,undefined` to `CFLAGS` and `CXXFLAGS` vars in the `environment` section.
`-fsanitize=thread` is also very useful, unfortunately it can't be used along with those so it has to be a separate build.
Usage is straightforward, use the app normally and it will log problems to console. ASAN/TSAN violations trigger an `abort()` by default so you can't miss those.

A word of warning: if you're using it on Windows, you might want to compile with MSVC (preferably from VS2026) to spare yourself some pain. If you use clang, you might need to add some runtime libraries to the linker CLI manually. It also doesn't work (well or at all) with debug mode CRT

PR #858, opened 2026-05-06
CapcomSnes - update V2+ volume calculation

mikelow:
The prior volume calculation for V2+ was only using the bottom half of the volume table. This meant both that volumes were lower than they should be and that we were distorting the non-linear curve of the volume table. I found that this new implementation is used even in Capcom's last Super Famicom title.

Like before, I used @loveemu's [reverse engineering work on the CapcomSnes driver](https://github.com/loveemu/vgm-disasm/tree/master/snes/Capcom).

### _Update_

I've now added:
```
void addMasterVol(u32 offset, u32 length, double volPercent, Resolution res, const std::string &sEventName = "Master Volume");
```
The equivalent to the addVol(...) overloaded method which accepts a double for volume and a resolution parameter. I've also added a corresponding constructor for `MastVolSeqEvent`.

We need to route the CapcomSnes master volume values through the same functions as channel volume. Since I've updated calculateVolV2 to return a double instead of a 7-bit uint for better precision, it follows that we need to support actually added a floating point master volume through the SeqTrack addX helper methods.

## How Has This Been Tested?
Tested a range of titles. No apparent regression.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/components/seq/SeqTrack.h
+++ b/src/main/components/seq/SeqTrack.h
@@ -198,7 +198,7 @@ private:
   void insertExpression(uint32_t offset, uint32_t length, uint8_t level, uint32_t absTime, const std::string &sEventName = "Expression");
   void insertExpressionNoItem(uint8_t level, uint32_t absTime);
   void addMasterVol(uint32_t offset, uint32_t length, uint8_t vol, const std::string &sEventName = "Master Volume");
-  void addMasterVol(u32 offset, u32 length, double volPercent, Resolution res, const std::string &sEventName = "Master Volume");
+  void addMasterVol(uint32_t offset, uint32_t length, double volPercent, Resolution res, const std::string &sEventName = "Master Volume");
   void addMasterVolNoItem(uint8_t newVol);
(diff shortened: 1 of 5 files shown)


mikelow:
@codex review

src/main/components/seq/SeqTrack.h
```diff
@@ -198,6 +198,7 @@ class SeqTrack : public VGMItem {
   void insertExpression(uint32_t offset, uint32_t length, uint8_t level, uint32_t absTime, const std::string &sEventName = "Expression");
   void insertExpressionNoItem(uint8_t level, uint32_t absTime);
   void addMasterVol(uint32_t offset, uint32_t length, uint8_t vol, const std::string &sEventName = "Master Volume");
+  void addMasterVol(u32 offset, u32 length, double volPercent, Resolution res, const std::string &sEventName = "Master Volume");
```
sykhro: for consistency, `uint32_t“ in place of `u32`

sykhro (changes requested):
just a nitpick

sykhro (approved):

PR #859, opened 2026-05-07
Add basic support for Legato Pedal events (CC68)

mikelow:
This adds support for creating legato pedal events via a new LegatoPedalEvent subclass of MidiEvent and a new SeqTrack::addLegatoPedalNoItem(...) helper method.

The legato pedal is used to indicate that notes should sound connected instead of newly struck. Unfortunately, neither the MIDI standard nor the SoundFont 2 spec define specifically how it should be interpreted for synthesis. It seems a common interpretation is that the attack phase of an envelope should be skipped. This aligns with well with the behavior of some sound drivers we support, like CapcomSnes and CPS. 

Unfortunately, the default fluidsynth implementation only behaves in this way when legato pedal is down or portamento is enabled and notes overlap. I believe there may be a fluidsynth setting to change this behavior, but it remains a compatibility concern.

Regardless, I think it makes sense to add support for legato pedal events and to add them where they are clearly appropriate.

## Motivation and Context
CapcomSnes has a slur state that this aligns with.

## How Has This Been Tested?
Verified that MIDI events are being created as expected.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/conversion/MidiFile.h
+++ b/src/main/conversion/MidiFile.h
@@ -415,8 +415,8 @@ public:
 class LegatoPedalEvent
     : public ControllerEvent {
 public:
-  LegatoPedalEvent(MidiTrack *prntTrk, uint8_t channel, uint32_t absoluteTime, uint8_t bOn)
-      : ControllerEvent(prntTrk, channel, absoluteTime, 68, (bOn) ? 0x7F : 0, PRIORITY_MIDDLE) { }
+  LegatoPedalEvent(MidiTrack *prntTrk, uint8_t channel, uint32_t absoluteTime, bool bOn)
+      : ControllerEvent(prntTrk, channel, absoluteTime, 68, (bOn) ? 0x7F : 0, PRIORITY_HIGH) { }
   MidiEventType eventType() override { return MIDIEVENT_LEGATOPEDAL; }
 };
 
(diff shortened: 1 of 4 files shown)


src/main/conversion/MidiFile.h
```diff
@@ -409,6 +412,14 @@ class PortamentoControlEvent
   uint32_t writeEvent(std::vector<uint8_t> &buf, uint32_t time) override;
 };
 
+class LegatoPedalEvent
+    : public ControllerEvent {
+public:
+  LegatoPedalEvent(MidiTrack *prntTrk, uint8_t channel, uint32_t absoluteTime, uint8_t bOn)
```
sykhro: is `bOn` intentionally a u8 here?

src/main/conversion/MidiFile.h
```diff
@@ -409,6 +412,14 @@ class PortamentoControlEvent
...
+      : ControllerEvent(prntTrk, channel, absoluteTime, 68, (bOn) ? 0x7F : 0, PRIORITY_MIDDLE) { }
```
sykhro: is `PRIORITY_MIDDLE` alright? I think serialization order for this event might be important

sykhro (changes requested):
couple comments

src/main/conversion/MidiFile.h
mikelow: Yeah, I'll make it a bool. The event interprets 0-63 as off and 64-127 as on. No reason our interface should be more than binary.

mikelow: And also, didn't even notice the more obvious issue of the `bOn` naming being inconsistent with the type.

src/main/conversion/MidiFile.h
mikelow: I'll make it PRIORITY_HIGH. The important thing is that its priority is higher than note priority, which is PRIORITY_LOWER. Can't think of other conflicts, but HIGH feels more correct.

sykhro (approved):

PR #860, opened 2026-05-07
CapcomSnes - make slurred notes overlap, and add legato pedal

mikelow:
This adds an extra tick to note duration when the slur flag is enabled. We do this so that connected notes will overlap, thus triggering the legato effect where the attack phase of the new note is skipped. In other words, the notes slide into each other. The driver anticipates that slurred notes will blend together, as it forces their duration to 100%, so this is not quite as hacky as it may seem.

Still, It's possible for rests to exist between slurred notes, in which case the driver still skip the attack phase of the new note. Unfortunately we can't support this short of forcing fluidsynth to use a different legato pedal mode. I expect this is pretty rare, though.

We're also adding legato pedal on and off events corresponding to slur on/off events. To clarify about behavior, the legato pedal isn't necessary to skip the attack phase when portamento (or, in our case, portamento control) is used, but it should matter if slur is enabled but portamento is not.

And of course, none of this works well in BASS MIDI.

## Audio demo

The effect is subtle. Try to spot the difference.

master

https://github.com/user-attachments/assets/20266076-f714-4aa0-8664-22bc252f21da

this branch

https://github.com/user-attachments/assets/31d774e7-a6ac-4b85-8127-ef0104d2d645

original

https://github.com/user-attachments/assets/555c877c-4fe6-424a-a9ff-a423a82fd810

## How Has This Been Tested?
Tested numerous CapcomSnes titles.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/formats/CapcomSnes/CapcomSnesSeq.cpp
+++ b/src/main/formats/CapcomSnes/CapcomSnesSeq.cpp
@@ -402,6 +402,15 @@ bool CapcomSnesTrack::readEvent() {
   bool bContinue = true;
 
   std::string desc;
+  auto applyNoteAttributes = [this](uint8_t attributes) {
+    const bool wasSlurred = isNoteSlurred();
+    noteAttributes &= ~(CAPCOM_SNES_MASK_NOTE_OCTAVE_UP | CAPCOM_SNES_MASK_NOTE_TRIPLET | CAPCOM_SNES_MASK_NOTE_SLURRED);
+    noteAttributes |= attributes;
+
+    if (isNoteSlurred() != wasSlurred) {
+      addLegatoPedalNoItem(isNoteSlurred());
+    }
+  };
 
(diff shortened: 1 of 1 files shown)


sykhro:
> And of course, none of this works well in BASS MIDI.

Funnily enough, I ran into issues while trying to do this for NDS when listening back on fluidsynth. I need to update my PR to use these new APIs

> The effect is subtle. Try to spot the difference.

I can hear it pretty distinctively at 0:10!

sykhro (approved):

mikelow:
> Funnily enough, I ran into issues while trying to do this for NDS when listening back on fluidsynth. I need to update my PR to use these new APIs

You can use the juce3 branch for in-app fluidsynth rendering. It's a bit of a mess, seek often breaks, and I'll probably majorly refactor or rewrite everything before bringing it to the codebase, but for testing features that don't work on BASS it's helpful.

PR #861, opened 2026-05-08
HexView Refactor 12: switch to Vulkan for Linux, add scissor calls, remove hexedge.frag

mikelow:
This is the final HexView-specific change for the `codex/piano-roll-view` merge project. It's not even that HexView-specific. 

- This switches Linux to use Vulkan instead of OpenGL. Now that we're creating our own title bar via qwindowkit, the prior issue we ran into from using vulkan - the window not being given a title bar - is obviated. 
- However, with this change alone, Vulkan renders blank views. Vulkan requires that we set scissor state, so I've added the required setScissor calls. My understanding is that this is best practice anyway.
- We now build the hidden QRhiWidget primer detached and set its parent to the main window _after_ setting  API to Vulkan. This ensures the backend/window isn't recreated after API selection. We do the same deferred parent assignment within HexViewRhiWidget.
- hexedge.frag is now unused, so we remove it.

### UPDATE

Codex brought to my attention that Vulkan setup may fail at runtime, in which case we should fall back to OpenGL. I've updated the code to do just that. The graphics API selection is now offloaded to a new file RhiApiSelector.h with the function `QRhiWidget::Api preferredRhiWidgetApi()`. This performs all of the checks for determining API. Additionally, on Linux, it will only return Api::Vulkan after successfully creating a QVulkanInstance, otherwise it falls back to OpenGL if viable.

## Types of changes

- [x] Refactor

## Checklist:


mikelow:
@codex review

mikelow:
I've decided we should just stick to OpenGL as default. I'll make another PR for the smaller changes here worth keeping.

PR #865, opened 2026-05-10
Vibrato delay modulator support

mikelow:
This extends the modulation model with support for the vibLfoDelay destination. ModAmount now provides the helper `fromSecondsRange(...)` to easily specify an 'amount' value for the modulator's range, and `midiValueForSecondsInRange(...)` generates controller values for sequence conversion.

Vibrato delay range is made an optional parameter to `addStandardVibratoHandling(...)`. When provided it is mapped CC93 (for now).

This also adds L_WARN() calls to the ModAmount helpers when values cannot be converted.

DLS does not support vibrato delay, so we ignore it for DLS conversion.

## Motivation and Context
Vibrato delay is used by KonamiSnes, and probably other drivers.

## How Has This Been Tested?
This was put into practice on the KonamiSnes format. Resulting values were tested there.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/components/instr/Modulation.cpp
+++ b/src/main/components/instr/Modulation.cpp
@@ -19,10 +19,6 @@ int32_t hertzToInstrumentCents(double hertz) {
   return static_cast<int32_t>(std::lround(1200.0 * std::log2(hertz / kSf2LfoReferenceHz)));
 }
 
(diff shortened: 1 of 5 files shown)


sykhro (approved):
this is great! I'm working on modulators for NDS, needed something like this

PR #870, opened 2026-05-13
Add Motion and Automation classes to simplify controller fades, pitch slides and vibrato/tremolo

mikelow:
## Purpose

Many sequence formats encode playback changes as stateful motion, not one-off events: pitch slides, pan/volume/tempo fades, delayed vibrato, vibrato fade-ins, and tremolo output gating. Drivers are often complex enough that we must effectively emulate them tick-by-tick, but this mandates bookkeeping of state structs and per-tick advancement logic.

That makes format code do two jobs at once:

- parse what the original driver command means
- manage how that command evolves over time and becomes MIDI output

This PR aims to simplify the bookkeeping. This adds a shared sequence automation layer so format code can describe driver behavior in source-format terms, while reusable helpers handle the mechanical parts: delays, per-tick advancement, target snapping, fixed-point interpolation, MIDI pitch-bend conversion, LFO output gating, and duplicate MIDI event suppression.

The goal is to make tick-by-tick format code easier to read and write, especially as we add more support for  stateful motion behavior.

## Summary

- Adds generic motion primitives in `SeqMotion.h`.
- Adds reusable automation helpers in `SeqAutomation.h`.
- Adds MIDI-facing automation helpers in `SeqMidiAutomation.h`.
- Adds `SeqTrack` bridge helpers for pitch bend automation and synth LFO output.
- Replaces Konami SNES custom fade/slide state with shared automation classes.
- Uses shared LFO automation in Capcom SNES for vibrato/tremolo depth and output gating.
- Adds shared vibrato/tremolo modulation specs for standard instrument LFO handling.

## Concept

The shared automation classes are built around a simple idea: a track or sequence owns a small state object for each value that can change over time. A format stores one automation object per active behavior, such as pitch bend, pan fade, volume fade, tempo fade, vibrato, or tremolo.

The intended usage pattern is:

1. Keep an automation object as member state.
2. When a command is parsed, translate the command into a small motion plan or automation configuration.
3. On each tick, advance the automation object.
4. Let the automation object decide whether anything changed enough to emit MIDI output.

This keeps format code focused on source-format semantics. The format still decides what a command means, but it no longer has to repeatedly reimplement the lifecycle of “start motion, wait through delay, advance each tick, clamp or snap at the end, and emit only when needed.”

## Details

`SeqMotion.h` contains the lowest-level motion primitive. `SeqMotionPlan` describes a value moving to a target, either over a tick count, over a tick count with an explicit step, or by step until the target is reached. `SeqLinearMotion` owns the active motion state and returns a `SeqMotionTick` on each advance so callers can distinguish inactive, delayed, running, and finished states.

`SeqAutomation.h` builds on that primitive for reusable non-MIDI automation. `SeqAutomatedValue` wraps a moving value with convenience `begin` and `tick` helpers. `SeqFixedPointAutomation` supports source formats whose controller values are raw integers but whose driver motion uses fractional accumulation. Format code can keep working in raw units while the helper stores fixed-point state internally, applies the configured rounding policy, and emits raw output only when the quantized raw value changes.

`SeqMidiAutomation.h` contains MIDI-specific automation. `SeqPitchBendAutomation` tracks an active note’s base pitch, advances pitch motion in source-space units, manages pitch bend range changes, converts the current offset to MIDI pitch bend, clamps to MIDI limits, and suppresses duplicate bend/range output. `SeqSynthLfoAutomation` tracks configured LFO delay/rate/depth separately from the currently emitted MIDI depth, supports reusable per-note fade-ins, and handles gated output without throwing away the stored LFO settings.

`SeqTrack` adds small bridge helpers that connect these automation objects to the existing MIDI emitters, such as `addPitchBendNoItem`, `addPitchBendRangeNoItem`, `addModulationNoItem`, and controller output.

Konami SNES now uses the shared automation classes for tempo, pan, volume, pitch slide, and vibrato fade behavior. The format code mostly reads driver commands and constructs motion plans; the shared helpers handle the common tick-by-tick mechanics.

Capcom SNES now uses `SeqSynthLfoAutomation` for vibrato and tremolo state. This keeps LFO rate/depth configuration separate from output gating, which better matches the driver behavior where rate zero disables output without clearing the stored vibrato or tremolo depth.

The modulation spec structs, `VibratoModulationSpec` and `TremoloModulationSpec`, provide a compact way for formats to describe how their native LFO behavior maps into VGMTrans’ standard instrument modulation setup. Capcom SNES and Konami SNES now keep those calibration details near their format definitions and pass normalized specs into the shared instrument helpers.

## How Has This Been Tested?
Examining MIDI data output, and listening. I've tested many tracks across many both KonamiSnes and NinSnes (future PR pending).

## Types of changes

- [x] Refactor

## Checklist:


sykhro:
I will look at this asap. I wonder if #867 could benefit from this

mikelow:
I'm considering refactoring this further. I've set this to a draft while I work on it. Sorry for the inconvenience!

mikelow:
Ok, updated and ready for review ✊🪵

src/main/components/instr/Modulation.h
```diff
@@ -7,6 +7,33 @@
...
+struct VibratoModulationSpec {
+  double maxDepthCents;
+  double minHertz;
+  double maxHertz;
+  std::optional<DelayRange> delayRange = std::nullopt;
+};
+
+enum class TremoloGainMode {
+  // SF2 modLfoToVolume is centered around nominal gain, so tremolo can boost above normal volume.
+  BipolarAroundNominal,
```
sykhro, line 26: not too sure what this means

mikelow: nominal here just means near-zero, as in the nominal gain point is where there is no change. So when tremolo is bipolar, it can be positive (boost gain) or negative (attenuation). The Capcom driver uses tremolo in a fashion that only attenuates the signal, hence why I added this option.

sykhro (approved):
interesting direction. I will consider rebasing my nds changes on top of this

mikelow:
> interesting direction. I will consider rebasing my nds changes on top of this

I am not married to this approach at all, so if you have ideas for a different direction, or criticisms for this paradigm, I would love to hear them.

sykhro:
> > interesting direction. I will consider rebasing my nds changes on top of this
> 
> I am not married to this approach at all, so if you have ideas for a different direction, or criticisms for this paradigm, I would love to hear them.

honestly, I don't know :) I like having an API to solve the problem but I can't be 100% sure about the shape until we actually try to use it in translating different formats. I think it's fine to go in as-is!

mikelow:
> honestly, I don't know :) I like having an API to solve the problem but I can't be 100% sure about the shape until we actually try to use it in translating different formats. I think it's fine to go in as-is!

Agreed. That's my biggest concern. I put the PR in draft mode earlier to try and make this approach more generalized. I would love to have an approach that's more declarative, though I expect that the idiosyncrasies of each format would make that very hard.

Just know that I'm totally open to a different approach, and I'm willing to throw this one out if we find a better one.

PR #881, opened 2026-05-21
Update changelog layout

sykhro:
## Description
This PR restructures and updates `CHANGELOG.md` because every time I need to add an item I end up asking myself phylosophical questions like "is this fix for a completely broken impl even a fix anymore?" something something ship of Theseus but for bugs.

The changelog has been reorganized into domain-specific categories (Audio Formats & Engines, Core & Playback Workflow, User Interface & Analysis Tools, Platform & Deployment). I'm by no means attached to any of them but I think it's important to keep the audio formats one at the top. 

## Motivation and Context
The previous changelog format buried audio driver updates under generic headers.

## How Has This Been Tested?
Verified that the modified layout parses properly.
Confirmed the flat list structure complies with `build-changelog.py` limitations (no multi-level nested bullet points).

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,41 +1,77 @@
 # Changelog
 
-
 ## [Unreleased]
 
-### Added
+### Music Formats
 
-- Added an in-app dialog to help with reporting bugs.
-- Added Flatpak builds for Linux. From now on, stable releases will be automatically available on Flathub.
-- Added stitched export for collection chunks: drag collections into a stitch queue, reorder them, and export as a merged MIDI + SF2.
-- Added a search field to the Collections panel
-- Added drum kit, vibrato, and pitch slide support to the standard N-SPC driver (Nintendo's SNES SDK driver).
-- Added drum kit, portamento, and slide support for volume/pan/pitch to the KonamiSnes driver.
-- Added vibrato, tremolo, more accurate portamento support, and fixed pan calculation in the CapcomSnes driver.
-- Added drum kit support to the SuzukiSnes driver (Seiken Densetsu 3, Super Mario RPG, etc)
-- Improved support for late-era Intelligent Systems SNES/SFC titles (Tetris Attack, Fire Emblem 4, Super Famicom Wars) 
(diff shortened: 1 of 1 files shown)


mikelow:
I think this is much cleaner. I have nitpicks around the section titles:
- "Audio Formats & Engines" - "& Engines" seems redundant. Torn between "Audio Formats" and "Music Formats"
- "Core & Playback Workflow" - this one is hard. Perhaps we rename it to "General", making it a catch-all for items that don't fit anywhere else. I would  move "Stitched Export" to the next section, and make this the last section listed.
- "User Interface & Analysis Tools" - maybe just "User Interface & Tools"

Even so, this doesn't seem perfect 🤷‍♂️. Let me know your thoughts.

sykhro:
>     * "Audio Formats & Engines" - "& Engines" seems redundant. Torn between "Audio Formats" and "Music Formats"

I think "Music Formats" is the best fit

>     * "Core & Playback Workflow" - this one is hard. Perhaps we rename it to "General",  making it a catch-all for items that don't fit anywhere else.

We can't run from this I guess. I'm not opposed to making it "General"

> I would  move "Stitched Export" to the next section, and make this the last section listed.

Even after "Deployment & Whatever"? see I already forgot, I think I'll make that "Platform Specific Notes" instead

>     * "User Interface & Analysis Tools" - maybe just "User Interface & Tools"

Makes sense, less confusing for users

mikelow:
> We can't run from this I guess. I'm not opposed to making it "General"

I don't love it either. We could make a "Stability" section, but then what do you do with something like Hot Keys? Throw it in the UI section? Maybe, actually.

> Even after "Deployment & Whatever"? 

Yeah, maybe make "Deployment & Whatever" last.

> "Platform Specific Notes" instead

That works. More flexible for future changes.

sykhro:
> Throw it in the UI section? Maybe, actually.

yeah that would work. I'll integrate the changes from the latest readme + wait until the GBA PR goes in to update this

PR #882, opened 2026-05-21
MP2K: Tempo & command accuracy, general improvements

### LINKED ISSUE #615: [MP2k] importing from a rom does not label songs based on song number
All of them are "MP2kSeq"

Image

### LINKED ISSUE #875: MP2k: Tempo is about 0.45% faster than real GBA hardware
### Where does the problem occur?

Inside VGMTrans & With exported MIDI

### System details

VGMTrans version: v1.3 (e3a5652a, master)
Operating system: Windows 11 Version 25H2

### Affected input files

_No response_

### Describe the problem

While it's commonly believed that MusicPlayer2000's tempo values should be interpreted as `bpm / 2`, this is only accurate when the GBA's frame rate is exactly **60.000 fps**.  On actual GBA hardware, the frame rate is approximately **59.727 fps** (16777216/280896).  As a result, when interpreting BPM as `value * 2`, the playback tempo becomes slightly faster than on real hardware (approximately **0.45%** faster).

Whether the current behavior of VGMTrans is by design or an issue may be a matter of opinion, but I wanted to report it here just in case.  Personally, I prefer perfect tempo synchronization, since it also makes A/B comparisons with actual hardware audio much easier.  It also makes A/B comparisons with actual hardware audio much easier.  Also, the difference is really small, but sometimes I can still kind of feel it.

### Steps to reproduce

Load and play GBA songs, export to MIDI if necessary.

Note: I often use **Castlevania: Harmony of Dissonance – 0023 (Juste Belmont's Theme Variation)** for tempo testing.  This track is implemented using streamed audio, and it switches between two audio segments during playback, which makes it convenient for checking overall timing and total length.

### Expected behavior (technical description required)

Actual:
Tempo value of 60 is converted to BPM 120.000000.

Expected:
Tempo value of 60 is converted to BPM 119.455001.

### Logs

_No response_

sykhro:
## Description
This PR fills in all the MP2K commands I couldn't be bothered implementing 4 years ago: pan, transpose, pitch bend (range), LFO modes, [...] while also fixing two issues: BPM being subtly wrong and release times sometimes being as long as MINUTES.

Echo gets tied to a reverb CC. It won't sound correct in-program but the idea is that one may use a GBA-specific reverb VST (I have it as a M4L device, I could totally make a standalone though. and perhaps even support it in VGMTrans? feature creep much? I'm glad I'm not a project manager)

I've also moved the square synthesis out of NDS code to re-use it here for PSG instruments. The same code will also be used in PSG instruments for Super Princess Peach (WIP and so specific that I'm unsure if I'll ever end up merging that!) and the DSE branch.
This is the kind of thing the future VST could replace but for now there's value in having it in the soundfonts.

There's some cleanups too. Mostly getting rid of global constants and not registering instruments if they're invalid (we keep an index of the latest valid instrument instead).

## Motivation and Context
Improving MP2K support.
- Fixes #875 
- Fixes #615 

## How Has This Been Tested?
Mostly listening to conversions Pokemon Leaf Green.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/components/seq/SeqTrack.h
+++ b/src/main/components/seq/SeqTrack.h
@@ -411,3 +411,5 @@ template<typename... Args>
 static std::string describeUnknownSubevent(uint8_t statusByte, Args... args) {
   return logEvent(statusByte, spdlog::level::off, "Unknown Subevent", args...);
 }
+
+#include "automation/SeqTrackAutomation.h"
--- a/src/main/components/seq/SeqTrack.h
+++ b/src/main/components/seq/SeqTrack.h
@@ -411,5 +411,3 @@ template<typename... Args>
 static std::string describeUnknownSubevent(uint8_t statusByte, Args... args) {
   return logEvent(statusByte, spdlog::level::off, "Unknown Subevent", args...);
 }
-
-#include "automation/SeqTrackAutomation.h"
--- a/src/main/formats/MP2k/MP2kInstrSet.cpp
+++ b/src/main/formats/MP2k/MP2kInstrSet.cpp
@@ -19,41 +19,34 @@
 #include "PSGDSP.h"
 
 namespace {
-constexpr double kPsgDutyRatios[4] = {0.125, 0.25, 0.5, 0.75};
-constexpr const char* kPsgDutyLabels[4] = {"12.5%", "25%", "50%", "75%"};
 constexpr uint8_t kPsgSquareCount = 4;
 constexpr uint8_t kPsgNoiseIndex = kPsgSquareCount;
 constexpr uint8_t kPsgUnityKey = 69;
 constexpr size_t kGbaRomPointerMask = 0x03FFFFFF;
-constexpr size_t kGbaRomAddressMask = 0x0E000000;
-constexpr size_t kGbaRomAddressMin = 0x08000000;
-constexpr size_t kGbaRomAddressMax = 0x0C000000;
-constexpr uint8_t kCgbWaveRamBytes = 16;
-constexpr uint8_t kCgbWaveRamSamples = kCgbWaveRamBytes * 2;
-constexpr uint32_t kCgbWaveSampleRate = kCgbWaveRamSamples * 440;
-constexpr double kMp2kEnvelopeTicksPerSecond = 60.0;
-constexpr double kMp2kEnvelopeMaxLevel = 256.0;
-constexpr double kCgbEnvelopeClockHz = 64.0;
-constexpr double kCgbEnvelopeMaxLevel = 15.0;
 
 double mp2kAttackTimeSeconds(int attack) {
   if (attack <= 0) {
     return 0.0;
   }
-  return kMp2kEnvelopeMaxLevel / (kMp2kEnvelopeTicksPerSecond * attack);
+  constexpr double kEnvelopeTicksPerSecond = 60.0;
+  constexpr double kEnvelopeMaxLevel = 256.0;
+  return kEnvelopeMaxLevel / (kEnvelopeTicksPerSecond * attack);
 }
 
 double mp2kEnvelopeRateTimeSeconds(int rate) {
   if (rate <= 0) {
     return 0.0;
   }
-  const double logMaxLevel = std::log(kMp2kEnvelopeMaxLevel);
+  constexpr double kEnvelopeTicksPerSecond = 60.0;
+  constexpr double kEnvelopeMaxLevel = 256.0;
+  const double logMaxLevel = std::log(kEnvelopeMaxLevel);
   return logMaxLevel / (logMaxLevel - std::log(static_cast<double>(rate))) /
-         kMp2kEnvelopeTicksPerSecond;
+         kEnvelopeTicksPerSecond;
 }
 
 double mp2kDecayTimeSeconds(int decay) {
-  return mp2kEnvelopeRateTimeSeconds(decay) * 10.0 / std::log(kMp2kEnvelopeMaxLevel);
+  constexpr double kEnvelopeMaxLevel = 256.0;
+  return mp2kEnvelopeRateTimeSeconds(decay) * 10.0 / std::log(kEnvelopeMaxLevel);
 }
 
 double cgbEnvelopeRateTimeSeconds(int rate) {
@@ -61,20 +54,26 @@ double cgbEnvelopeRateTimeSeconds(int rate) {
   if (rate == 0) {
     return 0.0;
   }
-  return kCgbEnvelopeMaxLevel * rate / kCgbEnvelopeClockHz;
+  constexpr double kEnvelopeClockHz = 64.0;
+  constexpr double kEnvelopeMaxLevel = 15.0;
+  return kEnvelopeMaxLevel * rate / kEnvelopeClockHz;
 }
 
 double cgbSustainLevel(int sustain) {
   if (sustain <= 0) {
     return 0.0;
   }
-  if (sustain >= kCgbEnvelopeMaxLevel) {
+  constexpr double kEnvelopeMaxLevel = 15.0;
+  if (sustain >= kEnvelopeMaxLevel) {
     return 1.0;
   }
-  return sustain / kCgbEnvelopeMaxLevel;
+  return sustain / kEnvelopeMaxLevel;
 }
 
 bool isGbaRomPointer(size_t pointer) {
+  constexpr size_t kGbaRomAddressMask = 0x0E000000;
+  constexpr size_t kGbaRomAddressMin = 0x08000000;
+  constexpr size_t kGbaRomAddressMax = 0x0C000000;
   size_t address_space = pointer & kGbaRomAddressMask;
   return address_space >= kGbaRomAddressMin && address_space <= kGbaRomAddressMax;
 }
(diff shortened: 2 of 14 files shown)


sykhro:
One more thing! I hate `namespace psg`, `PSGDSP` sounds very silly and borderline nonsensical as well. But I don't know where else to put these things

Bonboon229:
Is the build suppossed to have failed like this?

sykhro:
> Is the build suppossed to have failed like this?

`inline` claims victims. should be fine now

Bonboon229:
Heh, nice to see these improvements as of late!
Doing God's Work!

mikelow:
@codex review

src/main/formats/MP2k/MP2kSeq.cpp
```diff
@@ -266,65 +319,73 @@ void MP2kTrack::handleSpecialCommand(u32 beginOffset, u8 status_byte) {
...
+      addGenericEvent(beginOffset, curOffset - beginOffset, "Modulation Type", desc,
+                      Type::Modulation);
       break;
     }
 
     case 0xC8: {
-      curOffset++;
-      addGenericEvent(beginOffset, curOffset - beginOffset, "Microtune", "", Type::PitchBend);
+      const int16_t tune = static_cast<int16_t>(readByte(curOffset++)) - 0x40;
+      const double cents = static_cast<double>(tune) * (100.0 / 64.0);
```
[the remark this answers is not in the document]
sykhro: There's only 63 positive steps for max finetune, the range is -64 to 63.

mikelow:
> Echo gets tied to a reverb CC. It won't sound correct in-program but the idea is that one may use a GBA-specific reverb VST (I have it as a M4L device, I could totally make a standalone though. and perhaps even support it in VGMTrans? feature creep much? I'm glad I'm not a project manager)

This may already be understood, but my vision is to create a monolithic VST which supports SF2 playback and emulates other sound hardware, assigning each channel to a synth and mixing everything together. The juce3 branch is a proof-of-concept of this.

In the back of my mind, I figure we could add reverb support in some fashion.

sykhro:
> In the back of my mind, I figure we could add reverb support in some fashion.

I was thinking effects could be delegate to separate VSTs but if the goal is replicating how the mixer for these platforms sound then it would be best to add reverb in there. It's pretty easy for the GBA case! Other platforms use custom IRs, I would honestly be excited to write our own convolver.

mikelow:
> I was thinking effects could be delegate to separate VSTs but if the goal is replicating how the mixer for these platforms sound then it would be best to add reverb in there. It's pretty easy for the GBA case! Other platforms use custom IRs, I would honestly be excited to write our own convolver.

The primary goal of making it monolithic is simplicity of use, for our codebase and for users. Importing exported files into a DAW is a major pain point if users have to load multiple VSTs and then figure out which tracks map to which VSTs. Adding reverb VSTs on top of that would be another chore. With a monolithic approach, we can add sysex events in our exported MIDI files that tell our VST which synth to assign to the channel.

Still, I don't know how to gracefully solve handling more than 16 channels on the DAW side without resorting to very ugly hacks or having users manually assign.

sykhro:
> Still, I don't know how to gracefully solve handling more than 16 channels on the DAW side without resorting to very ugly hacks or having users manually assign.

The state of MIDI in many DAWs is far from optimal anyway. FL Studio, for example, often imports MIDI with wrong values despite the MIDI being correct. I think DAW users can (should) deal with some assembly required to make it work for their workflow

src/main/components/seq/SeqTrack.h
```diff
@@ -411,3 +411,5 @@ template<typename... Args>
 static std::string describeUnknownSubevent(uint8_t statusByte, Args... args) {
   return logEvent(statusByte, spdlog::level::off, "Unknown Subevent", args...);
 }
+
+#include "automation/SeqTrackAutomation.h"
```
mikelow: I suppose this means we should remove the `#include "automation/SeqTrackAutomation.h"` lines from CapcomSnes, KonamiSnes, and NinSnes?

sykhro: argh I missed this from an earlier fix, I'll remove it. it would introduce a recursive include and now that the helpers are not defined there it's not really useful

src/main/formats/common/PSGDSP.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "PSGDSP.h"
+
+#include <algorithm>
+#include <cmath>
+#include <iterator>
+#include <numeric>
+
+namespace psg {
+
+std::vector<uint8_t> synthesizeNoiseWave(uint32_t sampleCount, uint16_t lfsrSeed, uint16_t lfsrTap) {
```
mikelow: How much is this implementation specific to the GBA hardware? Should we rename it to synthesizeNoiseWaveGBA(), or if we think it's generalizable, add a comment stating that this emulates the GBA behavior?

src/main/formats/common/PSGDSP.cpp
```diff
@@ -0,0 +1,77 @@
...
+      value ^= lfsrTap;
+    } else {
+      output[i] = 0x7FFF;
+    }
+  }
+
+  return samples;
+}
+
+std::vector<uint8_t> synthesizePulseWave(double dutyCycle, uint32_t sampleRate,
```
mikelow: Same question here as above - is this an emulation of the GBA behavior and do we expect it to differ meaningfully from other PSG hardware behavior? Should we add a comment?

src/main/formats/common/PSGDSP.cpp
sykhro: It's GBA and NDS specific. However, it sounds pretty different since it's not going through the GBA DAC + noisy headphone port/puny speaker

src/main/formats/MP2k/MP2kInstrSet.cpp
```diff
@@ -16,16 +16,108 @@
...
+    if (!instr->loadInstr()) {
+      delete instr;
+      for (size_t j = i + 1; j < instrs.size(); j++) {
+        delete instrs[j];
+      }
+      for (auto *loadedInstr : aInstrs) {
+        delete loadedInstr;
+      }
+      aInstrs.clear();
+      return false;
```
mikelow, line 114: Maybe we should log a warning in this case, notifying that there was an error loading an instrument (and therefore no instruments were loaded).

src/main/formats/common/PSGDSP.cpp
sykhro: No, these are generic band-limited square waves. You can use them for anything, they will most likely sound nicer (less "stray" harmonics) than what most consoles output though

src/main/formats/MP2k/MP2kInstrSet.cpp
mikelow: Actually, this might warrant an error log, not a warning.

src/main/formats/MP2k/MP2kInstrSet.cpp
```diff
@@ -229,24 +347,53 @@ bool MP2kInstr::loadInstr() {
...
+            // rgn->sampCollPtr = static_cast<MP2kInstrSet *>(parInstrSet)->sampColl;
+            setADSR(rgn, rawFile()->get<u32>(off + 8));
+          }
+        } else if (cgb_type == 1 || cgb_type == 2) {
+          if (psgColl) {
+            uint8_t duty = static_cast<uint8_t>(raw_sample_pointer & 0x03);
+            VGMRgn *rgn = addRgn(off, 12, duty, split_list[i], split_list[i + 1] - 1);
+            rgn->sampCollPtr = psgColl;
+            rgn->setUnityKey(kPsgUnityKey);
+            setCgbADSR(rgn, rawFile()->get<u32>(off + 8));
```
mikelow: We repeat these 4 lines two more times below. Wonder if we break refactor them into a lambda expression or something else to reduce repetition.

src/main/formats/common/PSGDSP.cpp
sykhro: Used the Doxygen format out of muscle memory, let me know if you prefer a simpler command format (I don't mind this because it shows up nicely in VSCode hints)

src/main/formats/MP2k/MP2kInstrSet.cpp
```diff
@@ -301,6 +469,13 @@ bool MP2kInstr::loadInstr() {
     }
   }
 
+  if (!regions().empty()) {
+    constexpr double kDefaultLfoFrequencyHz = 60.0 * 22.0 / 256.0;
+    constexpr double kVibratoMaxDepthCents = 127.0 * 4.0 * 100.0 / 64.0;
```
mikelow, line 453: For the other formats using modulation, I moved constants like these into the top of a header file (CapcomSnesDefinitions.h, KonamiSnesDefinitions.h, NinSnesVibrato.h). Do we have other constants that would warrant creating a similar header (the tempo constant comes to mind)? If not, should we move these into an anonymous namespace at the top of this file?

sykhro: There are many constants but most of them aren't shared between code so I didn't see the value in making a separate header. Could go for the anonymous namespace though

src/main/formats/MP2k/MP2kSeq.cpp
```diff
@@ -388,6 +466,123 @@ void MP2kTrack::handleSpecialCommand(u32 beginOffset, u8 status_byte) {
...
+    vibratoLfo.tickFadeToDepth(
+        0,
+        [](int32_t depth) { return depth; },
+        [this](uint8_t depth) { addModulationNoItem(depth); });
+  } else if (modType == kMp2kModTypeTremolo) {
+    tremoloLfo.tickFadeToDepth(
+        0,
+        [](int32_t depth) { return depth; },
+        [this](uint8_t depth) {
+          addControllerEventNoItem(kMp2kTremoloDepthController, depth);
```
mikelow, line 581: I think we'll soon need to revise the underlying system so that the specific MIDI events used for tremolo and vibrato params are user-configurable. Here we would just call addTremoloDepth(), which would sort things out for us.

src/main/formats/MP2k/MP2kInstrSet.cpp
```diff
@@ -301,6 +469,13 @@ bool MP2kInstr::loadInstr() {
...
+    addStandardVibratoHandling(kVibratoMaxDepthCents, kDefaultLfoFrequencyHz,
```
mikelow, line 454: I see that you're adding vibrato depth events via CC93. Is there no corresponding modulator on the instrument side?

mikelow (approved):
Left a few comments here and there, but this LGTM. Didn't dig too deeply into the format-specific logic

src/main/formats/MP2k/MP2kInstrSet.cpp
sykhro: Moved it behind `addPsgRgn`. Will look at it closely later to make sure I didn't accidentally break anything there

src/main/formats/MP2k/MP2kInstrSet.cpp
sykhro: whoops, messed it up after I moved the code to these new things. will fix

mikelow: I miswrote that. I meant to say that you're adding tremolo depth, but I didn't see a corresponding tremolo modulator

loveemu:
I reviewed the 8a48a344 tempo change only. LGTM.

I enjoy seeing how actively the core team members are working. Thank you as always.

PR #885, opened 2026-05-23
Main: Remove opinionated type aliases

sykhro:
## Description
This is the first PR in a series focused on code hygiene.

This removes the global primitive type aliases from `common.h`, including `u8`, `u16`, `u32`, `u64`, `s8`, `s16`, `s32`, `s64`, `sptr`, `uptr`, and the `size_t` alias.

## Motivation and Context

These aliases were opinionated shorthand I added back in 2019 and that opinion has aged poorly. They obscure standard types, make includes and dependencies harder to reason about, and keep more code tied to `common.h` than necessary.

The reason I added them in the first place is that at the time I was doing a lot of embedded and needed my own type aliases against weird SDK things. What does that have to do with VGMTrans? Nothing. Youth mistake.

The longer-term goal is to kill `common.h` entirely and improve header includes (I hate how editing one file results in many of them being rebuilt for no reason!).

## How Has This Been Tested?
It builds, it runs

## Types of changes
- [x] Code cleanup

## Checklist:


mikelow:
> These aliases were opinionated shorthand I added back in 2019 and that opinion has aged poorly. They obscure standard types, make includes and dependencies harder to reason about, and keep more code tied to common.h than necessary.

I actually wanted to do the opposite of this PR, and update our code to use these shorthand types universally.

I like these shorthand type aliases a lot. They help reduce verbosity, are easy to learn, and once learned are clearer than the standard types. Who would argue `uint32_t` is more intuitive than `u32`? I find these standard C++ types comically bad. I also find it easier to mistake `int32_t` for `uint32_t` than it is `s32` for `u32` - I've done that before.

Given my understanding of the downsides, I'm not convinced they outweigh the benefit of keeping the shorthand. Code that spans fewer columns and fewer lines and uses clearer types is easier to mentally parse. We also rely on these bit-width-specific types _a lot_, largely because our format code needs to be precise about bit-width. That's why kernels and game engines use this shorthand in the first place, right?

I agree that common.h has taken on way too much responsibility and is a problem. I would rather gut common.h but keep the shorthand.

But I'm also not a professional C++ developer, and I'm open to getting rid of the type aliases. I would like to better understand why you disagree.

sykhro:
> > These aliases were opinionated shorthand I added back in 2019 and that opinion has aged poorly. They obscure standard types, make includes and dependencies harder to reason about, and keep more code tied to common.h than necessary.
> 
> I actually wanted to do the opposite of this PR, and update our code to use these shorthand types universally.
> 
> I like these shorthand type aliases a lot. They help reduce verbosity, are easy to learn, and once learned are clearer than the standard types. Who would argue `uint32_t` is more intuitive than `u32`? I find these standard C++ types comically bad. I also find it easier to mistake `int32_t` for `uint32_t` than it is `s32` for `u32` - I've done that before.
> 
> Given my understanding of the downsides, I'm not convinced they outweigh the benefit of keeping the shorthand. Code that spans fewer columns and fewer lines and uses clearer types is easier to mentally parse. We also rely on these bit-width-specific types _a lot_, largely because our format code needs to be precise about bit-width. That's why kernels and game engines use this shorthand in the first place, right?
> 
> I agree that common.h has taken on way too much responsibility and is a problem. I would rather gut common.h but keep the shorthand.
> 
> But I'm also not a professional C++ developer, and I'm open to getting rid of the type aliases. I would like to better understand why you disagree.

My case for removing them is that they are non-standard so it can be easy to forget they exist. That's why the codebase is inconsistent, I think (I get the impression that most of the code doesn't use these nor exact bit types, I might be wrong). Legibility wise they read almost the same to me, I do have a hard time parsing code that uses _both_.
I'm more pressed about consistent types and removing/simplifying `common.h` than strictly removing the shorthands. This was an "easy" way to do it.

Ultimately I'm a big supporter of whatever decreases mental load so, if you find value in keeping them, that's the better choice.
We could keep these in `common.h` and get rid of all the rest or create a new `vgmtypedef.h` and place them there.

sykhro:
In general, this kinda practice feels ancient to me. Times where `cstdint`/`cstddef` wasn't a thing and you needed to
```
using u8  = unsigned char;
using u16 = unsigned short;
using u32 = unsigned int;
using u64 = unsigned long long;
```
or aliasing to WORD DWORD LPCSTR MSWTFPLS to get the same code to compile across platforms. One could argue `cstddef` solves this problem badly.. on that I could agree. If the intent is just legibility, that's a fair concern

mikelow:
> My case for removing them is that they are non-standard so it can be easy to forget they exist. That's why the codebase is inconsistent, I think (I get the impression that most of the code doesn't use these nor exact bit types, I might be wrong).

I think it would be easier to remember if our code were consistently using the shorthand. Not sure about prevalence of fixed-width types but they are certainly very common in our codebase.

> Legibility wise they read almost the same to me, I do have a hard time parsing code that uses both.

Agreed that it's bad that we're mixing both. We need to be consistent. I would be happy to make a PR that (similar to this one) updates all usage to the shorthand.

> I'm more pressed about consistent types and removing/simplifying common.h than strictly removing the shorthands. This was an "easy" way to do it.

Makes sense. And I'm totally in agreement with getting rid of or simplifying common.h.

> We could keep these in common.h and get rid of all the rest or create a new vgmtypedef.h and place them there.

I'm down to make a new header file to house these type aliases alone. 

> In general, this kinda practice feels ancient to me. 

I've definitely come across modern C++ codebases using the same or similar shorthand. The [Dolphin](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Common/CommonTypes.h) emulator is an example.

Other modern languages have opted for fixed-width types that are less gross. Rust has `i16`, `u16`, etc. Swift and Go drop the `_t`.

> If the intent is just legibility, that's a fair concern

That is entirely my concern. Looking at the changes in this PR, I think switching to the standard types makes some of our code harder to read.

After doing a little research, and I'm wondering if it's now more common to use i16/u16 over s16/u16. Perhaps we should switch to that?

sykhro:
> I'm down to make a new header file to house these type aliases alone.

Let's do it then!

> Other modern languages have opted for fixed-width types that are less gross. Rust has i16, u16, etc. Swift and Go drop the _t.

I like Rust's, just wondering what to do about the float types. There are no fixed-width floats before c++23, I think it's probably fair to stick with `double` and `float`. Most of those usages should be aliased to units anyway, e.g. `Hz`

mikelow:
> Let's do it then!

Any other name suggestions? I don't love `vgmtypedef.h`. Perhaps`commontypes.h`?

> I like Rust's, just wondering what to do about the float types. There are no fixed-width floats before c++23, I think it's probably fair to stick with `double` and `float`.

I think we should use float / double. I don't think our usage of floating point values ever mandates precise fixed-width precision, so it's best they remain width-agnostic. `float` and `double` are very standard and much more readable than the fixed-width int types.

sykhro:
> Any other name suggestions? I don't love `vgmtypedef.h`. Perhaps`commontypes.h`?

`commontypes.h` works but i'm a little disturbed by the casing. then again `CommonTypes` is not a class so um

PR #887, opened 2026-05-23
Add modulator source mappings, add abstracted SeqTrack vibrato/tremolo methods

mikelow:
This adds ModDest -> ModSource mappings for SF2 and DLS (separately) in ConversionOptions, laying the groundwork that will allow users to specify which controllers or events to use for vibrato and tremolo modulators. This PR doesn't include an UI to change these mappings - that will have to come down the road. But this does establish sensible defaults, and it means that SF2 can utilize more appropriate CC sources instead of being constrained by DLS.

Of course, the mappings grant us a convenient layer of abstraction: SeqTrack can now provide methods for adding vibrato and tremolo events which resolve the configured MIDI output under the hood. Those new methods are:
- addVibratoDepth()
- addVibratoFrequency()
- addVibratoDelay()
- addTremoloDepth()
- addTremoloFrequency()
- addTremoloDelay()

There are also *NoItem() variants for each of these methods. I have updated all of the format call sites to utilize these methods instead of hard-coded controller / channel pressure events.

Separately, I've updated addStandardTremoloHandling() so that it uses a unique controller for LFO frequency instead of sharing the same source as vibLfoFreq. CapcomSnes, which shares a single LFO for vibrato and tremolo, has been updated to write MIDI events for both vibrato and tremolo frequency.

I've also renamed a number of methods for clarity.
- SeqTrack::setSynthLfoModulationDepth() -> emitVibratoDepth()
- SeqTrack::setSynthLfoControllerDepth() -> emitTremoloDepth()
- SeqTrack::advanceSynthLfoFadeToModulation() ->advanceVibratoDepthFade()

## Motivation and Context
Hardcoding modulator sources and their sequence-side event outputs was a bad scene. This simplifies modulator handling, makes the code easier to read, allows exported SF2s to use the full range of CCs, and will eventually give users more flexibility.

## How Has This Been Tested?
Tested a few of the formats utilizing modulators for deprecation by listening for vibrato and tremolo.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/components/seq/SeqTrack.cpp
+++ b/src/main/components/seq/SeqTrack.cpp
@@ -321,7 +321,9 @@ void SeqTrack::addForModSourceNoItem(ModSource source, uint8_t value) const {
       break;
     }
     case ModSource::None:
+      [[fallthrough]];
     case ModSource::PolyPressure:
+      [[fallthrough]];
     default:
       break;
   }
--- a/src/main/components/seq/SeqTrack.cpp
+++ b/src/main/components/seq/SeqTrack.cpp
@@ -301,26 +301,72 @@ void SeqTrack::addControllerSlide(uint32_t dur,
   prevVal = targVal;
 }
 
-// Emit synth vibrato depth through CC1/modulation if it changed.
-bool SeqTrack::setSynthLfoModulationDepth(SeqSynthLfoAutomation& automation, uint8_t depth, bool force) {
+void SeqTrack::addModSourceNoItem(ModSource source, uint8_t value) const {
+  if (readMode != READMODE_CONVERT_TO_MIDI) {
+    return;
+  }
+
+  if (auto controller = midiControllerForModSource(source)) {
+    pMidiTrack->addControllerEvent(channel, *controller, value);
+    return;
+  }
+
+  switch (source) {
+    case ModSource::ChannelPressure:
+      pMidiTrack->addChannelPressure(channel, value);
+      break;
+    case ModSource::PitchWheel: {
+      const auto bend = static_cast<int16_t>(std::clamp<int>((static_cast<int>(value) * 128) - 8192,
+                                                             -8192,
+                                                             8191));
+      pMidiTrack->addPitchBend(channel, bend);
+      break;
+    }
+    case ModSource::None:
+    case ModSource::PolyPressure:
+    default:
+      break;
+  }
+}
+
+void SeqTrack::addModDestSourceNoItem(ModDest destination, uint8_t value) const {
+  const auto target = ConversionOptions::the().midiModulationSourceTarget();
+  addModSourceNoItem(ConversionOptions::the().modSourceFor(target, destination), value);
+}
+
+void SeqTrack::addLfoModulationEvent(ModDest destination,
+                                     uint32_t offset,
+                                     uint32_t length,
+                                     uint8_t value,
+                                     const std::string& eventName,
+                                     Type type) {
+  const bool isNewOffset = onEvent(offset, length);
+  recordSeqEvent<SeqEvent>(isNewOffset,
+                           getTime(),
+                           offset,
+                           length,
+                           eventName,
+                           type,
+                           fmt::format("Value: {:d}", value));
+  addModDestSourceNoItem(destination, value);
+}
+
+// Emit synth vibrato depth through the configured ModDest source if it changed.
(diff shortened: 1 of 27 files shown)


mikelow:
I've addressed both points of PR feedback from codex.

src/main/components/seq/SeqTrack.cpp
```diff
@@ -301,27 +301,71 @@ void SeqTrack::addControllerSlide(uint32_t dur,
...
+  switch (source) {
+    case ModSource::ChannelPressure:
+      pMidiTrack->addChannelPressure(channel, value);
+      break;
+    case ModSource::PitchWheel: {
+      const auto bend = static_cast<int16_t>(std::clamp<int>((static_cast<int>(value) * 128) - 8192, -8192, 8191));
+      pMidiTrack->addPitchBend(channel, bend);
+      break;
+    }
+    case ModSource::None:
```
sykhro, line 323:
```suggestion
    case ModSource::None:
         [[fallthrough]];
```

src/main/components/seq/SeqTrack.cpp
```diff
@@ -301,27 +301,71 @@ void SeqTrack::addControllerSlide(uint32_t dur,
...
+    case ModSource::PolyPressure:
```
sykhro, line 325: ditto

src/main/conversion/VGMExport.h
```diff
@@ -44,6 +45,11 @@ void saveAs(const VGMColl &coll, const std::filesystem::path &dir_path) {
   auto filepath = dir_path / filename;
 
   if constexpr ((options & Target::MIDI) != 0) {
+    // ScopedMidiModulationSourceTarget uses RAII to set whether we're exporting for SF2 or DLS in ConversionOptions.
+    constexpr auto midiModulationTarget =
+        (options & Target::DLS) != 0 ? ModulationSourceTarget::DLS : ModulationSourceTarget::SoundFont;
+    const ScopedMidiModulationSourceTarget scopedModulationTarget(midiModulationTarget);
```
sykhro, line 51: So this is a thing that mutates a global for as long as it exists in scope?

sykhro (approved):
I think this is a good direction, I've left a couple small comments. In the future I think I'd like to have less reliance on globals/singletons and have conversion carry options in a "conversion context" of sorts. It's getting a bit hard to keep track of everything. 
Love the abstractions, I think it's great for us to be able to encode as much info as possible without being too tied to what DLS/SF2 can do. I've been toying with the idea of being able to export patches for decentsampler and this kind of work might make that a realistic long term goal

mikelow:
> In the future I think I'd like to have less reliance on globals/singletons and have conversion carry options in a "conversion context" of sorts. It's getting a bit hard to keep track of everything.

I figured the alternative was a much larger refactor. But I agree this is icky.

I've now implemented a refactor in an attempt to address this. I figure it's easier to review if I put it in a [new PR](https://github.com/vgmtrans/vgmtrans/pull/889) rather than combine it with this one.

> Love the abstractions, I think it's great for us to be able to encode as much info as possible without being too tied to what DLS/SF2 can do. I've been toying with the idea of being able to export patches for decentsampler and this kind of work might make that a realistic long term goal

Thanks! I agree totally on that point. That was my thinking behind SynthFile - to create an abstracted representation of an instrument agnostic of the exported format. I don't think I quite followed through.

src/main/conversion/VGMExport.h
mikelow: Yeah. It's a bit icky. [This PR](https://github.com/vgmtrans/vgmtrans/pull/889) attempts to address the issue.

sykhro:
> That was my thinking behind SynthFile - to create an abstracted representation of an instrument agnostic of the exported format. I don't think I quite followed through.

It's hard to design an abstration when the only two customers have decent overlap (and confusing docs, poor public implementations, ...)

mikelow:
> It's hard to design an abstration when the only two customers have decent overlap (and confusing docs, poor public implementations, ...)

That's more sympathy than I deserve.

sykhro:
> > It's hard to design an abstration when the only two customers have decent overlap (and confusing docs, poor public implementations, ...)
> 
> That's more sympathy than I deserve.

nah man, hacking on this program is what got me started with c++ as a teen. you have influenced my career!

mikelow:
> nah man, hacking on this program is what got me started with c++ as a teen. you have influenced my career!

I'm delighted to know that! Thank you for sharing.

PR #889, opened 2026-05-24
ConversionContext refactor

mikelow:
This removes the `ScopedMidiModulationSourceTarget` class, which used RAII to temporarily mutate the global `ConversionOptions` during export. Instead, conversion-time settings are placed into a `ConversionContext` struct which is and passed through VGMSeq and SF2/DLS export paths.

The ConversionContext object can be constructed using a reference to ConversionOptions, and the SynthTarget (SF2 or DLS). ConversionContext stores the following:
- bank select style
- sequence loop count
- channel 10 skip behavior
- SF2/DLS modulation source maps
- synth target (SF2 or DLS)

There is still some statefulness in this approach. We store the active conversion context on a `VGMSeq` during conversion, which allows `SeqTrack` to access it. Passing the context object throughout the entire VGMSeq conversion pipeline seemed like a messier solution.

`ModulationSourceTarget` has been renamed to `SynthTarget`, and it along with BankSelectStyle have been moved into a new ConversionTypes.h file for cleaner header inclusion.

## Motivation and Context
The previous RAII approach made conversion behavior depend on temporarily mutating global state, which is not the cleanest way to do things.

## How Has This Been Tested?
Verified that conversion is still working for both SF2 and DLS.

## Types of changes

- [x] Refactor

## Checklist:


src/main/components/seq/VGMSeq.cpp
```diff
@@ -96,6 +103,7 @@ MidiTrack *VGMSeq::firstMidiTrack() {
 }
 
 bool VGMSeq::load() {
+  setConversionContext(ConversionContext::fromOptions(ConversionOptions::the(), SynthTarget::SoundFont));
```
sykhro: is this alright to keep here considering the things that may override/virtual the `load()` method?

sykhro:
I like this, unfortunately removing all the statefulness sounds like a much much much bigger refactor (probably very bug inducing too)

mikelow:
> I like this, unfortunately removing all the statefulness sounds like a much much much bigger refactor (probably very bug inducing too)

I would gladly see this refactored this further, I'm just not sure what the better approach is here.

src/main/components/seq/VGMSeq.cpp
mikelow: Good catch. I've moved this `setConversionContext()` call into `VGMSeq::loadVGMFile()` before the `load()` call.

sykhro:
> > I like this, unfortunately removing all the statefulness sounds like a much much much bigger refactor (probably very bug inducing too)
> 
> I would gladly see this refactored this further, I'm just not sure what the better approach is here.

Maybe a policy-based design where the policy is injected at conversion time? (making a copy?) It could potentially replace the `CONVERT_TO_MIDI`/`ADD_TO_UI` mode thing too.
It's not the easiest thing to picture though.

PR #894, opened 2026-05-27
Remove common.h

mikelow:
## Summary

This PR does 3 things:

1. It replaces standard fixed-width types with our shortand across the codebase.
2. It standardizes project headers to use PascalCase names (no more lowercase headers).
3. It removes `src/main/util/common.h` and `common.cpp`, which had grown too expansive, and replaces them with smaller more specific headers:

- `base/Types.h`: fixed-width aliases such as `u8`, `s16`, `u32`, `s64`
- `base/Binary.h`: `Endianness`, `Signedness`, `swap_bytes16`, `swap_bytes32`
- `components/VGMFileVariant.h`: `VGMFileVariant` and its forward declarations
- `util/Text.h/.cpp`: `toUpper`, `toLower`
- `util/Path.h/.cpp`: `makeSafeFileName`, `pathToUtf8String`
- `util/SizeOffsetPair.h`: `SizeOffsetPair`

Call sites now include the narrower headers they actually depend on. 

I stuck to our old u32/s32 style-naming out of personal preference, but if you feel strongly, we can change this to u32/i32.

## How Has This Been Tested?
Code is compiling. CI will tell us if this is working across other platforms.

## Types of changes

- [x] Refactor

## Checklist:


sykhro:
I have one request, it doesn't have to be this PR though. I would like for headers to be included in this order.
```
<corresponding .h file for the current .cpp file>

<system headers>
<c/c++ standard library headers>

<external library headers>

<other vgmtrans headers used by this file>
```
I offer an alternative, if you're interested in controlling depenencies harder:
```
<corresponding .h file for the current .cpp file>

<other vgmtrans headers used by this file>

<system headers>
<c/c++ standard library headers>

<external library headers>
```

both of these in alphabetical order. 

it makes it easier to see what the dependencies are at a glance, it prevents third party libs from messing with system includes (e.g. windows.h macros). it helps agents too. i'm also somewhat ocd about this for no reason in particular

mikelow:
> I have one request, it doesn't have to be this PR though. I would like for headers to be included in this order.

Sounds good. Which of those orders do you prefer? The second one? I'll try to refactor to use the new ordering after I fix the compilation errors on Linux and Windows.

sykhro:
> > I have one request, it doesn't have to be this PR though. I would like for headers to be included in this order.
> 
> Sounds good. Which of those orders do you prefer? The second one? I'll try to refactor to use the new ordering after I fix the compilation errors on Linux and Windows.

I tend to use the first one out of habit (old google style guide). someone pointed out that by having the project headers _before_ other headers though you should be able to reduce transitive standard library includes. so let's try the second one!

mikelow:
Ok, include order has been updated.

mikelow:
I'll add a PR after this one adding the squashed commit to .git-blame-ignore-revs.

sykhro:
push that directly to master, github will end up with a new hash otherwise. other commits in the ignore revs file have the same issue

sykhro (approved):

mikelow:
> push that directly to master, github will end up with a new hash otherwise. other commits in the ignore revs file have the same issue

Happy to push a commit to master, but confused why it would matter. Squashing and rebasing changes the hash, but if I'm doing a second PR to add the prior commit to .git-blame-ignore-revs, that hash should remain stable, should it not?

sykhro:
> > push that directly to master, github will end up with a new hash otherwise. other commits in the ignore revs file have the same issue
> 
> Happy to push a commit to master, but confused why it would matter. Squashing and rebasing changes the hash, but if I'm doing a second PR to add the prior commit to .git-blame-ignore-revs, that hash should remain stable, should it not?

...right, I guess we didn't do that last time

mikelow:
> ...right, I guess we didn't do that last time

I'll investigate.

PR #895, opened 2026-05-27
add commits to .git-blame-ignore-revs

mikelow:
This adds a few commits to .git-blame-ignore-revs:

- commits circa 2016 when I did a codebase wide reformat
- a commit for a PR that renamed Matcher classes which I forgot to add
- the recent common.h removal / fixed-width type alias PR commit


sykhro (approved):

mikelow:
FYI, all of the other commit hashes seem to be valid.

PR #902, opened 2026-06-04
Fix warnings and centralize warning policy

mikelow:
## Summary

This PR updates and centralizes our compiler warning policy and fixes the resulting warnings.

The warning policy is centralized in `cmake/VGMTransWarnings.cmake`, which is now included in the root `CMakeLists.txt`. The cmake file contains the function `vgmtrans_enable_project_warnings(target)`, and we apply it to each of our targets (`vgmtranscore`, `vgmtrans`, `vgmtrans-shell`).

This PR also adds a `VGMTRANS_WARNINGS_AS_ERRORS` CMake option, implemented with CMake's `COMPILE_WARNING_AS_ERROR` target property. It's defaulted to off, but I wonder if we should enable it for CI builds.

@sykhro I consider you the CMake/build system expert, so I defer decisions around this to you. My goal is to clean up all of our compiler warnings. Don't hesitate to point out problems with the approach I've taken.

 ## Warning Policy Changes

Project targets now use the following shared warning setup:

- `-Wall`
- `-Wextra`
- `-Wno-unused-parameter`
- `-Wcast-align`
- `-Wnull-dereference`
- `-Woverloaded-virtual` for C++
- Clang/AppleClang: `-Wshadow`
- GCC: `-Wshadow=local`

## Rationale

Previously, project warning flags were defined in the Linux/macOS platform files and applied through global CMake compiler settings. That tied warning behavior to the selected platform configuration and allowed our warning policy to leak into vendored targets.

This PR moves warning policy into an explicit helper applied only to our targets, leaving platform files responsible for platform/toolchain concerns.

The CMake minimum version is raised to 3.24 so we can use `COMPILE_WARNING_AS_ERROR` directly instead of maintaining a manual `-Werror` / `/WX` path.

## Warning Selection Notes

`-Wunused` is no longer listed explicitly because it's encapsulated by `-Wall`.
`-Wunused-parameter` is suppressed because the codebase has many fixed interface signatures where some implementations naturally do not use every argument.

The biggest change is that `-Wold-style-cast` is not included in the new shared warning policy. It was unclear to me whether this was worth adding - a lot of major C++ projects suppress it, and we have about 170 instances of this warning throughout our codebase. If we do add it, it is probably best in a supplemental PR.

Shadow warnings are enabled so that we catch local shadowing issues without forcing broad constructor/member parameter renaming churn, which would provide little to no value anyway. I have a separate branch that cleans up constructor parameter naming

## MSVC

MSVC now gets an explicit warning configuration as well. This is intended to approximate the GCC/Clang policy on Windows without allowing MSVC to become the strictest compiler in the matrix.

The MSVC setup uses `/W4` with narrow suppressions for diagnostics that do not match the GCC/Clang policy closely enough, such as unused parameters and class-member shadowing noise.

## Details

This refactor was executed with Codex, and I had it document all of its changes. You can see that in addition to fixing warnings, it performed some more general cleanup:

#### Build System Policy

- Moved VGMTrans warning flags out of platform toolchain files and into `cmake/VGMTransWarnings.cmake`, applied only to VGMTrans-owned targets. This keeps third-party code under `lib/` from inheriting project warning flags.
- Added `VGMTRANS_WARNINGS_AS_ERRORS` as an opt-in CMake option instead of forcing warnings-as-errors by default. This matches the common maintainer/CI pattern while keeping local dependency/compiler drift from breaking ordinary builds.
- Chose a GCC/Clang/AppleClang policy centered on `-Wall`, `-Wextra`, `-Wcast-align`, `-Wnull-dereference`, `-Wshadow`, and `-Woverloaded-virtual`, with `-Wno-unused-parameter` because this codebase has intentional virtual/default callback parameters.
- Left `-Wpedantic` out of the project policy because project code uses compiler-supported extensions such as anonymous structs in binary-layout helpers.
- Left `-Wold-style-cast` out of the project policy. It is useful in small modernized projects, but in VGMTrans it produced noise around C APIs and vendored macro expansions without a proportional bug-finding payoff for this cleanup.
- For C sources, omitted `-Wpedantic` because `src/ui/shell/linenoise.c` uses common C variadic macro patterns that are not worth rewriting as part of the C++ warning policy.
- Suppressed Qt's private-module configure warning after confirming the project intentionally links `Qt6::GuiPrivate`.

#### Mechanical Warning Cleanup

- Reordered constructor initializer lists to match declaration order in core sample, sequence, RIFF, format-specific instrument, and Qt RHI window classes. These changes remove `-Wreorder-ctor` diagnostics without changing construction semantics.
- Added missing `override` annotations to format macros and several derived parser/scanner classes so `-Woverloaded-virtual` can remain enabled with useful signal.
- Marked the PSX ADSR helper as `inline` because it is defined in a header included by multiple translation units.
- Corrected mismatched `class`/`struct` forward declarations for AKAO articulation data and MAME metadata.
- Removed an unused local reverb flag in AKAO drum-region parsing while preserving the UI annotation for the pan/reverb byte.

#### Parser And State Warning Cleanup

- Replaced unused parser byte variables with explicit offset advancement when the data is only skipped, preserving event lengths and stream position without manufacturing dummy reads.
- Removed private parser state fields and local constants that had no readers, including stale sequence-loop, heartbeat-key, PS1 scanner, and FFT reverb placeholders.
- Added missing undefined-version handling in CPS switch statements so enum switches are exhaustive under the warning policy.
- Preserved useful decoded values in event descriptions where the warning revealed that the parser was reading data but formatting a different field.

#### Additional Warning Cleanup

- Fixed Konami Arcade tempo-slide reset to update sequence members rather than declaring unused locals.
- Kept Square PS2 BGM header validation effective by honoring the existing `bValid` flag before constructing a sequence.
- Removed redundant direct `fmt::fmt` linkage from the shell target; `vgmtranscore` already provides the needed transitive fmt dependency through spdlog.
- Scoped custom title-bar constants and icon helpers to platforms that use them so macOS builds do not carry dead UI declarations.

#### Shadow Warning Cleanup

- Enabled `-Wshadow` for VGMTrans-owned C and C++ targets while keeping dependency targets under `lib/` on their own warning settings.
- Renamed short parser locals and loop counters that hid outer parser state, including CPS, Konami, Rare SNES, Square PS2, TamSoft PS1, and Tri-Ace PS1 helpers.
- Renamed member-shadowing parameters in sequence tempo and volume helpers so function arguments remain clear without hiding stored sequence state.
- Renamed a repeated OPM operator reference in `YM2151.h` so every translation unit including the header is clean under `-Wshadow`.

## How Has This Been Tested?
AppleClang compilation is working without warnings. Tested playback of a number of formats.

## Types of changes

- [x] Refactor

## Checklist:


cmake/VGMTransWarnings.cmake
```diff
@@ -0,0 +1,37 @@
+# VGMTrans (c) 2002-2026
+# Licensed under the zlib license
+# Check the included LICENSE.txt for details
+
```
sykhro, line 4: missing `include_guard(GLOBAL)`

cmake/VGMTransWarnings.cmake
```diff
@@ -0,0 +1,37 @@
...
+      $<$<COMPILE_LANG_AND_ID:C,MSVC>:/W4>
+      $<$<COMPILE_LANG_AND_ID:C,MSVC>:/wd4100>
+      $<$<COMPILE_LANG_AND_ID:C,MSVC>:/wd4458>
+      $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/W4>
+      $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/wd4100>
+      $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/wd4458>
+  )
+
+  set_target_properties(${target} PROPERTIES COMPILE_WARNING_AS_ERROR ${VGMTRANS_WARNINGS_AS_ERRORS})
```
sykhro: here, I would argue for the following:
```
  if(VGMTRANS_WARNINGS_AS_ERRORS)
    set_property(TARGET ${target} PROPERTY COMPILE_WARNING_AS_ERROR ON)
  endif()
```

because otherwise the value of `VGMTRANS_WARNINGS_AS_ERRORS` dominates over setting `CMAKE_COMPILE_WARNING_AS_ERROR=ON` from the CLI/a preset

CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # Licensed under the zlib license
 # Check the included LICENSE.txt for details
 
-cmake_minimum_required(VERSION 3.23)
+cmake_minimum_required(VERSION 3.24)
```
sykhro: for consideration: if we bump to 3.25 (released 2022), we get to do this with vendored dependencies:
```
add_subdirectory(lib/foo SYSTEM)
```
which makes cmake pass include directories as -isystem or similar to the compiler. This has the effect of stopping any warning from third-party headers (at that point it would also be good to add `EXCLUDE_FROM_ALL` to prevent building _all_ of their targets by default)

sykhro (commented):
Thanks for doing this, it was on my todo list at some point. I have left a couple suggestions.

As for CI, yes, I think we should have warnings as errors. I would avoid flipping it in the local settings to avoid breaking compilation if someone is using a newer compiler than us

mikelow:
I've updated the branch with your feedback.

This includes updating the CI configs to treat warnings as errors. For the flatpak manifest, I'm utilizing the `secret-opts` field to pass in the `VGMTRANS_WARNINGS_AS_ERRORS` env var.

cmake/VGMTransWarnings.cmake
```diff
@@ -32,7 +33,7 @@ function(vgmtrans_enable_project_warnings target)
       $<$<COMPILE_LANG_AND_ID:C,MSVC>:/wd4458>
       $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/W4>
       $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/wd4100>
-      $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/wd4458>
+      $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/wd4æ58>
```
sykhro: typo?

mikelow: Yeeep

sykhro:
> Captura de tela 2026-06-04 132246
> 
> i've downloaded the last version [b058a17](https://github.com/vgmtrans/vgmtrans/commit/b058a1779c6315503b7be92a40a348ea002e9fff) until your build of windows failed bro

No need to be downloading binaries from PRs bro, use https://github.com/vgmtrans/vgmtrans-preview/releases/latest

migueldeoliveiravalentin84-arch:
yes ok

migueldeoliveiravalentin84-arch:
downloaded the latest version (the last one for windows)

mikelow:
I've fixed the remaining GCC warnings, and suppressed a larger set of MSVC warnings. MSVC has a more comprehensive interpretation of `-Wall`, and clang-cl attempts to interpret warning flags consistent with MSVC. MSVC warnings are more strict in a way I'm not sure we want to support.

However, MSVC did surface some warnings that we might want to adopt, so I've added `Wdeprecated-copy-with-dtor` to the main clang warning flag list and fixed code that triggered it.

sykhro (approved):

sykhro:
MSVC's /W4 is already broader than GCC's -Wall, /WAll is literally all the possible warnings (I think)

PR #904, opened 2026-06-06
Modernize ownership with smart pointers

mikelow:
## Summary

This PR attempts to migrate our pointer usage toward something resembling modern practices. Objects are now held with std::unique_ptr, while raw pointers are kept only as non-owning references.

The goal is to make lifetime rules easier to understand and safer to maintain. Most of the changes are mechanical updates across scanners, matchers, sequence/instrument/sample loaders, and export code to use the new ownership APIs.

@sykhro I'm curious to get your take on this approach. If you have a different vision for how to refactor the code, don't hesitate to share. We can scrap this PR and/or plan on another refactor down the road.

## What Changed

VGMRoot is now the central owner and registrar for loaded raw files, VGMFiles, and collections. Scanners and Matchers now hand newly created files and collections to Root through smart-pointer based APIs.

load() methods now focus on parsing and validation. They no longer register themselves globally as a side effect - that is instead handled by Root. This makes loading easier to reason about: construct, load, then let Root adopt/register the object.

Public raw-pointer containers are now private and exposed through read-only accessors.

New builder/adoption patterns were added:

- add*() creates and owns a child object in one step.
- adopt*() accepts an already-created unique_ptr.
- attach*() associates an existing object without taking ownership (like with VGMColl).
- raw pointers returned from these APIs are non-owning references for immediate use.

Synth, SoundFont, DLS, RIFF, and MIDI objects now use std::unique_ptr instead of manual deletion.

Format factories now return std::unique_ptr for scanners, matchers, and collections, removing another ownership handoff through raw pointers.

Several old raw-pointer helper patterns were removed, including obsolete vector/list/map deletion helpers.

## New Pattern

Code should no longer create owned objects with `new` and pass raw pointers around expecting another object to delete them later.

The preferred pattern is now:

```
auto* seq = pRoot->loadVGMFile<MySeq>(file, offset);
auto* region = instr->addRgn<MyRegion>(instr, offset, length);
auto* event = midiTrack->addEvent<NoteEvent>(...);
```

The caller can use the returned pointer, but it does not own it. Ownership lives in the parent object or in VGMRoot.

When ownership is transferred explicitly, the API takes a std::unique_ptr:

```
pRoot->loadVGMFile(std::move(file));
pRoot->loadVGMColl(std::move(coll));
parent.adoptChild(std::move(child));
```

Raw pointers still exist where they represent non-owning relationships, such as links between formats, UI selections, parent references, Qt-owned objects, and temporary views into owned containers.

## How Has This Been Tested?
- Tested loading and playback of each format
- Tested that SF2/MIDI export is working
- Tested the collection stitcher
- ~~I plan to test for memory leaks using Xcode Instruments~~
- ran ASan and loaded a files across (virtually) all formats. Also ran macOS leaks test, since LSan apparently doesn't work on macOS. Detected a leak in SonyPS2Scanner and fixed it.

## Types of changes

- [x] Refactor

## Checklist:

### DIFF
--- a/src/main/components/VGMItem.cpp
+++ b/src/main/components/VGMItem.cpp
@@ -18,14 +18,6 @@ VGMItem::VGMItem(VGMFile *vgmfile, u32 offset, u32 length, std::string name, Typ
       m_name(std::move(name)) {
 }
 
-VGMItem::VGMItem(const VGMItem& other)
-    : type(other.type),
-      m_vgmfile(other.m_vgmfile),
-      m_offset(other.m_offset),
-      m_length(other.m_length),
-      m_name(other.m_name) {
-}
-
 VGMItem::VGMItem(VGMItem&& other) noexcept
     : type(other.type),
       m_ownedChildren(std::move(other.m_ownedChildren)),
--- a/src/main/components/VGMItem.cpp
+++ b/src/main/components/VGMItem.cpp
@@ -18,9 +18,32 @@ VGMItem::VGMItem(VGMFile *vgmfile, u32 offset, u32 length, std::string name, Typ
       m_name(std::move(name)) {
 }
 
-VGMItem::~VGMItem() {
-  deleteVect(m_children);
+VGMItem::VGMItem(const VGMItem& other)
+    : type(other.type),
+      m_vgmfile(other.m_vgmfile),
+      m_offset(other.m_offset),
+      m_length(other.m_length),
+      m_name(other.m_name) {
 }
+
+VGMItem::VGMItem(VGMItem&& other) noexcept
+    : type(other.type),
+      m_ownedChildren(std::move(other.m_ownedChildren)),
+      m_children(std::move(other.m_children)),
+      m_childrenSorted(other.m_childrenSorted),
+      m_childrenPrefixMaxEnd(std::move(other.m_childrenPrefixMaxEnd)),
+      m_vgmfile(other.m_vgmfile),
+      m_offset(other.m_offset),
+      m_length(other.m_length),
+      m_name(std::move(other.m_name)) {
+  for (auto *child : m_children) {
+    child->m_parent = this;
+  }
+  other.m_children.clear();
+  other.m_childrenPrefixMaxEnd.clear();
+}
+
+VGMItem::~VGMItem() = default;
 
 bool operator>(const VGMItem &item1, const VGMItem &item2) {
   return item1.offset() > item2.offset();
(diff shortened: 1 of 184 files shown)


sykhro:
I love this, I think it was long time coming.
I have to say, modern C++ tries to be very value-oriented. It has its benefits (immutability, potentially easier concurrency, cache friendliness). But for our codebase, a mechanical conversion to (smart) pointers feels infinitely more realistic. 
I'm going to take a look and review it now

sykhro:
> This PR attempts to migrate our pointer usage toward something resembling modern practices. Objects are now held with std::unique_ptr, while raw pointers are kept only as non-owning references.

I would take this a step further and, in our types header, introduce a NonNull ptr type (see gsl::not_null). This way we can force raw ptr to a semantic of "optional, non-owning reference"

> load() methods now focus on parsing and validation. They no longer register themselves globally as a side effect - that is instead handled by Root. This makes loading easier to reason about: construct, load, then let Root adopt/register the object.

haven't looked at the code yet, but I attempted to do this in the past. returning a vector of "detected files" instead of registering them through the load method

> Public raw-pointer containers are now private and exposed through read-only accessors.

splendid. a small step towards basic UI concurrency...

> adopt*() accepts an already-created unique_ptr.

the accepted verb in the c++ community for this is _sink_

> I plan to test for memory leaks using Xcode Instruments

`-fsanitize=memory` is a thing too

src/main/components/instr/VGMInstrSet.cpp
```diff
@@ -28,49 +30,61 @@ VGMInstrSet::VGMInstrSet(const std::string &format, RawFile *file, u32 offset, u
...
-  bool val = load();
-  if (!val) {
-    return false;
-  }
+VGMInstr* VGMInstrSet::adoptInstr(std::unique_ptr<VGMInstr> instr) {
+  auto* rawInstr = instr.get();
+  m_instrs.push_back(rawInstr);
+  m_ownedInstrs.emplace_back(std::move(instr));
+  return rawInstr;
+}
```
sykhro: the idiomatic c++ way for this("sink") is to take `std::unique_ptr<T> &&` and then move it

sykhro: so this would become:
```cpp
VGMInstr* VGMInstrSet::sinkInstr(std::unique_ptr<VGMInstr>&& instr) {
  auto rawInstr = m_ownedInstrs.emplace_back(std::move(instr)).get();
  m_instrs.push_back(rawInstr);
  return rawInstr;
}
```

sykhro: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r32-take-a-unique_ptrwidget-parameter-to-express-that-a-function-assumes-ownership-of-a-widget

src/main/components/instr/VGMInstrSet.cpp
```diff
@@ -99,17 +110,17 @@ bool VGMInstrSet::parseInstrPointers() {
 }
 
 bool VGMInstrSet::loadInstrs() {
-  size_t nInstrs = aInstrs.size();
```
sykhro, line 102: at this point, might as well turn this into a range-based for loop

src/main/components/instr/VGMInstrSet.cpp
```diff
@@ -121,17 +132,27 @@ void VGMInstrSet::prepareForExport(const VGMColl* coll) {
...
+  m_tempInstrs.clear();
 }
 
 const std::vector<VGMInstr*>& VGMInstrSet::exportInstrs() const {
-  return m_exportInstrs.empty() ? aInstrs : m_exportInstrs;
+  return m_exportInstrs.empty() ? m_instrs : m_exportInstrs;
 }
 
-void VGMInstrSet::addTempInstr(VGMInstr* instr) {
+void VGMInstrSet::adoptTempInstr(std::unique_ptr<VGMInstr> instr) {
```
sykhro: ditto for the `sink`

src/main/components/instr/VGMInstrSet.h
```diff
@@ -38,22 +39,51 @@ class VGMInstrSet : public VGMFile {
   void prepareForExport(const VGMColl* coll);
   void cleanupAfterExport();
 
+  [[nodiscard]] const std::vector<VGMInstr*>& instrs() const { return m_instrs; }
```
sykhro, line 42: jsyk, starting c++20 `[[nodiscard]]` can take a string to explain the reason, `[[nodiscard("THIS IS WHY YOU NEED TO BE CAREFUL")]]`

src/main/components/instr/VGMInstrSet.h
```diff
@@ -38,22 +39,51 @@ class VGMInstrSet : public VGMFile {
...
+  SampCollType* addSampColl(Args&&... args) {
+    auto newSampColl = std::make_unique<SampCollType>(std::forward<Args>(args)...);
+    auto* rawSampColl = newSampColl.get();
+    adoptSampColl(std::move(newSampColl));
+    return rawSampColl;
+  }
+  void clearSampColl();
+
+  // Observer; use adoptSampColl/addSampColl<T> when the instrument set owns the sample collection.
   VGMSampColl *sampColl;
```
sykhro: might as well enforce this with a setter imo

sykhro (changes requested):
before I look at much of the rest, the sink change is important

src/main/components/instr/VGMInstrSet.cpp
mikelow: Thanks for the lesson. I've updated all adopt*() methods to use the pattenr you described and renamed them to sink*().

src/main/components/instr/VGMInstrSet.h
mikelow: I went a little further and made it a private member `m_sampColl`, gave it an public accessor `sampColl()`, in addition to a new method `attachSampColl()` for non-owning assignment

mikelow:
> I would take this a step further and, in our types header, introduce a NonNull ptr type (see gsl::not_null). This way we can force raw ptr to a semantic of "optional, non-owning reference"

Perhaps we should implement this in a supplemental PR? Looks like it will touch a lot of code.

> haven't looked at the code yet, but I attempted to do this in the past. returning a vector of "detected files" instead of registering them through the load method

Having load() return a list of detected files is an obviously cleaner design. Probably worth refactoring for in a supplemental PR.

sykhro:
> Perhaps we should implement this in a supplemental PR? Looks like it will touch a lot of code.

that's fair

> Having load() return a list of detected files is an obviously cleaner design. Probably worth refactoring for in a supplemental PR.

I remember running into issues with semantics of matchers/loaders, but it was probably because I was trying to make things parallel at the same time

src/main/components/instr/VGMInstrSet.cpp
```diff
@@ -25,68 +27,77 @@
...
+VGMInstr* VGMInstrSet::sinkInstrAsChild(VGMItem& parent, std::unique_ptr<VGMInstr>&& instr) {
+  auto* rawInstr = instr.get();
+  m_instrs.push_back(rawInstr);
+  parent.sinkChild(std::move(instr));
+  return rawInstr;
+}
+
+std::vector<std::unique_ptr<VGMInstr>> VGMInstrSet::releaseInstrs() {
+  m_instrs.clear();
+  return std::move(m_ownedInstrs);
```
sykhro: `std::move` will leave `m_ownedInstrs` in an unspecified state. given the class is stateful, `return std::exchange(m_ownedInstrs, {});` would be safer

src/main/components/instr/VGMInstrSet.cpp
```diff
@@ -144,6 +169,20 @@ VGMInstr::VGMInstr(VGMInstrSet *instrSet, u32 offset, u32 length, u32 bank,
       bank(bank), instrNum(instrNum), parInstrSet(instrSet), reverb(reverb) {
 }
 
+VGMInstr::VGMInstr(const VGMInstr& other)
```
sykhro: is this intended to be a shallow copy? (I don't see it copying any owned things)

sykhro (commented):
I think this is fine! I would be consistent with `std::exchange(..., {})` in release APIs + the two comments I've left. the rest seems fine!

src/main/components/instr/VGMInstrSet.cpp
mikelow: Good catch. It's leftover junk from a bad change. it's unused and I've deleted it.

src/main/components/VGMItem.cpp
```diff
@@ -18,10 +18,33 @@ VGMItem::VGMItem(VGMFile *vgmfile, u32 offset, u32 length, std::string name, Typ
       m_name(std::move(name)) {
 }
 
-VGMItem::~VGMItem() {
-  deleteVect(m_children);
+VGMItem::VGMItem(const VGMItem& other)
```
sykhro: ditto suspicious copy ctor

mikelow: deleted!

sykhro (approved):
I think we're good to go! I hope I didn't miss anything

PR #906, opened 2026-06-06
Load track titles from GSF, 2SF, NCSF

sykhro:
## Description
Adds a generic metadata hint provider system for `RawFile`, then uses it to preserve PSF-family soundtrack metadata when scanners operate on reconstructed virtual ROM/SDAT contents.

This change adds indexed metadata hints that loaders can attach to derived `VirtFile`s. `PSFLoader` now harvests sibling  files sharing the same `_lib`, extracts relevant tags and attaches hints for:
- GSF/miniGSF MP2k tracks, indexed primarily by MP2k song index
- 2SF/NCSF NDS tracks, indexed by SDAT sequence lookup key from `origFilename`

`MP2kScanner` and `NDSScanner` consume those hints to name matching sequences/collections from soundtrack tags while still scanning all valid entries.

Also adds reading the length tag from PSF files.

`VGMMetadataHintTarget`/`VGMMetadataHintQuery` describe how a scanner can identify a track-like object inside the reconstructed data. 

Scanners build one when they want to ask, “do we have metadata for this thing I just found?”. For example, MP2k can query by song-table index, while NDS can query by SDAT sequence name via `lookupKey`.

`VGMMetadataHint` wraps that target with the metadata source:

```cpp
target
sourceFormat
sourcePath
tag
```

So the hint says both “what scanned object this applies to” and “which source file/tag produced it.”

`VGMMetadataHintProvider` is the lookup interface attached to `RawFile`. Loaders can attach a provider to a derived `VirtFile`; scanners then call `file->findMetadataHint(query)` without needing to know where the hints came from.

The current implementation, `IndexedMetadataHintProvider`, builds indexes at construction time for O(1) lookup by song index, ROM address, file offset, or lookup key. Its stored hints and indexes are immutable after construction. If an indexed lookup is unavailable or ambiguous, it can still fall back to checking the stored hints using the full query match.

## Motivation and Context
Sometimes, opening *SF files results in many songs being loaded and the user doesn't know which collection maps to which collection. This results in a very time consuming process to go find the right sequence by converting them en masse (e.g. certain Mother 3 rips).

Opening files such as `.minigsf` or `.minincsf` reconstructs the whole backing ROM/SDAT, so scanners lose the per-track metadata from the file the user actually opened. This made MP2k tracks show as generic `MP2k Collection #n`, and NDS tracks show internal SDAT names (when available) like `BGM_JOHN`.

## How Has This Been Tested?
Tested on macOS.

- miniGSF Advance Wars shows soundtrack titles like `Opening`, `Eagle's Theme`
- direct `.gsflib` still shows fallback `MP2k Collection #...`
- miniNCSF Advance Wars Dual Strike shows soundtrack titles like `Jake's Theme`
- direct `.ncsflib` still shows internal names like `BGM_JOHN`
- PS1 PSF loading remains unchanged

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,11 +19,13 @@
 
 * **Sappy (mp2k):** Added support for missing commands (LFO, pitch bend range, transpose, etc.).
 * **Sappy (mp2k):** Added support for GBA PSG instruments and programmable wavetables.
+* **Sappy (mp2k):** When possible, track names from GSF files are used instead of the song table index.
 * **Sappy (mp2k):** Fixed sequences being exported 0.45% faster than they should be.
 * **Sappy (mp2k):** Fixed instruments potentially having very long release times.
 
 **Nintendo DS (SDAT)**
 
+* **2SF/NCSF:** When possible, track names from 2SF/NCSF is used instead of "NDS Seq" or internal track names.
 * **Instruments:** Improved accuracy of instrument ADSR.
 * **Instruments:** Fixed an issue where PSG instruments were exported too quiet (-4.6dB attenuated).
 * **Samples:** Fixed an issue where samples were being read with the wrong sample rate, resulting in instruments playing at the wrong pitch or seemingly being absent.
--- a/src/main/CMakeLists.txt
+++ b/src/main/CMakeLists.txt
@@ -501,6 +501,7 @@ target_sources(vgmtranscore
       components/VGMColl.h
       components/VGMFile.h
       components/VGMItem.h
+      components/VGMMetadataHint.h
       components/VGMMiscFile.h
       components/VGMSampColl.h
       components/VGMTag.h
--- a/src/main/components/PSFFile.cpp
+++ b/src/main/components/PSFFile.cpp
@@ -10,17 +10,70 @@
 #include "io/RawFile.h"
 #include "util/Decompression.h"
 
+#include <cerrno>
+#include <optional>
+#include <cstdlib>
+#include <string>
+#include <string_view>
+
 constexpr auto PSF_TAG_SIG = "[TAG]";
 constexpr auto PSF_TAG_SIG_LEN = 5;
 
+namespace {
+
+std::optional<double> parsePsfLength(std::string_view value) {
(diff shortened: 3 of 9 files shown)


sykhro:
This being said, I'm open to an even more generic design. But would need a usecase to think about it...

mikelow:
@codex review

Awesome. I'll take a look at this tomorrow.

mikelow:
Sorry the delayed review!

This is overall very good and solves a messy problem. My concerns are:
-  PSFLoader now takes on more platform-specific knowledge than I would expect it to (for instance, the constants at line 54-56), for the purpose of hint collection.
- There's a bit of reused logic between `collectGsfMetadataHints()` and `collectNdsMetadataHints()`.
- We're doing some things that are a little unexpected (like scanning the directory for sibling minixsf files) but not explaining this explicitly either via self-explanatory code or comments.
- Minor quibble: when xsflib files are directly loaded, we don't collect tags.

I wonder if we could extract the hint collection logic out of PSFLoader and into a set of helper functions in a separate file.

This would also give us an opportunity to make the rules for extracting hints more explicit. Right now, the rules are implicit inside of `collectGsfMetadataHints()` and `collectNdsMetadataHints()`. For instance, both share this logic:

```
  auto lib = findLibTag(psf);
  if (!lib) {
    return hints;
  }
```

I imagine something like this: 
```
std::vector<VGMMetadataHint> collectForOpenedFile(
    const RawFile& file, const PSFFile& psf) {
  std::vector<VGMMetadataHint> hints;

  append(hints, collectOwnHint(file, psf));

  if (auto lib = findLibTag(psf)) {
    append(hints, collectSiblingHintsForSameLib(file, psf));
  } else if (isLikelyLibraryFile(file, psf)) {
    append(hints, collectSiblingHintsReferencingThisLib(file, psf));
  }

  return hints;
}
```

I've actually gone ahead and implemented a refactor around this idea. Let me know your thoughts there.

I think it's fine to merge this now and then look at the refactor separately. If you'd rather merge the refactor into this branch, that's fine too.

mikelow (approved):

PR #913, opened 2026-06-24
Value-oriented rewrite proposal

mikelow:
@sykhro your [comment](https://github.com/vgmtrans/vgmtrans/pull/904#issuecomment-4640312627) on the smart pointer PR made me wonder what a more modern C++ value-oriented design would look like. This led me down a rabbit hole of re-architecting the app with heavy assistance from Codex over the past two weeks.

This PR is meant as a loose proposal. I would like to hear your feedback and ideas, including any criticism.

My main goals:
- Migrate to a value-oriented design that uses immutable structures and eschews objects and pointers as much as possible
- Decouple format code from MIDI. Format code should describe driver behavior and parse sequences into a format-agnostic IR which core code can later transform into MIDI or other formats.
- Keep format code intuitive and readable
- Add tests, including parity tests to prevent regressions for ported formats

The new code lives in the `src/value` directory. It exists in parallel with the old code, though the two are decoupled from each other. Because the scope is so large, I had ChatGPT write a [guide](https://github.com/user-attachments/files/29279313/vgmtrans_value_architecture_guide.md), which is probably clearer than what I could write anyway.

You should probably ignore the files in the docs directory. I had Codex write them to maintain context during major refactors. They likely contain outdated information, but for now I'll leave them for completeness' sake.

The new architecture has not been integrated into the UI yet; it can only be interacted with via vgmtrans-shell using the `value` commands. To prevent the scope from growing larger before any feedback, only three formats are ported to the new architecture. They pass the midi and synth parity tests I've thrown at them (though they deserve more comprehensive testing). Of course, integrating this architecture with the UI and adding new formats will require the core code to evolve further, though I've tried to account for some foreseeable needs.


sykhro:
Been very busy in the past week, will review soon! I like the general direction, need to gather a few thoughts

sykhro:
I didn't lose sight of this, there's many new concepts and I'm slowly working my way through every file. The rewrite is beneficial, I wonder if there's some room for simplification. It would be nice to see what potential UI code would look like and maybe one more format

mikelow:
Hey, sorry, for the late reply! I apologize for the massive size of this PR. Perhaps I took the wrong approach here. Maybe it would be better to regard this as a proof-of-concept for new architecture ideas, rather than a straight-up proposal.

Perhaps we should discuss this from the ground up, starting with:
- should we do a rewrite?
- if so, what should the goals be?

FWIW, after creating this PR, I ported over two more formats: AkaoSnes and KonamiSnes, but I didn't push those changes so as to prevent this PR from becoming a moving target. I've pushed them now to a new branch: `value-rewrite-akao-konami-snes`. I chose those formats to so as to integrate SequenceMotionPlan. Notably, I added an `ExportRequest` field to specify whether to use modulators or sequence events for vibrato/tremolo. That means that vibrato can be generated using pitch bend events instead of instrument-side modulators + control events, and tremolo can use expression events. That logic resides in PerformanceMidiRenderer, so it's independent of format code. It's trivial to push those changes into this branch if you would prefer.

I'm also happy to integrate this new architecture with the Qt UI code to test how well the design handles that.

sykhro:
> should we do a rewrite?

i'm not sure what constitutes a rewrite (that is, where is the line between refactor and rewrite) here, but...

> if so, what should the goals be?

...there are certainly some pain areas in the codebase. imo:

1. global & mutable state; this also has the effect of no ui multithreading
2. not enough abstraction between format code and output formats, meaning sometimes we don't represent format code as best as possible just to fit within the limitations of sf2/midi
3. no concept of a "session" so scanning a bunch of unrelated files can have weird results
4. no (regression) tests whatsoever
5. related to 2 and 4: sometimes a conversion bug is fixed the wrong layer

plus whatever random thing we've found over the years that doesn't fit the current model super well (instrument sets with more than one sample collection? sequences with more than 1 instrument set?).

I think a rewrite should address this stuff. there's definitely overlap with what you've done here! what do you find you don't like about the codebase in its current state?

sykhro:
the performancerenderer thing is a great example actually.
in my nds-mods(-2) branch, I've implemented complete LFO support for NDS (and made pitch bends and portamento as accurate as possible). to do this, I straight up render the LFO signal as MIDI CC. 
...however, FL studio will completely break the midi: it throws away events, reduces resolution, their amplitude, ...
so, for fl studio users, having a "use simplified tremolo" option would be useful

PR #914, opened 2026-06-28
AkaoMatcher: allow collections without a sample collection

CBongo:
## Description
The AKAO matcher requires a sequence, an instrument set, and at least one sample collection before it forms a VGMColl. As a result, an AKAO sequence recovered without its sample data (e.g. a sequence block extracted on its own, separate from the sample bank) produces no output at all, even though the sequence is fully usable as MIDI.

This PR makes the sample collection optional: a sequence plus its instrument set is sufficient. MIDI export is unaffected; SF2/DLS export already degrades gracefully when no samples are present; the normal PSF path (sequence + instrument set + samples) is unchanged.

## Motivation and Context
This allows for MIDI extraction from raw AKAO files containing only the sequence data; currently vgmtrans returns "File xxx is not a recognized music file" because no collections were formed, even though it successfully found sequence data.

## How Has This Been Tested?
Tested on Windows 11 cli-only build.
- PSF (Chrono Cross "The Brink of Death"): unchanged -- MIDI (50 KB) + SF2 (~960 KB with samples). No regression.
- Standalone AKAO sequence blocks (Chrono Cross, FF9) with no sample data: previously "not a recognized music file", now export valid MIDI.

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

Depending on how you interpret the error situation, this is either fixing the "bug" of not allowing extraction without samples, or is adding the "new feature" of allowing extraction without samples.  We report; you decide.

## Checklist:
- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.

The existing AkaoMatcher code does not strictly adhere to the CONTRIBUTING style, but the changes do.

### DIFF
--- a/src/main/formats/Akao/AkaoMatcher.cpp
+++ b/src/main/formats/Akao/AkaoMatcher.cpp
@@ -91,95 +91,97 @@ bool AkaoMatcher::tryCreateCollection(int id) {
   auto itSeq = seqs.find(id);
   auto itInstrSet = instrSets.find(id);
 
-  if (itSeq != seqs.end() && itInstrSet != instrSets.end() && !sampColls.empty()) {
-    AkaoSeq* seq = itSeq->second;
-    AkaoInstrSet* instrSet = itInstrSet->second;
+  // A sequence and its instrument set are required to form a collection. The sample
+  // collection is optional: a sequence recovered without its sample data (for example a
+  // sequence block extracted on its own, separate from the sample bank) still produces a
+  // collection so it can be exported to MIDI. SF2/DLS export simply has no samples to use.
+  if (itSeq == seqs.end() || itInstrSet == instrSets.end()) {
+    return false;
+  }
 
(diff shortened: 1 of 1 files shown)


CBongo:
Note: this addresses the restriction at the matcher level; a more general follow-up (exporting sequences that don't match a full collection, at a higher level) is planned.

sykhro:
If found, MIDI sequences can be exported even if they are not in a collection. Am I missing something?

CBongo:
In AkaoMatcher at least, the sequences will not be returned (and thus not found) unless it also finds both instrument definitions AND samples; this PR relaxes the requirement to only require a sequence and instrument definitions.

CBongo:
The longer term plan is to support loading from a wider variety of files than "cooked" .SPC, .PSF, etc. such as game ROMS directly (with a little help from hint files that tell where to find the data and how to decode it).

PR #921, opened 2026-07-26
Fix NinSnes level conversion

mikelow:
## Summary

- convert NinSnes 8-bit level controls to squared linear gain before performance rendering
- apply the shared conversion to melodic, percussion, tie, channel, master, fade, and Intelligent Systems voice-record paths
- add regression coverage for all 17 supported NinSnes profiles and the Corneria-style 252/152 velocity pair

## Root cause

The value port passed `raw / 255.0` to the performance model as linear gain. The MIDI renderer correctly applies the inverse amplitude curve, so this inflated lower source levels and compressed dynamics. Squaring the normalized source value restores the established legacy `raw / 2` MIDI mapping while keeping the performance model in linear-gain space.

## Impact

Corneria's representative velocity pair now renders as 126/76 instead of 126/98. Channel and master levels follow the same corrected curve across every profile path that exposes those commands. No shared renderer or other-format behavior is changed.

## Validation

- `cmake --build cmake-build-debug --target vgmtrans-value-core-tests -j 8`
- `cmake-build-debug/tests/vgmtrans-value-core-tests`
- `git diff --check`
- full CTest run: 7/9 passed; the two failures were unrelated existing CapcomSnes MIDI parity corpus mismatches (`MMX-04` and `Launch Octopus`)


sykhro:
I presume you meant to cherry-pick something from that branch :D

mikelow:
Sorry! I think codex made this PR even though I didn't request it. Kind of concerning.

As you can see I'm working on improving the rewrite. Sorry I've been absent!

sykhro:
No worries! I'm as busy as ever myself. Haven't had much time to look at anything 🥲

